PR #3, opened 2023-06-29
[SCC-3597] Initialize production and development docker config

dgcohen:
This PR initializes the Docker configuration for the new research catalog app.

To run locally:
cd into research-catalog
run `docker-compose up --build --force-recreate`


docker-compose.yml
```diff
@@ -0,0 +1,16 @@
+version: '1'
```
EdwinGuzman, line 1: I get this error locally:
Screen Shot 2023-06-30 at 9 51 28 AM

Try version `'2'`, that works for me. Looks like v1 won't be supported anymore https://docs.docker.com/compose/compose-file/compose-versioning/

EdwinGuzman (changes requested):

docker-compose.yml
EdwinGuzman: Or version 3. I only just tried 2 but maybe 3 is the way to go.

dgcohen: Good call! Updated to version 3.

EdwinGuzman (approved):

nonword (approved):
Looks good to me!

PR #4, opened 2023-07-17
SCC-3594 - Implement homepage

dgcohen:
This PR introduces all the existing homepage components to the new Next app. It also introduces several bits of necessarey infrastructure (SASS based CSS modules, Layout component, etc) to construct it.

Part of reviewing this PR should involve deciding whether these patterns work for the team, as they are not set in stone and are more of an initial proof-of-concept.

NB: I've also downgraded Node to 16 and added the baseurl of /research/research-catalog to nextConfig, so make sure you are running the proper node version locally and testing using this url : http://localhost:3000/research/research-catalog


components/ImageCard/ImageCard.tsx
```diff
@@ -0,0 +1,20 @@
...
+import styles from "@/styles/components/ImageCard.module.scss"
+
+interface ImageCardProps {
+  imageSrc: string
+  children: ReactElement
+}
+
+const ImageCard = ({ imageSrc, children }: ImageCardProps) => {
+  return (
+    <article className={styles.imageCardContainer}>
```
EdwinGuzman, line 11: Why the use of `article`? We should get Clare's input on this.

dgcohen: I've seen article used as the semantic tag for this type of element in the past, but it could just as easily be a div or something else. Happy to consult Clare on this.

EdwinGuzman (changes requested):
Overall the homepage looks good but there are some questions and edits to be made. This review was done with Vera and Daniel.

styles/utils/_breakpoints.scss
```diff
@@ -0,0 +1,8 @@
+// Reservoir DS breakpoint note. There are four SCSS variables for
+// breakpoint values imported from the DS `resources.scss` file:
+$nypl-breakpoint-small: 320px;
```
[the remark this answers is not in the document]
dgcohen: can remove them and add them back as needed

styles/utils/_colors.scss
```diff
@@ -0,0 +1,26 @@
...
+$grayMedium: var(--nypl-colors-ui-gray-medium);
+$grayLightCool: var(--nypl-colors-ui-gray-light-cool);
+$grayXLightCool: var(--nypl-colors-ui-gray-x-light-cool);
+$grayXXLightCool: var(--nypl-colors-ui-gray-xx-light-cool);
+$grayXLightWarm: var(--nypl-colors-ui-gray-x-light-warm);
+
+$linkPrimary: var(--nypl-colors-ui-link-primary);
+$linkSecondary: var(--nypl-colors-ui-link-secondary);
+
+$brandPrimary: var(--nypl-colors-brand-primary);
```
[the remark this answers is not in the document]
dgcohen: good call, will remove the unused ones and add them back as needed

.nvmrc
```diff
@@ -1 +1 @@
-16
\ No newline at end of file
+18
```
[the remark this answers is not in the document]
dgcohen: I actually misremembered this, bumped it up from 16 but forgot what the issue was that necessitated this.

pages/_document.tsx
```diff
@@ -1,14 +1,37 @@
...
+        }
+      `}</style>
       <Head>
-        <meta name="description" content="NYPL Research Catalog" />
+        <meta name="description" content="Research Catalog | NYPL" />
       </Head>
       <body>
+        {/* NYPL Header script and container */}
+        <Script
+          strategy="lazyOnload"
```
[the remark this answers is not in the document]
dgcohen: It doesn't have it. Just replaced it with a regular <script> tag with async.

components/SubNav/SubNavLink.tsx
```diff
@@ -0,0 +1,20 @@
...
+interface SubNavLinkProps {
+  active?: boolean
+  href: string
+  children: string | ReactElement
+}
+
+const SubNavLink = ({ active, href, children }: SubNavLinkProps) => {
+  return (
+    <Link href={href} passHref>
+      <DSLink className={active && styles.active}>{children}</DSLink>
```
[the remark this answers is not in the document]
dgcohen: This is a good idea. Unfortunately the design system Link component doesn't support inline styles at the moment.

components/ImageCard/ImageCard.tsx
EdwinGuzman: This was a question out of curiosity and not mean to be something that you _had_ to update.

pages/_document.tsx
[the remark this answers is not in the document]
EdwinGuzman: Does using Next's Script component  have any benefits?

components/SubNav/SubNav.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * and passes the active prop to the correct link based on the route.
+ */
+const SubNav = () => {
+  const router = useRouter()
+
+  return (
+    <nav className={styles.subNav} aria-label="Research Catalog main menu">
+      <ul>
+        <li>
+          <RCLink href="/" active={router.pathname === "/"}>
```
EdwinGuzman, line 16: Similar comment regarding inline logic checks. There should be a more manageable and scalable way to have an `active` prop set.

components/Search/Search.tsx
```diff
@@ -0,0 +1,47 @@
...
+import RCLink from "@/components/RCLink/RCLink"
+import { SearchBar } from "@nypl/design-system-react-components"
+
+/**
+ * The Search component renders and controls the Search form and
+ * advanced search link.
+ */
+const Search = () => {
+  return (
+    <section role="search" className={styles.searchContainer}>
```
EdwinGuzman, line 11: The DS' `SearchBar` component already adds a `role="search"` to the form element so this is redundant.

styles/components/Search.module.scss
```diff
@@ -0,0 +1,44 @@
...
+
+.searchContainer {
+  background-color: var(--nypl-colors-ui-gray-x-light-cool);
+  padding: var(--nypl-space-l) 0 var(--nypl-space-xs);
+}
+
+.searchFormContainer {
+  @include maxWidth;
+
+  @include media(var(--nypl-breakpoint-xl)) {
```
EdwinGuzman, line 11: This doesn't work. CSS var arguments don't actually get passed to mixins at compile time.

styles/components/Search.module.scss
```diff
@@ -0,0 +1,44 @@
...
+    > div {
+      width: 85%;
+    }
+  }
+
+  form {
+    .searchBarContainer {
```
EdwinGuzman, line 18: I don't think this rule does anything.

EdwinGuzman (changes requested):
Some more minor updates to do.

styles/components/Search.module.scss
dgcohen: Good catch, this was a holdover from a different approach I'd taken. Removed.

styles/components/Search.module.scss
dgcohen: Now I understand why the hardcoded values were in the breakpoints scss file on discovery-front-end. I restored it, but let me know if I should take a different approach.

components/Search/Search.tsx
dgcohen: Good catch, fixed.

components/RCLink/RCLink.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * The RCLink component is a utility that wraps the design system Link component with
+ * Next's Link component to allow for correct navigation in Next per the design system's
+ * docs. It also includes an 'active' prop used for styling the SubNav component.
+ */
+const RCLink = ({ href, children, active = false, ...rest }: RCLinkProps) => {
+  return (
+    <Link href={href} passHref {...rest}>
+      <DSLink fontWeight={active && 700}>{children}</DSLink>
```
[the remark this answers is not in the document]
dgcohen: Good call, replaced with the design system variable for bold font weight.

components/Layout/Layout.tsx
```diff
@@ -0,0 +1,48 @@
...
+            />
+            <div className={styles.researchHeadingContainer}>
+              <Heading id="heading-h1" level="one" text="Research Catalog" />
+              <SubNav />
+              {showSearch && <Search />}
+            </div>
+          </>
+        }
+        contentPrimary={children}
+      ></TemplateAppContainer>
```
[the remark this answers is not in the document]
dgcohen: Good call, fixed.

components/RCLink/RCLink.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * The RCLink component is a utility that wraps the design system Link component with
+ * Next's Link component to allow for correct navigation in Next per the design system's
+ * docs. It also includes an 'active' prop used for styling the SubNav component.
+ */
+const RCLink = ({ href, children, active = false, ...rest }: RCLinkProps) => {
+  return (
+    <Link href={href} passHref {...rest}>
+      <DSLink fontWeight={active && "var(--nypl-fontWeights-bold)"}>
```
EdwinGuzman, line 19: Use design tokens in jsx files and css vars in the css module files. In this case, this can be `"bold"` instead of `"var(--nypl-fontweights-bold)"`.

EdwinGuzman (approved):

dgcohen:
@charmingduchess No reason for it other than the convenience of building paths from the root rather than having to count backtraces ('../../..') but i wasn't aware that this caused issues in VS code. Happy to used .. syntax if you prefer it.

PR #6, opened 2023-07-21
Add pre-commit hooks for linting/formatting

dgcohen:
This PR adds husky and lint-staged to automatically lint check and format the files on commit. This PR also includes some linting fixes that were performed by the new pre-commit checks.


charmingduchess (approved):
Looks great! What do you think about adding an npm script to run these on the fly and document in readme?

dgcohen:
> Looks great! What do you think about adding an npm script to run these on the fly and document in readme? Also, there are now a bunch of linting/formatting files. Would it be super difficult to move those into their own directory so we can keep the root looking sleek?

Good call on adding docs, I'll add some in a moment. There are currently npm commands for running prettier and lint independently (will document these), not sure if there needs to be a command that runs both but let me know if you disagree. Btw I have my IDE configured to format and lint when a file is saved (in addition to a key command), which avoids having to re-commit all the formatting/linting changes run by the pre-commit hooks.

Not sure if the prettier and lint config files can be moved out of the root, the documentation for both specifies that they need to be there but there could be a workaround I'm not familiar with.

EdwinGuzman (approved):

PR #7, opened 2023-07-21
SCC-3593 - AWS config

dgcohen:
This PR ports the AWS KMS decryption utility script from discovery-front-end to the new research-catalog app, fixing deprecations and typescript errors.

It also includes documentation pertaining to Env variables derived from the old app, with some changes reflecting differences in file names, as well as the removal of env variables that haven't been added to the new app yet.


ENVIRONMENTVARS.md
charmingduchess, line 1: 🤩 shiny!

ENVIRONMENTVARS.md
```diff
@@ -0,0 +1,59 @@
...
+
+As previously mentioned in the [README](README.md), we are using environment variables to make authorized requests to NYPL's API platform. In order to be secure, we are encrypting and decrypting those environment variables using AWS KMS. Please get these variables from someone in the NYPL Digital Department.
+
+| Variable | Description |
+| -------- | ----------- |
+| `KMS_ENV` | Determines whether to interpret ..CLIENT_ID and ..CLIENT_SECRET variables as "encrypted" or "unencrypted". Default "encrypted". | 
+| `PLATFORM_API_CLIENT_ID` | Platform client id. If KMS_ENV is "encrypted", this value must be encrypted. |
+| `PLATFORM_API_CLIENT_SECRET` | Platform client secret. If KMS_ENV is "encrypted", this value must be encrypted. |
+
+### Encrypting
```
charmingduchess, line 43: have you met my friend [kms-util](https://github.com/NYPL-discovery/kms-util)?

src/server/nyplApiClient/index.ts
```diff
@@ -0,0 +1,128 @@
...
+  api: {
+    platform: {
+      development:
+        process.env.PLATFORM_API_BASE_URL ||
+        "https://qa-platform.nypl.org/api/v0.1",
+      production:
+        process.env.PLATFORM_API_BASE_URL ||
+        "https://platform.nypl.org/api/v0.1",
+    },
+    // The 'discovery' base URL should use DISCOVERY_API_BASE_URL if set,
```
charmingduchess, line 20: What is the difference between DISCOVERY_API_BASE_URL and PLATFORM_API_BASE_URL?

charmingduchess: I'm unsure if we should have a fallback here. In the case that we are missing such a crucial env var, I think it would be better to have the app break than to mistakenly have the production serving QA data.

src/server/nyplApiClient/index.ts
```diff
@@ -0,0 +1,128 @@
...
+const appEnvironment = process.env.APP_ENV || "production"
+const kmsEnvironment = process.env.KMS_ENV || "encrypted"
+let decryptKMS
+let kms
+
+if (kmsEnvironment === "encrypted") {
+  kms = new aws.KMS({
+    region: "us-east-1",
+  })
+  // console.log({kms})
```
charmingduchess, line 52: rm comment

src/server/nyplApiClient/index.ts
```diff
@@ -0,0 +1,128 @@
...
+          reject(err)
+        } else {
+          resolve(Plaintext.toString())
+        }
+      })
+    })
+  }
+}
+
+const clientId = process.env.clientId || process.env.PLATFORM_API_CLIENT_ID
```
charmingduchess, line 74: Why do we want to support both formats of clientId/PLATFORM_API_CLIENT_ID?

ENVIRONMENTVARS.md
dgcohen: Good to know about kms-util! Not sure if we need to support unencrypted vars, unsure if there's a historical reason for this but happy to remove the support for unencrypted vars if everyone agrees.

src/server/nyplApiClient/index.ts
dgcohen: That's a good point, again unsure if there are historical reasons for that but happy to consolidate the team agrees it's not necessary anymore.

src/server/nyplApiClient/index.ts
dgcohen: Good catch, we can probably deprecate that.

ENVIRONMENTVARS.md
EdwinGuzman: Wow `kms-utils` is very useful. Should that be moved to the `NYPL` Github org (eventually)?

My initial thought is to support unencrypted env vars for local development if a dev doesn't have AWS set up locally or have the encrypted vars. But this doesn't really make sense and it seems fair to say that a dev working on this app _must_ have AWS access to nypl-digital-dev and be logged in locally. Thoughts?

src/server/nyplApiClient/index.ts
EdwinGuzman: This was legacy. You're right and now would be a good chance to just use `PLATFORM_API_CLIENT_ID`.

src/server/nyplApiClient/index.ts
EdwinGuzman: Context is in this PR: https://github.com/NYPL/discovery-front-end/pull/2109

charmingduchess (commented):
Looks good! My only note is that you can probably simplify things by eliminating support for so many vars - clientId/PLATFORM_CLIENT_ID, discovery vs. platform api's (they are the same base url).

src/server/nyplApiClient/index.ts
charmingduchess: ah ok i didnt realize that this was mostly just what was already happening in the DFE. i must not have the most recent version on my local. makes sense!

ENVIRONMENTVARS.md
charmingduchess: I agree with that, Edwin! Supporting unencrypted vars also runs a higher risk of accidentally committing secrets.

ENVIRONMENTVARS.md
```diff
@@ -0,0 +1,59 @@
...
+
+General environment variables are declared in the `.env.example` file. A copy of this file should be made and saved as `.env.local` where real values should be added.
+
+Generally, environment variables are meant to be read through the `process.env` object _on the server_. Variables intended for use on the client side should be prefaced with NEXT_PUBLIC_ per Next's [docs](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables).
+
+If an environment variable is updated, make sure to restart the server for the application to pick up the new value.
+
+## Server and Application-wide Variables
+
+These variables are used to configure server settings and application-wide settings.
```
EdwinGuzman, line 23: Is this section suppose to be empty for now?

ENVIRONMENTVARS.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Application Variables
```
EdwinGuzman, line 25: Seems like "Application-wide" and "Application" variables are very similar or the same. Maybe consolidate into one section.

ENVIRONMENTVARS.md
```diff
@@ -0,0 +1,59 @@
...
+
+In order to encrypt, please download and install the `aws` [cli tool](https://aws.amazon.com/cli/). The command to encrypt is
+
+    aws kms encrypt --key-id [your IAM key from AWS] --plaintext [value to encrypt] --output text --query CiphertextBlob
+
+The `aws kms encrypt` commands returns and object with a `CiphertextBlob` property. Since we only want that value, we use the `--query` flag to retrieve just that. This value can be copied and pasted into the AWS EBS configuration in the UI for the app's environment.
+
+More information can be found in the [encrypt docs](http://docs.aws.amazon.com/cli/latest/reference/kms/encrypt.html).
+
+NOTE: This value is base64 encoded, so when decoding make sure to decode using base64.
```
EdwinGuzman, line 55: This entire section is good to keep, but add an "Alternatively, you can use the kms-utils helper package...."

EdwinGuzman (commented):
Since this is being changed to only use encrypted variables, some of the relevant text also needs an update.

ENVIRONMENTVARS.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Encrypting
+
+Two variables are assumed encrypted when `KMS_ENV` is "encrypted": `PLATFORM_API_CLIENT_ID` and `PLATFORM_API_CLIENT_SECRET`. We need these variables to create an instance of the `nypl-data-api-client` npm package and make authorized requests to the NYPL Digital API endpoints. This is needed for the Discovery UI app to make requests itself to the APIs.
+
+In order to encrypt, please download and install the `aws` [cli tool](https://aws.amazon.com/cli/). The command to encrypt is
+
+    aws kms encrypt --key-id [your IAM key from AWS] --plaintext [value to encrypt] --output text --query CiphertextBlob
+
+The `aws kms encrypt` commands returns and object with a `CiphertextBlob` property. Since we only want that value, we use the `--query` flag to retrieve just that. This value can be copied and pasted into the AWS EBS configuration in the UI for the app's environment.
```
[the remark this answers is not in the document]
dgcohen: Good catch, will fix

ENVIRONMENTVARS.md
dgcohen: Good call, fixed.

ENVIRONMENTVARS.md
dgcohen: Adding a feature-flag section is a great idea. Gonna wait to add til we add that functionality.

ENVIRONMENTVARS.md
dgcohen: Good call! removed the ability to use unencrypted credentials and updated the docs.

src/server/nyplApiClient/index.ts
dgcohen: Just removed clientId.

ENVIRONMENTVARS.md
dgcohen: Nope, consolidated to just Application Variables

ENVIRONMENTVARS.md
```diff
@@ -0,0 +1,59 @@
...
+
+## AWS Elastic Beanstalk Environment Variables
+
+As previously mentioned in the [README](README.md), we are using environment variables to make authorized requests to NYPL's API platform. In order to be secure, we are encrypting and decrypting those environment variables using AWS KMS. Please get these variables from someone in the NYPL Digital Department.
```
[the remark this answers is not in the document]
dgcohen: Good call, done.

ENVIRONMENTVARS.md
```diff
@@ -0,0 +1,54 @@
...
+
+In order to encrypt, please download and install the `aws` [cli tool](https://aws.amazon.com/cli/). The command to encrypt is
+
+    aws kms encrypt --key-id [your IAM key from AWS] --plaintext [value to encrypt] --output text --query CiphertextBlob
+
+The `aws kms encrypt` commands returns and object with a `CiphertextBlob` property. Since we only want that value, we use the `--query` flag to retrieve just that. This value can be copied and pasted into the AWS ECS configuration in the UI for the app's environment.
+
+More information can be found in the [encrypt docs](http://docs.aws.amazon.com/cli/latest/reference/kms/encrypt.html).
+
+Alternatively, you can use the [kms-util](https://github.com/NYPL-discovery/kms-util) helper package."
```
EdwinGuzman, line 48: Minor: extra `"` at the end.

EdwinGuzman: Not for right now but this should be expanded with more information on `kms-util` and an example usage. We're saying alternatively that this package can be used but not why or how or differences between the `aws kms ...` script.

src/server/nyplApiClient/index.ts
```diff
@@ -0,0 +1,107 @@
+import NyplApiClient from "@nypl/nypl-data-api-client"
+import aws from "aws-sdk"
+
+interface KMSCache {
+  clients: string[]
+  clientId: string
+  clientSecret: string
+}
+
+const config: any = {
```
EdwinGuzman, line 10: Eventually this should be moved to a separate config file.

src/server/nyplApiClient/index.ts
```diff
@@ -0,0 +1,107 @@
...
+      }
+    })
+  })
+}
+
+const clientId = process.env.PLATFORM_API_CLIENT_ID
+const clientSecret = process.env.PLATFORM_API_CLIENT_SECRET
+
+const keys = [clientId, clientSecret]
+const CACHE: KMSCache = { clients: [], clientSecret, clientId }
```
EdwinGuzman, line 73: Super nitpick! (so safe to ignore)

I like placing const variables together and functions afterwards (since they tend to depend on variables defined before). In this case, these can be moved up near line 43 since that's another instance of defining a variable based on the env var.

EdwinGuzman (approved):
One minor typo but good otherwise.

src/server/nyplApiClient/index.ts
dgcohen: Good call! done.

src/server/nyplApiClient/index.ts
dgcohen: Just moved it to a config file. Let me know if this naming/placement makes sense.

PR #8, opened 2023-08-01
add travis.yml

shuynh:


shuynh:
> Generally, this looks fine to me. I think you mentioned you will need to test a bit more to get it right which is fine. This app is not and won't be used in qa or production for a while so testing is completely okay.
> 
> Just curious, did you make this changes directly into the `qa` branch and then merging back to `main`? It's not a problem but we're want a flow of `pr-branch` gets merged into `main`, `main` gets merged into `qa`, and `qa` gets merged into `production`.

soo there wasnt a `qa` branch in this repo before so i made one and then just pushed the whole thing up in one go. i agree that should be the flow, we should def make the `main`, `qa`, and `production` branches protected on github to require a PR.

PR #11, opened 2023-08-11
SCC 3600 - Initialize Search Results components with results display and refactor API fetching

dgcohen:
This PR adds the initial scaffolding for search results display, as well as some small changes to the API routes that facilitate this. Since the initial [Search Results ticket](https://jira.nypl.org/browse/SCC-3600) was not granular enough, we have created tickets for subcomponents and additional search results functionality. The acceptance criteria for this ticket has been modified to only include the ability to perform a basic keyword search and see a list of titles on the results page.

Changes/additions in this PR:
1. The initial scaffolding for the search index page
2. Search results data fetched in the results page via getServerSideProps which automatically handles client-side refetching on route changes, which means I was able to get rid of the default API route handler and move the fetching util function to /src/server
3. Replacement of the success callback in the fetchResults function with a Promise that returns the results on success.
4. Renaming the Search component to SearchForm
5. BASE_URL has now been added as a constant since the Next config value is not accessible on the client side.
6. Dummy display of result titles and "no results" messaging that will be fleshed out in further tickets.


EdwinGuzman (approved):
Peer reviewed with Vera

PR #16, opened 2023-08-16
Scc 3606/drbb server side fetch

dgcohen:


src/config/config.ts
```diff
@@ -21,12 +21,12 @@ export const apiConfig = {
...
         "https://platform.nypl.org/api/v0.1",
     },
-    drbb: {
+    drb: {
       development:
         process.env.DRB_API_BASE_URL || "http://drb-api-qa.nypl.org/search/",
       production:
         process.env.DRB_API_BASE_URL ||
-        "https://digital-research-books-api.nypl.org/v3/sfr/search",
+        "https://digital-research-books-api.nypl.org/search",
```
EdwinGuzman, line 29: Who verified this is the stable version?

dgcohen: This url is what's being used on production currently. The old fallback url is for v3.

src/utils/drbUtils.ts
```diff
@@ -0,0 +1,151 @@
...
+    })
+    .filter((str) => str)
+    .join(",")
+}
+
+/**
+ *  Given a hash of SearchFilters, returns an array of DRBFilters as expected by the DRB API
+ */
+function mapSearchFiltersToDRBFilters(filters: SearchFilters = {}): DRBFilters {
+  let drbFilters = [] as DRBFilters
```
EdwinGuzman, line 40: Did you run into issues with the following syntax? `let drbFilters: DRBFilters = []`

EdwinGuzman (changes requested):
Looks and works fine overall. This could use some documentation for swr and how it works, maybe a small section in the readme or a separate md file. There's also a `discoveryApiResearchNowApiMapping.md` file in the current repo that provides interesting documentation but I don't know how accurate or up-to-date that is. @charmingduchess  and @danamansana can you take a glance and see if that's useful (or parts of it) to carry over?

Either way, a section or new md file explaining that fetching results from DRB is client side, why it is that way, and general overview will be helpful.

pages/search/index.tsx
```diff
@@ -18,21 +19,25 @@ import type { SearchResultsItem } from "../../src/types/searchTypes"
...
         <title>NYPL Research Catalog</title>
       </Head>
       {results?.results?.totalResults ? (
-        <>
-          <Heading level="three">
-            {`Displaying 1-50 of ${results.results.totalResults.toLocaleString()} results for keyword "${
-              searchParams.searchKeywords
-            }"`}
-          </Heading>
+        <div style={{ display: "flex" }}>
```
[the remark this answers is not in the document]
dgcohen: Agreed, this is intended to be temporary until the search results component is built out.

src/utils/drbUtils.ts
dgcohen: No issues with that, I believe it's just an alternative syntax. Replaced with the one you proposed since it looks a bit cleaner.

package.json
```diff
@@ -20,6 +20,7 @@
     "next": "12.3.4",
     "react": "17.0.2",
     "react-dom": "17.0.2",
+    "swr": "^2.2.1",
```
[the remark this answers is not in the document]
dgcohen: Good call, done.

pages/search/index.tsx
```diff
@@ -18,21 +19,25 @@ import type { SearchResultsItem } from "../../src/types/searchTypes"
...
 export default function Search({ results }) {
   const { query } = useRouter()
+
   const searchParams = mapQueryToSearchParams(query)
 
+  // Remove page and identifiers fields from drbParams to prevent re-fetches
+  const drbParams = { ...searchParams, page: undefined, identifiers: undefined }
+
   return (
     <div style={{ paddingBottom: "var(--nypl-space-l)" }}>
```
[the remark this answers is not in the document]
dgcohen: Good call, removed all instances of wrapper divs with bottom padding and instead wrapped the main content in Layout with a <Box pb="l">

src/components/Layout/Layout.tsx
```diff
@@ -50,7 +51,7 @@ const Layout = ({ children, activePage }: LayoutProps) => {
             </>
           )
         }
-        contentPrimary={children}
+        contentPrimary={<Box pb="l">{children}</Box>}
```
EdwinGuzman, line 54: 👍

EdwinGuzman:
Latest updates look good. There's just the comment about the `discoveryApiResearchNowApiMapping.md` file and whether that is useful to bring in or not.

dgcohen:
> Latest updates look good. There's just the comment about the `discoveryApiResearchNowApiMapping.md` file and whether that is useful to bring in or not.

I think it makes sense to have documentation around the expected response structure. @charmingduchess and @danamansana can you weigh in on whether we should keep them, and if they look correct?

discoveryApiResearchNowMapping.md
```diff
@@ -0,0 +1,478 @@
...
+
+Single quotes (') are used for frontend terminology. Italics are used to describe frontend features. Code styling (
+e.g. `filters`) is used for the parameters the respective APIs except.
+
+| Research Catalog front end                    | Discovery API                                         | ResearchNow API                                 |
+|-----------------------------------------------| ----------------------------------------------------- | ----------------------------------------------- |
+| _Search field_                                | `q`                                                   | `queries`, "value"                              |
+| _Search field dropdown_                       | `search_scope`                                        | `queries`, "field"                              |
+| <ul>'All Fields'                              | <ul>"all"                                             | <ul>"keyword"                                   |
+| <ul>'Title'                                   | <ul>"title"                                           | <ul>"title"                                     |
```
EdwinGuzman, line 125: These `ul`s can be better formatted to be `li`s and have closing tags but this is not a deal breaker to merge this PR in right now.

EdwinGuzman (approved):

charmingduchess:
Sorry for delayed review! Here's a couple notes:

> NB: The results were previously being loaded on the server side alongside search results (preventing the display of search results til DRB request completed or timed out after 5 seconds). For now we are going to change this to just load on the client side to avoid blocking search results or adding the timeout (approved by Sean), but we can re-implement to the previous solution if the client-loading is disruptive to the user experience.
- The 5 second timeout was a fix for a development kink. When running locally, the connection to that api doesn't work. It sometimes would lead to endless loading layer. I don't remember exactly the conditions that created the problem, though. It was definitely a clunky fix, but made development much smoother. It doesn't have to be implemented the exact same way, but I think we do need the stop gap (unless something change and we now get drb [?] api results locally).

> There's also a discoveryApiResearchNowApiMapping.md file in the current repo that provides interesting documentation but I don't know how accurate or up-to-date that is
- Wow I've literally never in my life looked at this file before! I don't personally see the use in having api responses documented in the app. It's also weird to me that the documentation focuses equally on discovery api and research now. I think this file could be removed in favor of a couple of lines in README calling out:
  -  multiple requests are made
  - what the endpoints are
  - links to github repo's for research now and dapi.

EdwinGuzman:
"Research Now" was the previous name for what is now DRB.

> The 5 second timeout was a fix for a development kink. 

Can we remove it and add it back only if we encounter any issues?

>  I think this file could be removed in favor of a couple of lines in README calling out:
when and why these tandem (?) requests are made
what the endpoints are
links to github repo's for research now and dapi

This sounds good. Can we make this a separate documentation ticket? It'd be great if we can get a DRB engineer to review the documentation separately, so why it should be a separate ticket.

charmingduchess:
> Can we remove it and add it back only if we encounter any issues?
That's fine by me.

> Can we make this a separate documentation ticket? 
I guess it could be? I'm really imagining something very simple and small.

pages/api/drb.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Getter function for DRB results. Accepts a SearchParams object as an argument.
+ */
+export async function fetchDRBResults(
+  searchParams: SearchParams
+): Promise<DRBResultsResponse | Error> {
+  const drbQueryString = getDRBQueryStringFromSearchParams(searchParams)
+
+  return nyplApiClient({ apiName: DRB_API_NAME })
+    .then((client) => client.get(drbQueryString))
```
charmingduchess, line 19: Apologies if there is already a bunch of code written in .then() format, but can we opt for async await for this entire repo?

dgcohen: Yeah that's cool with me. Will fix in a bit.

dgcohen: @charmingduchess I replaced the then statements with async/await in both drb.ts and search.ts, which cleaned it up substantially.

However, I think the Promise.all in search.ts looks a little cleaner with the .then syntax. How would you feel about keeping it in this case?

Screenshot 2023-09-18 at 12 12 04 PM

Screenshot 2023-09-18 at 12 11 40 PM

EdwinGuzman: Is something missing from the second screenshot? Such as the `nyplApiClient` call and where `client` comes from?

dgcohen: Yes, forgot to include that line in the screenshot, here it is with it:

Screenshot 2023-09-18 at 7 14 24 PM

EdwinGuzman: The more I look at it the more I kinda like the second option. Wouldn't it work with a `Promise.resolve(client.get(...))`? So you don't have to create the `new Promise(...)`?

charmingduchess: I like the sound of what Edwin is proposing. I'd rather not mix async await with .then's. I'm also a fan of this pattern, if it interests you (you may or may not need the async/await keywords here... not 100% on that syntax):
`Promise.all([resultsQuery, aggregationsQuery].map(async query => await client.get(${route}${query})`

But moreover, if the syntax of the first one works where you're awaiting promises, would this not work as well:
`client = await nyplClient()`
`Promise.all([await client.get(resultsPath), await client.get(aggsPath)]`

dgcohen: Good call @charmingduchess ! Replaced with the syntax you suggestd in your last comment.

PR #17, opened 2023-08-23
SCC-3713 - Search Results Bib model

dgcohen:
This PR introduces the SearchResultsBib class, which handles the formatting and other utility functions for displaying search results. The majority of the logic is extracted from the ResultsList component in DFE.

The scope of this work does not include attributes pertaining to items or the items table in general, as this will be handled in a separate PR.

It also technically doesn't include the rendering of the Search Result data, although I've modified the temporary search results component to make use of the new SearchResultsBib model.

### DIFF
--- a/pages/search/index.tsx
+++ b/pages/search/index.tsx
@@ -30,7 +30,7 @@ export default function Search({ results }) {
     }
     return new SearchResultsBib(result?.result)
   })
-
+  console.log(searchResultBibs)
   console.log(results?.results?.itemListElement)
 
   return (
--- a/pages/search/index.tsx
+++ b/pages/search/index.tsx
@@ -11,6 +11,7 @@ import { fetchResults } from "../api/search"
 import { mapQueryToSearchParams } from "../../src/utils/searchUtils"
 import RCLink from "../../src/components/RCLink/RCLink"
 import type { SearchResultsItem } from "../../src/types/searchTypes"
+import SearchResultsBib from "../../src/models/SearchResultsBib"
 
 /**
  * The Search page is responsible for fetching and displaying the Search results,
@@ -20,12 +21,17 @@ export default function Search({ results }) {
   const { query } = useRouter()
   const searchParams = mapQueryToSearchParams(query)
 
+  const searchResultBibs = results?.results?.itemListElement.map(
+    (result: SearchResultsItem) => new SearchResultsBib(result?.result)
+  )
+  const totalResults: number = results?.results?.totalResults
+
   return (
     <div style={{ paddingBottom: "var(--nypl-space-l)" }}>
       <Head>
         <title>NYPL Research Catalog</title>
       </Head>
-      {results?.results?.totalResults ? (
+      {totalResults ? (
         <>
           <Heading level="three">
             {`Displaying 1-50 of ${results.results.totalResults.toLocaleString()} results for keyword "${
@@ -33,20 +39,16 @@ export default function Search({ results }) {
             }"`}
           </Heading>
           <SimpleGrid columns={1} gap="grid.m">
-            {results.results.itemListElement.map(
-              (result: SearchResultsItem) => {
-                // TODO: Create SearchResult component to manage result display (https://jira.nypl.org/browse/SCC-3714)
-                return (
-                  <Card key={result.result["@id"]}>
-                    <CardHeading level="four">
-                      <RCLink href="/bib">
-                        {result.result["titleDisplay"][0]}
-                      </RCLink>
-                    </CardHeading>
-                  </Card>
-                )
-              }
-            )}
+            {searchResultBibs.map((bib: SearchResultsBib) => {
+              // TODO: Create SearchResult component to manage result display (https://jira.nypl.org/browse/SCC-3714)
+              return (
+                <Card key={bib.id}>
+                  <CardHeading level="four">
+                    <RCLink href="/bib">{bib.title}</RCLink>
+                  </CardHeading>
+                </Card>
+              )
+            })}
           </SimpleGrid>
         </>
       ) : (
--- a/pages/search/index.tsx
+++ b/pages/search/index.tsx
@@ -6,7 +6,7 @@ import {
   SimpleGrid,
 } from "@nypl/design-system-react-components"
 import { useRouter } from "next/router"
-import { isEmpty as _isEmpty } from "underscore"
+import { isEmpty } from "underscore"
 
 import { fetchResults } from "../api/search"
 import { mapQueryToSearchParams } from "../../src/utils/searchUtils"
@@ -25,7 +25,7 @@ export default function Search({ results }) {
   const { itemListElement, totalResults } = results.results
 
   const searchResultBibs = itemListElement.map((result: SearchResultsItem) => {
-    if (_isEmpty(result) || (result.result && _isEmpty(result.result))) {
+    if (isEmpty(result) || (result.result && isEmpty(result.result))) {
       return null
     }
     return new SearchResultsBib(result.result)
--- a/pages/search/index.tsx
+++ b/pages/search/index.tsx
@@ -6,6 +6,7 @@ import {
   SimpleGrid,
 } from "@nypl/design-system-react-components"
 import { useRouter } from "next/router"
+import { isEmpty as _isEmpty } from "underscore"
 
 import { fetchResults } from "../api/search"
 import { mapQueryToSearchParams } from "../../src/utils/searchUtils"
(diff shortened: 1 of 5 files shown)


pages/search/index.tsx
```diff
@@ -6,11 +6,13 @@ import {
   SimpleGrid,
 } from "@nypl/design-system-react-components"
 import { useRouter } from "next/router"
+import { isEmpty as _isEmpty } from "underscore"
```
EdwinGuzman, line 9: I _think_ I started this convention so we'd know where these functions were coming from explicitly. It doesn't have to be this way now and I wonder if it is actually helpful or not. I'd be happy with whatever is chosen but just want to point out that we don't have to strictly stick to some conventions if they don't make sense anymore.

pages/search/index.tsx
```diff
@@ -20,33 +22,38 @@ export default function Search({ results }) {
   const { query } = useRouter()
   const searchParams = mapQueryToSearchParams(query)
 
+  const { itemListElement, totalResults } = results.results
+
+  const searchResultBibs = itemListElement.map((result: SearchResultsItem) => {
+    if (_isEmpty(result) || (result.result && _isEmpty(result.result))) {
+      return null
+    }
+    return new SearchResultsBib(result?.result)
```
EdwinGuzman, line 31: Does typescript give you a warning here if `?` is not used? I guess it can't tell if the previous logic check failed or not.

pages/search/index.tsx
```diff
@@ -20,33 +22,38 @@ export default function Search({ results }) {
   const { query } = useRouter()
   const searchParams = mapQueryToSearchParams(query)
 
+  const { itemListElement, totalResults } = results.results
+
+  const searchResultBibs = itemListElement.map((result: SearchResultsItem) => {
```
EdwinGuzman, line 27: Try this: manually make `itemListElement=[{}, {}, {}]` or any variation with undefined or null. Because this is a `map` and is returning `null` for the first case, the `searchResultsBibs` variable can turn out to be `[undefined, { ...}, { ... }, undefined, ...]` if the returned results from the API are wrong (empty or null or undefined).

In this case a forEach might be better.

src/models/SearchResultsBib.ts
```diff
@@ -0,0 +1,81 @@
+import type { SearchResult } from "../types/searchTypes"
+import type { ElectronicResource } from "../types/bibTypes"
+
+export default class SearchResultsBib {
```
EdwinGuzman, line 4: The class name is pretty descriptive already but can you add a comment overview about the changes. It's a bit hard because some reasons for the business logic, such as taking the first item in the `publicationStatement` array, isn't documented anywhere (as far as I know). That may be a good question for Brooke or Sean?

EdwinGuzman (commented):
Looks good! Cleaner to get the data through this object and not mix jsx with the business logic. In the PR description, to keep track of where the code is being refactored from, can you state the files? I think it's mostly the `ResultsList` component, right?

src/models/SearchResultsBib.ts
```diff
@@ -0,0 +1,81 @@
...
+  id: string
+  title: string
+  yearPublished?: string
+  materialType?: string
+  publicationStatement?: string
+  eResources?: ElectronicResource[]
```
charmingduchess, line 10: Can this be `electronicResources`? I think it would be a good idea to match these property names exactly to the discovery api properties.

src/models/SearchResultsBib.ts
charmingduchess: the majority if not all properties that only refer to the first element follow this pattern:
```
 "dateEndString": [
    "1990"
  ],
```
where there is a single-element array (due to JSON LD and or Elastic Search requirements). If there is a place where the expected interface of the api response is defined, it may be worth creating a type that denotes a single element array.

src/models/SearchResultsBib.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  get numElectronicResources() {
+    return this.eResources?.length || 0
+  }
+
+  get hasPhysicalItems() {
+    return this.numPhysicalItems > 0
+  }
+
+  get numItems() {
```
charmingduchess, line 37: would love to discuss this definition. we've ended up in some snarls in the past due to confusion about what constitutes the number of items. In this definition, num electronic resources will not be a part of the count of items unless there are NO physical items. That is helpful for determining whether to display the items table, but not helpful for displaying item counts.
If this numItems property exists to determine whether we render the items table or not, then maybe we could create a separate property (thats a big question mark). 
Lets circle back on this one when I'm back from vacation. There is potentially some product questions here as well regarding full digital representations of works vs supplemental content like table of contents. 
If we really iron this out and document i hope we can avoid bringing legacy headaches into this new repo!

src/models/SearchResultsBib.ts
```diff
@@ -0,0 +1,81 @@
...
+  constructor(result: SearchResult) {
+    this.id = result["@id"] ? result["@id"].substring(4) : ""
+    this.title = this.getTitleFromResult(result)
+    this.yearPublished = this.getYearFromResult(result)
+    this.materialType = result.materialType[0]?.prefLabel || null
+    this.publicationStatement = result.publicationStatement?.length
+      ? result.publicationStatement[0]
+      : null
+    this.eResources = result.electronicResources || null
+    this.numPhysicalItems = result.numItemsTotal || 0
```
charmingduchess, line 22: same idea as [this comment](https://github.com/NYPL/research-catalog/pull/17/files#r1313098518). lets double down on what these item counts are and mean.

src/models/SearchResultsBib.ts
EdwinGuzman: I don't think this is possible in typescript or may involve some intricate type definitions. Worth looking into but we could define the pattern in a general "how the API properties work"?

src/models/SearchResultsBib.ts
```diff
@@ -0,0 +1,81 @@
...
+          : ""
+      return result.title && result.title.length
+        ? `${result.title[0]}${author}`
+        : ""
+    }
+    return result.titleDisplay[0]
+  }
+
+  getYearFromResult(result: SearchResult) {
+    const { dateStartYear, dateEndYear } = result
```
charmingduchess, line 67: This could use a confirmation with Sean/Brooke. I know that this is a tricky area that sometimes leads to weird results.

src/types/searchTypes.ts
charmingduchess, line 1: I see electronicResources type on the inner SearchResults interface. Why does that need to be passed to the SearchResultsItem interface? Is SearchResultsItem supposed to be just an element in the Search Results or is it technically an item in the bib/item sense?

pages/search/index.tsx
dgcohen: Thanks! I figured that was the reason for the pattern, I think I'd prefer to use the default naming if there is no naming clash, but happy to keep the aliasing convention if anyone prefers it.

pages/search/index.tsx
dgcohen: There's no typescript warning since the if statement above catches the nullish value. Added it out of force of habit but will remove since that condition is unreachable.

pages/search/index.tsx
dgcohen: Good catch, I moved the null check into a filter before the map.

src/models/SearchResultsBib.ts
dgcohen: Good call! Renamed to electronicResources

src/types/searchTypes.ts
dgcohen: SearchResultsItem refers to a search result in the itemListElement array parameter of the Search Results API response (i.e. a raw search result from the API). I named it that due to the inclusion of the word "item" in itemListElement, but we may want to consider renaming it to avoid confusing it with bib "items".

charmingduchess: Let's come up with a new name and keep the term 'item' only for components that directly deal with item level rendering. Could use `searchResultElement` and `searchResults`?

dgcohen: Good call! done

src/models/SearchResultsBib.ts
dgcohen: I added a comment above SearchResultsBib with some context as well as a note about extracting single-item arrays in the constructor. I think it makes to have some data formatting in the class constructor, so not sure it's necessary to add a section to the main docs about this, but happy to add one if you think it would be helpful to have a high-level overview of this pattern.

EdwinGuzman (approved):
This is a good first step for the model and I'm sure we'll have to revisit it later when more properties are added and use cases for the UI. I think we can work with this now while we keep building out the app.

PR #19, opened 2023-09-07
SCC-3618 - 404 page

dgcohen:
This PR adds the 404 page template and logic to hide the Research Catalog header on 404 pages.

Unrelated to the ticket, but I also changed the localhost port to 8080 temporarily to avoid conflicts with port 3000 on DFE for convenience.

https://jira.nypl.org/browse/SCC-3618

### DIFF
--- a/pages/404/redirect.tsx
+++ b/pages/404/redirect.tsx
@@ -1,0 +1,32 @@
+import Head from "next/head"
+import { Heading } from "@nypl/design-system-react-components"
+
+import RCLink from "../../src/components/RCLink/RCLink"
+import {
+  CIRCULATING_CATALOG_URL,
+  LEGACY_CATALOG_URL,
+} from "../../src/config/constants"
+
+export default function Redirect404() {
+  return (
+    <div style={{ paddingBottom: "var(--nypl-space-l)" }}>
+      <Head>
+        <title>404 Not Found</title>
+      </Head>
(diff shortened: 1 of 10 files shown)


pages/404/redirect.tsx
```diff
@@ -0,0 +1,32 @@
+import Head from "next/head"
```
EdwinGuzman, line 1: I can't access this page directly but I should be able to, right?

src/components/Layout/Layout.tsx
```diff
@@ -23,29 +23,32 @@ interface LayoutProps {
...
-                },
-              ]}
-            />
-            <div className={styles.researchHeadingContainer}>
-              <Heading id="heading-h1" level="one" text="Research Catalog" />
-              <SubNav activePage={activePage} />
-              {showSearch && <SearchForm />}
-            </div>
-          </>
+          showHeader && (
```
EdwinGuzman, line 32: The 404 page could use a new design 😭  this is fine right now but hopefully we can get something better than just text.

pages/404/redirect.tsx
dgcohen: You should be able to see it at http://localhost:8080/research/research-catalog/404/redirect

EdwinGuzman: Yeah. I think it was cached or something but now I do see it. Thanks.

EdwinGuzman (approved):

PR #20, opened 2023-09-14
Header URL env var update

EdwinGuzman:
It seems the travis var name was not consistent with what docker and the app expected the env var to be so the Header and Footer never loaded in the AWS QA instance. This should fix that.


shuynh:
👍 👍 👍

shuynh (approved):

PR #24, opened 2023-09-15
SCC-3725 - Advanced Search UI

dgcohen:
This PR adds the Advanced Search page in addition to it's UI, utilities that pertain to the UI, and it's submit handler.

I've recreated some of the necessary styling overrides of the old App (primarily on the buttons), but I've kept this as close to the design system defaults in order for Apoorva to make the call whether we should override things or fix via the design system.

I'm currently using useReducer to set overall form state in order to build the query string before submission and to clear the form. I would ideally like the form fields to be uncontrolled to prevent re-renders on every keystroke when editing fields but I believe this is currently not possible with the TextInput component.

NB: The material type filters currently do not get parsed correctly in searches, but this will be handled in a separate ticket.


pages/search/advanced.tsx
```diff
@@ -0,0 +1,209 @@
...
+                />
+              </FormField>
+            </FormRow>
+            <Label htmlFor="formats" id="formats-label">
+              Format
+            </Label>
+            <Fieldset id="formats" className={styles.formatFields}>
+              <FormRow>
+                <FormField className={styles.formField}>
+                  {materialTypeOptions.slice(0, 4).map((materialType) => {
```
EdwinGuzman, line 157: Try `CheckboxGroup`.

pages/search/advanced.tsx
```diff
@@ -0,0 +1,209 @@
...
+                        key={materialType.value}
+                        labelText={materialType.label}
+                        name={materialType.value}
+                        value={materialType.value}
+                      />
+                    )
+                  })}
+                </FormField>
+                <FormField className={styles.formField}>
+                  {materialTypeOptions.slice(4).map((materialType) => {
```
EdwinGuzman, line 171: Reduce magic numbers "4". Perhaps include also an explanation of why only four from PM.

pages/search/advanced.tsx
```diff
@@ -0,0 +1,209 @@
...
+      input.checked = false
+    })
+
+    Array.from(document.getElementsByTagName("select")).forEach((select) => {
+      select.value = ""
+    })
+  }
+
+  return (
+    <div
```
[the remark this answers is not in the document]
dgcohen: Yeah this will be fixed when the other PR is merged in

pages/search/advanced.tsx
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    const queryString = getQueryString(searchParams)
+    await router.push(`/search/?${queryString}`)
+  }
+
+  const handleClear = (e: SyntheticEvent) => {
+    // clear text inputs
+    e.preventDefault()
+    Array.from(document.getElementsByTagName("input")).forEach((input) => {
```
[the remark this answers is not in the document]
dgcohen: Clearing is working now that that useReducer has been implemented

pages/search/advanced.tsx
```diff
@@ -0,0 +1,209 @@
...
+      </Head>
+      <Heading level="two">Advanced Search</Heading>
+      <Form
+        id="advancedSearchForm"
+        className={styles.advancedSearchForm}
+        method="post"
+        action={`${BASE_URL}/search`}
+        onSubmit={handleSubmit}
+      >
+        <SimpleGrid columns={2} gap="grid.m">
```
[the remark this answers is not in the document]
dgcohen: I've replaced SimpleGrid and Fieldset with FormRow and FormField, the spacing has increased someone but we can figure out how we want to address that in visual QA.

EdwinGuzman (commented):
Not directly related but I made a search for "title" = "Harry Potter" and got this error:
Screen Shot 2023-09-19 at 9 57 52 AM

pages/search/advanced.tsx
```diff
@@ -0,0 +1,209 @@
...
+              {languageOptions.map((language) => {
+                return (
+                  <option value={language.value} key={language.value}>
+                    {language.label}
+                  </option>
+                )
+              })}
+            </Select>
+          </Fieldset>
+          <Fieldset id="advancedSearchRight">
```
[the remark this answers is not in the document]
dgcohen: Great to know! Fixed.

pages/search/advanced.tsx
```diff
@@ -0,0 +1,209 @@
...
+            <Label htmlFor="dates" id="dates-label">
+              Date
+            </Label>
+            <FormRow id="dates" className={styles.dateFields}>
+              <FormField className={styles.formField}>
```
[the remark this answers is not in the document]
dgcohen: Yes! Replaced SimpleGrid and Fieldset with FormRow and FormField.

pages/search/advanced.tsx
```diff
@@ -0,0 +1,209 @@
...
+                  labelText="To"
+                  helperText="e.g. 2000"
+                  initialDate=""
+                />
+              </FormField>
+            </FormRow>
+            <Label htmlFor="formats" id="formats-label">
+              Format
+            </Label>
+            <Fieldset id="formats" className={styles.formatFields}>
```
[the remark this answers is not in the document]
dgcohen: Moved the label to CheckboxGroup's labelText prop

pages/search/advanced.tsx
dgcohen: Done!

dgcohen:
> Not directly related but I made a search for "title" = "Harry Potter" and got this error: Screen Shot 2023-09-19 at 9 57 52 AM

Good catch! Added a length check on that prop in the constructor.

pages/search/advanced.tsx
```diff
@@ -0,0 +1,225 @@
...
+                  </option>
+                )
+              })}
+            </Select>
+          </FormField>
+          <FormField id="advancedSearchRight" gap="grid.s">
+            <Fieldset legendText="Date" id="dates">
+              <FormRow id="dates" gap="grid.m">
+                <FormField>
+                  <DatePicker
```
EdwinGuzman, line 160: Ah. I was wondering how `handleDateChange` was being used and just realized these are two DatePickers. Did you try making it a single DatePicker as a range type and find out that it didn't work for this situation?

pages/search/advanced.tsx
```diff
@@ -0,0 +1,225 @@
...
+                  />
+                </FormField>
+              </FormRow>
+            </Fieldset>
+            <CheckboxGroup
+              id="formats"
+              name="formats"
+              labelText="Formats"
+              onChange={handleCheckboxChange}
+              value={searchFormState["selectedFilters"].materialType}
```
EdwinGuzman, line 188: This should be `defaultValue` to be controlled. Then you can remove the `name` from the individual Checkbox component on line 197. For more info: https://nypl.github.io/nypl-design-system/reservoir/v1/?path=/docs/components-form-elements-checkboxgroup--docs

pages/search/advanced.tsx
dgcohen: No but I assumed that users might want to only select one date, rather than a range. Not sure if this is possible with a date range?

src/utils/advancedSearchUtils.ts
```diff
@@ -0,0 +1,38 @@
+import type { SearchParams, SearchFormInputField } from "../types/searchTypes"
+import { searchAggregations } from "../config/aggregations"
+
+export const textInputFields: SearchFormInputField[] = [
+  { name: "q", key: "searchKeywords", label: "Keywords" },
```
EdwinGuzman, line 5: I think this creates an issue in `advanced.tsx` line 133 where the value is `searchFormState[name]`. For this first object, the name is `"q"` but that doesn't exist in the `initialSearchFormState` object.

EdwinGuzman: I think making that name consistent will fix this issue:
Screen Shot 2023-09-20 at 11 47 32 AM

src/utils/advancedSearchUtils.ts
```diff
@@ -0,0 +1,38 @@
...
+  { name: "subject", key: "subject", label: "Subject" },
+]
+
+export const initialSearchFormState: SearchParams = {
+  searchKeywords: "",
+  title: "",
+  contributor: "",
+  subject: "",
+  selectedFilters: {
+    language: "",
```
EdwinGuzman, line 17: There's an issue selecting a language because there's no key:
Screen Shot 2023-09-20 at 11 34 23 AM

EdwinGuzman (commented):
Let's leave the gap between text input fields as the default and get a VQA from Apoorva and Marty. It actually seems correct it's just that now there is more space -- but it seems intentional from the UX/DS point of view.

pages/search/advanced.tsx
EdwinGuzman: hmm true. that's a good product question. Leave it as  is for now and we should get more info on this.

pages/search/advanced.tsx
dgcohen: I believe value has to be set in order to clear the values based on the form state.

dgcohen: I was able to remove the name on the individual checkboxes now that CheckboxGroup was added, however.

src/utils/advancedSearchUtils.ts
dgcohen: Good catch, fixed and improved the comment explaining this key lookup.

src/utils/advancedSearchUtils.ts
dgcohen: I might actually rename serachKeywords to q in SearchParams for consistency. This will eliminate the need for the key lookup.

src/reducers/searchFormReducer.ts
charmingduchess, line 1: Is this reducer an artifact from the original DFE? I'm not opposed to this pattern, just wondering. I don't exactly know where we left the conversation of context vs. redux in the end. If we are going to consolidate into using context over redux, maybe we want to start that now?

dgcohen: This is a new pattern suggested by Edwin to reduce the decrease the reliance on useState for updating the internal form state. This technically isn't redux and can't really be replicated by context, but happy to open the conversation on using useState instead of useReducer if you prefer it.

src/components/SearchForm/SearchForm.tsx
```diff
@@ -34,7 +34,7 @@ const SearchForm = () => {
           <SearchBar
             id="mainContent"
             action={`${BASE_URL}/search`}
-            method="get"
+            method="post"
```
charmingduchess, line 37: why is this method changing?

src/reducers/searchFormReducer.ts
charmingduchess: It doesn't bother me! Just wanted to know.

src/utils/appUtils.ts
```diff
@@ -7,8 +7,10 @@ import type { RCPage } from "../types/pageTypes"
  * conditionally rendering the Search.
  */
 export const getActivePage = (pathname: string): RCPage => {
-  if (pathname === "/" || pathname.includes("search")) {
+  if (pathname === "/" || pathname === "/search") {
```
charmingduchess, line 10: i like the specificity of === "/search" instead of includes("search"). Is there a reason the other pathnames are using includes instead of the === check?

dgcohen: The includes matches substrings in the case of sub paths (e.g.  /404/redirect) but we can definitely use exact matchers in cases where there are none (e.g. "/search/advanced")

src/components/SearchForm/SearchForm.tsx
dgcohen: I did it to have parity with the advanced search form (it's using a post request in the DFE app) but it might make sense to change them both to GET since it'll include the query string in the request.

dgcohen: Replaced both with a GET, think that makes more sense for search requests.

src/utils/appUtils.ts
dgcohen: I added a constant with the absolute paths, using those with exact matching which i think will be less flakey than .includes. We can add a regex later if we dealing with a bunch of sub paths but don't think we'll need to for this use case.

src/components/SearchForm/SearchForm.tsx
EdwinGuzman: Does that make it still work when js is turned off? I thought POST was the more appropriate one for the static (basic?) html form?

pages/search/advanced.tsx
```diff
@@ -0,0 +1,232 @@
...
+              })}
+            </CheckboxGroup>
+          </FormField>
+        </FormRow>
+        <HorizontalRule sx={{ margin: 0 }} />
+        <ButtonGroup
+          id="advancedSearchButtons"
+          buttonWidth="default"
+          sx={{
+            gap: "var(--nypl-space-xs)",
```
EdwinGuzman, line 212: I think just having `gap: "xs"` should work the same.

src/components/SearchForm/SearchForm.tsx
dgcohen: After taking a closer look, I think I understand why advanced search was using a post on the old app. Rather than using a query string to process the search, the request body parameters are used to build the redirect query string on the server side. I will change to a post (only on advanced search) for now and add this functionality in my advanced search utils PR)

PR #30, opened 2023-09-29
Homepage relative urls

EdwinGuzman:
Small update to get relative urls for NYPL.org links. This is to get the appropriate QA or production link for testing.

I could not use the `RCLink` component since next.js thinks a relative link is part of the app and injects the root URL. So instead just the DS `Link` is used so that the URL is relative to the domain and not the root path (/research/research-catalog).

This won't work locally or on Vercel but _should_ work on the QA site _after_ we do the reverse proxy. For example, 
`https://qa-research-catalog.nypl.org/about/divisions` won't work but when the app is reverse proxied onto qa-www.nypl.org, it'll work as intended.

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -5,6 +5,7 @@ import {
   Card,
   CardHeading,
   CardContent,
+  Link,
 } from "@nypl/design-system-react-components"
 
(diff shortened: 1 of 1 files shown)


dgcohen (approved):
Looks good! Just did something similar for DRB results.

PR #31, opened 2023-10-02
Scc 3773/utils testing

charmingduchess:
- Added tests for `appUtils`,`bibUtils, `drbUtils, `searchUtils`.
- created test helper that checks query strings for matching key=value pairs. This is so we are not stuck on order if the order changes with the new query param parsing library.
- Not 100% sure if it's necessary, but i added some logic into `getSearchQuery` to exclude the `sort_direction` parameter if none is supplied.
- nested `__test__` folder so we can include files that are not test suites in the `test` directory (looking forward to test data and helpers). jest returns a failure if a file within __test__ does not contain a test suite. Also can be changed in various ways in jest.config.js, but I thought this was the most easily grokked way to go about it.


src/utils/searchUtils.ts
```diff
@@ -21,9 +21,10 @@ import type {
 function getSortQuery(sortBy = "", order = ""): string {
   const reset = sortBy === "relevance"
   let sortQuery = ""
+  const sortDirectionQuery = order === "" ? "" : `&sort_direction=${order}`
 
   if (sortBy?.length && !reset) {
-    sortQuery = `&sort=${sortBy}&sort_direction=${order}`
+    sortQuery = `&sort=${sortBy}${sortDirectionQuery}`
```
EdwinGuzman, line 27: Should `sort=${...}` also be dealt the same way?

test/__test__/utils/drbUtils.test.tsx
```diff
@@ -0,0 +1,109 @@
+import { getDRBQueryStringFromSearchParams } from "../../../src/utils/drbUtils"
+import { DRB_RESULTS_PER_PAGE } from "../../../src/config/constants"
+import { queryParamsEquality } from "../../helpers/searchHelpers"
+
+const checkQueryParamsEquality = queryParamsEquality(
+  getDRBQueryStringFromSearchParams
+)
+
+describe("researchNowUtils", () => {
```
EdwinGuzman, line 9: This should be `drbUtils`.

test/__test__/utils/drbUtils.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+    })
+
+    it("should handle subject param", () => {
+      const testQuery = `?query=keyword%3A*%2Csubject%3Acorvids&page=1&source=catalog&size=${DRB_RESULTS_PER_PAGE}`
+      expect(checkQueryParamsEquality(testQuery, { subject: "corvids" })).toBe(
+        true
+      )
+    })
+
+    it("should combine query params", () => {
```
EdwinGuzman, line 98: The more I look at this file the more complex the query object vs the url query is. All seems fine but just pointing out the complexity.

EdwinGuzman (commented):
One minor naming comment. Otherwise, I'd prefer if the jest.config file can be updated to exclude specific files rather than nest `__test__` in `test`. If it doesn't work then we can keep the nested directory but it makes it unclear why we need the nesting unless if you're writing tests.

dgcohen:
Looks good! After our conversation, I think it makes sense to try to [extend the jest configuration](https://stackoverflow.com/questions/70832134/is-it-possible-to-put-test-files-under-pages-directory-in-next-js) to allow for tests to be picked up in the pages directory. Could you look into that as part of this work?

src/utils/searchUtils.ts
charmingduchess: line 27 is only reached if `sortBy?.length` is true. are you concerned about the logic being parallel or sort being empty?

test/__test__/utils/drbUtils.test.tsx
charmingduchess: how so?

src/utils/searchUtils.ts
EdwinGuzman: you're right. i just ignored that check initially so ignore the comment.

EdwinGuzman:
The other test PR resolves the comment above ✅

EdwinGuzman (approved):

PR #34, opened 2023-10-12
Removing nextjs config for page routing

EdwinGuzman:
Removing `pageExtension` from nextjs' config. It did not allow the render locally or on Vercel.

### DIFF
--- a/__test__/pages/404/404.test.tsx
+++ b/__test__/pages/404/404.test.tsx
@@ -1,9 +1,9 @@
 import React from "react"
 import { render, screen } from "@testing-library/react"
 
-import Custom404 from "."
-import Redirect404 from "./redirect"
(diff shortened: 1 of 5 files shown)


EdwinGuzman:
@charmingduchess sorry, but it turns out that you _do_ have to move page-specific unit test files outside of the `pages` directory. Nextjs picks it up and thinks they're pages and, because they're not, the app doesn't render.

I moved them to the `__tests__` directory.

This is not true for unit test files that are for components. Those test files can still be located in the same directory as the react component.

dgcohen (approved):

charmingduchess:
Oh darn! I had removed it but wanted to try again. The tests passed, but I neglected to run the actual app... My bad!

charmingduchess (approved):

EdwinGuzman:
Not a problem!

PR #36, opened 2023-10-24
Make search form controlled component

charmingduchess:
- Search form prepopulates search bar with the value of q from the url (not global state). 
- Select and text input are now controlled components sharing a handler function
- tests!

- also updated search results string to be dynamic to less than 50 results

### DIFF
--- a/pages/search/index.tsx
+++ b/pages/search/index.tsx
@@ -61,7 +61,9 @@ export default function Search({ results }) {
         {totalResults ? (
           <>
             <Heading level="three">
-              {`Displaying 1-50 of ${totalResults.toLocaleString()} results for keyword "${
+              {`Displaying ${
+                totalResults > 50 ? "1-50" : totalResults.toLocaleString()
+              } of ${totalResults.toLocaleString()} results for keyword "${
                 searchParams.q
               }"`}
             </Heading>
--- a/src/components/SearchForm/SearchForm.test.tsx
+++ b/src/components/SearchForm/SearchForm.test.tsx
@@ -13,6 +13,9 @@ describe("SearchForm", () => {
       screen.getByRole("button", { name: "Search" }),
       new MouseEvent("click")
     )
+  beforeEach(() => {
+    mockRouter.query.q = ""
+  })
   afterEach(async () => {
     const input = screen.getByRole("textbox")
     await userEvent.clear(input)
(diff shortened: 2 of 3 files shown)


src/components/SearchForm/SearchForm.test.tsx
```diff
@@ -0,0 +1,51 @@
+import React from "react"
```
EdwinGuzman, line 1: Very nice 👍

src/components/SearchForm/SearchForm.tsx
```diff
@@ -14,19 +14,30 @@ import { BASE_URL, PATHS } from "../../config/constants"
  */
 const SearchForm = () => {
   const router = useRouter()
+  const [searchTerm, setSearchTerm] = useState(
+    (router?.query?.q as string) || ""
+  )
+  const [searchScope, setSearchScope] = useState("all")
```
EdwinGuzman, line 20: For this PR, you should also set the select to use `router?.query?.search_scope` or `"all"`. Right now, reading `q` from the URL works but you can do the same for `search_scope`. Unless if we need to wait to implement this?

EdwinGuzman (commented):
This works fine but if there's nothing blocking updating this to include reading the `search_scope` from the url query, then it should be done now.

src/components/SearchForm/SearchForm.tsx
charmingduchess: this is the current behavior on the discovery front end. this url https://www.nypl.org/research/research-catalog/search?q=&contributor=spaghetti leads you to an empty search bar with all fields selected. 
when we discussed in refinement, it sounded like Sean wanted to keep everything as close to prod as possible. I don't love a lot the behavior between advanced and regular search, but it starts to get dicey when you consider inputting multiple fields in advanced search or fields that aren't in the regular search select.

EdwinGuzman (approved):

pages/search/index.tsx
```diff
@@ -61,7 +61,9 @@ export default function Search({ results }) {
         {totalResults ? (
           <>
             <Heading level="three">
-              {`Displaying 1-50 of ${totalResults.toLocaleString()} results for keyword "${
+              {`Displaying ${
+                totalResults > 50 ? "1-50" : totalResults.toLocaleString()
```
dgcohen, line 65: This should eventually include logic to show the actual number in case the number of results is less than 50, but that's out of scope for this PR.

dgcohen (approved):
Looks good!

PR #37, opened 2023-10-26
SCC-3742 - Upgrade DS version and update heading props

dgcohen:
This PR updates the Design System library version to 2.1.0 and replaces the heading level key values with the new element-based keys (h1, h2 etc).

It also includes a couple of minor miscellaneous changes, including removing optional chaining from theDRBCard component and replacing the sx prop with __css based on Marty's recommendation.

NB: The links are still appearing underlined in spite of the inUnderlined={false} prop. I'm currently working with Marty to figure out why this is happening.


EdwinGuzman (approved):
Looks good. I know the underlined links are still an issue but the general update for DS 2.x itself is good.

charmingduchess (approved):

PR #38, opened 2023-10-26
SCC-383x/vqa

charmingduchess:
This combines work called out in 3 VQA tickets (SCC-3830, 3831, and 3832), including:

-  Page title change to “Research Catalog | NYPL”.  
-  Add Article Search link
- SubNav wrapping keeping link text together
   - Note this one i did not follow the spec 100%, but instead implemented a simpler fix... TBD if it is acceptable.
- Placeholder text in the text input field italicized.
- An ‘X’ clear button should appear in the text input field once the patron starts typing - I believe this functionality exists in the DS search bar, but let me know if that isn’t the case. 
  - this and previous was fixed with my search form update to include the DS search bar
- 16px margin for Advanced Search button

### DIFF
--- a/src/components/SearchForm/SearchForm.tsx
+++ b/src/components/SearchForm/SearchForm.tsx
@@ -78,10 +78,10 @@ const SearchForm = () => {
             }}
           />
         </div>
-        <div className={styles.advancedSearchContainer}>
+        <div className={styles.auxSearchContainer}>
+          <EDSLink />
           <RCLink href={"/search/advanced"}>Advanced Search</RCLink>
         </div>
-        <EDSLink />
       </div>
     </div>
   )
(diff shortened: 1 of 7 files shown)


src/components/SearchForm/SearchForm.tsx
```diff
@@ -63,6 +64,7 @@ const SearchForm = () => {
               ],
             }}
             textInputProps={{
+              isClearable: true,
```
EdwinGuzman, line 67: This is also missing a call to clear the value from the internal state. This is noticeable when you try this:
1. Search for "spaghetti"
2. Click on the clear X button
3. Click on "Search"

This returns 426 results for "spaghetti". The value is removed from the search display but is still in the state. You can pass in `isClearableCallback` and pass in a call to `setSearchTerm("")`.

EdwinGuzman (changes requested):
Just one update needed to remove the search value from the internal state.

dgcohen (approved):
Agree with Edwin's comment, but otherwise looks good to me.

src/components/SearchForm/SearchForm.tsx
EdwinGuzman: That sounds great, thanks.

EdwinGuzman (approved):

PR #40, opened 2023-11-01
Scc 3664/item filters

charmingduchess:
Beginning work for item filters. I stopped writing tests because I realized a lot of the behavior we will need to test is dependent on some more of the styling and api stuff to teased out.
The component is currently rendered in the advanced search page. For that reason, the advanced search page tests are failing. I'll remove the component from that page before merging.
Built out behavior:
- maintaining state in the parent component. the checked/unchecked state, as well as the selected values is controlled by the checkbox group, which is in turn controlled by an overall selected filters state.
- selecting options
- clearing options per filter
- parsing query params (try this http://localhost:8080/research/research-catalog/search/advanced?item_location=loc%3Amakk3%2Cloc%3Amal82&item_format=Text based on production query string)
- generating query param - hitting apply on a filter should cause the computed query string to display


src/models/itemFilterData.ts
```diff
@@ -0,0 +1,70 @@
+import type {
+  ItemAggregation,
+  ItemAggregationOption,
+  option as optionType,
+} from "../types/filterTypes"
+
+export class ItemFilterData {
+  options: ItemAggregationOption[]
+  _field: string
```
dgcohen, line 9: @charmingduchess If the _field attribute doesn't need to be accessed externally, I'd suggest only having one "field" attribute and using a helper function to preprocess it for assignment in the constructor.

You can look at the getYearFromResult() helper function used to initialize the yearPublished attribute in the SearchResultsBib model as an example.

src/types/filterTypes.ts
```diff
@@ -0,0 +1,30 @@
...
+  count: number
+  label: string
+}
+
+export interface ItemAggregation {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  "@type": string
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  "@id": string
+  id: string
```
dgcohen, line 19: Out of curiosity, what's the purpose of the second id field?

src/components/ItemFilters/utils.ts
dgcohen, line 1: For consistency, can you move this into the utils directory and make the filename and function names more specific to item filters?

src/components/ItemFilters/testAggregations.js
dgcohen, line 1: This can probably go in config/aggregations. If it's too much for one file, maybe we can add an aggregations directory.

dgcohen (changes requested):
Looks good! just a couple of small comments.

src/components/ItemFilters/utils.ts
```diff
@@ -0,0 +1,50 @@
+import type { selectedFilters as selectedFiltersType } from "../../types/filterTypes"
+
+const isRecapLocation = (loc: string) => {
```
dgcohen, line 3: Screenshot 2023-11-02 at 1 35 48 PM

This we may need optional chaining or some other type of undefined check or other error handling behavior here.

src/components/ItemFilters/ItemFilters.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+    const checkboxes = screen.getAllByRole("checkbox", { checked: true })
+    expect(checkboxes.length).toBe(3)
+    const selectedValues = ["Available", "Text", "Offsite"].map((label) =>
+      screen.getByLabelText(label)
+    )
+    selectedValues.forEach((checkbox) => expect(checkbox).toBeChecked())
+  })
+
+  it("clears selection per filter", async () => {
```
EdwinGuzman, line 33: Nice.

src/components/ItemFilters/ItemFilters.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+import mockRouter from "next-router-mock"
+
+import FiltersContainer from "./FiltersContainer"
+import userEvent from "@testing-library/user-event"
+
+// Mock next router
+jest.mock("next/router", () => jest.requireActual("next-router-mock"))
+
+describe("Filters container", () => {
+  it("renders three filter boxes", () => {
```
EdwinGuzman, line 12: Can you add one with only one "agg" to show it can render dynamically not just three but also one (or two) depending on the API response.

src/components/ItemFilters/utils.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/* eslint-disable @typescript-eslint/naming-convention */
+export const combineRecapLocations = (locations: string[]) => {
+  if (locations.find(isRecapLocation)) {
+    return [...locations.filter(isRecapLocation), "Offsite"]
+  } else return locations
+}
+
+type bibPageQueryParams = {
```
EdwinGuzman, line 14: This should also be capitalized.

EdwinGuzman (commented):
My comments are minor. This is looking great!

src/types/filterTypes.ts
charmingduchess: maybe @nonword could weigh in here. The @ properties i believe refer to linked json values that are attached to a fixed URI somewhere... and that is the extent of what i can say about json ld. i assume the un-prefixed id is there for convenience on the front end.

src/components/ItemFilters/testAggregations.js
charmingduchess: this is just for development. not going to keep it here after we have api responses working. it will probably have to be used as a test fixture though, so i'll move it to test/fixtures

src/types/filterTypes.ts
nonword: Yes the `@id` is a json-ld entity id, so it's theoretically namespaced to make it unique in the graph (e.g. 'resources:1001'). Entities in JSON-LD have to define an `@id`. The `id` property is the same thing but not namespaced (e.g. '1001'), so can be used without needing to drop the namespace

src/components/ItemFilters/utils.ts
charmingduchess: i fixed this! it was actually a pretty big bug.

PR #41, opened 2023-11-03
SCC-3731 - Item Model

dgcohen:
This PR adds the Item class and the required param assigment/functions necessary for the Item table.

It also includes a couple of necessary constants for the ItemTable and some small miscellaneous fixes (util tests renamed, constant used for DRB url).


src/types/itemTypes.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export interface ItemSourceID {
+  "@type": string
+  "@value": string
+}
+
+export interface ItemStatus {
+  "@id": string
+  prefLabel?: string
```
EdwinGuzman, line 32: This pair
```
"@id": string
  prefLabel?: string`
```

appears a few times. I'm not sure if there's a better interface name for this that can be more generic. Both `ItemAccessMessage` and `ItemStatus` are the same, so it can be confusing when implementing them whether there's a difference or not.

src/utils/itemUtils.ts
```diff
@@ -0,0 +1,35 @@
+import type {
+  ItemLocationKey,
+  ItemLocation,
+  ItemLocationEndpoint,
+} from "../types/itemTypes"
+
+export const itemAvailabilityKeys = ["available", "useinlibrary"]
```
charmingduchess, line 7: this constant name could be a little clearer - even just by replacing availability with available. It's not immediately clear that these are the options are available.

src/models/Item.ts
```diff
@@ -0,0 +1,118 @@
...
+    return this.source.indexOf("Recap") !== -1
+  }
+
+  // It's an NYPL-owned ReCAP item if item source is Sierra and location is ReCAP
+  isNYPLReCAP(): boolean {
+    return this.isSierraItem() && this.locationIsReCAP()
+  }
+
+  // It's non-ReCAP NYPL-owned item if item source is Sierra and location is not ReCAP
+  isNYPLNonReCAP(): boolean {
```
charmingduchess, line 106: Similar to non-recap, I'd prefer onsite here as opposed to non-recap because its more precise

src/models/Item.ts
```diff
@@ -0,0 +1,118 @@
...
+  volume?: string
+  format?: string
+  barcode?: string
+  location?: ItemLocation
+  aeonUrl?: string
+  isPhysicallyRequestable: boolean
+  isEDDRequestable: boolean
+
+  constructor(item: SearchResultsItem, bib: SearchResultsBib) {
+    this.id = item["@id"] ? item["@id"].substring(4) : ""
```
charmingduchess, line 38: you could also use `item.uri` and skip the string manipulation.

src/types/itemTypes.ts
charmingduchess: I think it would be worthwhile to have a generic `JSONLDValue` type.

src/models/Item.ts
```diff
@@ -0,0 +1,118 @@
...
+    this.callNumber = item.shelfMark.length ? item.shelfMark[0] : null
+    this.volume = item.enumerationChronology?.length
+      ? item.enumerationChronology[0]
+      : null
+    this.format = item.formatLiteral?.length
+      ? item.formatLiteral[0]
+      : bib.materialType
+    this.barcode = item.idBarcode?.length ? item.idBarcode[0] : null
+    this.location = this.getLocationFromItem(item)
+    this.aeonUrl = item.aeonUrl?.length ? item.aeonUrl[0] : null
```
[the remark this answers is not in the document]
dgcohen: Good call!

src/types/itemTypes.ts
```diff
@@ -0,0 +1,44 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+
+// Item structure coming from the Search Results API response
+export interface SearchResultsItem {
+  "@id": string
+  idNyplSourceId?: ItemSourceID
+  accessMessage?: ItemAccessMessage[]
+  shelfMark?: string[]
+  electronicLocator?: string[]
```
charmingduchess, line 9: electronicLocator has been deprecated, so I think it's safe to remove it. We used to return items from the discovery api with electronic resources mixed in, so electronicLocator was where the urls were stored. we don't do that anymore. electronic resources are attached to the bib only.

src/models/Item.ts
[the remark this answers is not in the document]
dgcohen: Actually, a generic helper could be a bit tricky to implement due to the differing types for both the object and the accessed field. Should be doable but I think it falls out of the scope of this PR.

src/models/Item.ts
```diff
@@ -0,0 +1,118 @@
...
+    this.isPhysicallyRequestable = item.physRequestable
+    this.isEDDRequestable = item.eddRequestable
+  }
+
+  // Item availability is determined by the existence of status label in the availability keys list
+  get isAvailable(): boolean {
+    // Lowercase and remove non-word characters from status label
+    const availability =
+      !isEmpty(this.status) && this.status?.prefLabel
+        ? this.status.prefLabel.replace(/\W/g, "").toLowerCase()
```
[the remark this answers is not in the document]
dgcohen: Good call! I changed it to instead check on the status id rather than the prefLabel

charmingduchess:
Looks good! Not sure if I'm not seeing the most updated code, but can we replace all instances of nonNypl with partner and nonRecap with onsite?

EdwinGuzman (approved):

PR #42, opened 2023-11-03
Functional Filters!

charmingduchess:
NB: to see the filters in action, head to the vercel preview and click on the advanced search button. clicking apply will show you the query that would be added to the url.

Basic overview:
- Filters open and close. 
   - I did not use the Accordion component because there is a ton of styling overrides required. Also much of the open/closed logic would still need to be ported from this solution, such as only having one filter open at a time, and closing on blur and escape key.
- options selected in a checkbox group are not persisted when that filter closes unless the filter has been applied
- applied filters are the filters that generate the url (and are in turn generated by the url)

honestly more than anything i would appreciate just folks clicking around and trying to break things! the expected behavior is a little kooky but I'm confident this work does what DFE does, for better or for worse.

I closed 3664 because i was getting very tangled going back and forth between the two branches. i incorporated the rest of yall's requests into this PR.


charmingduchess:
Added styling!

src/utils/utilsTests/itemFilterUtils.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+} from "../itemFilterUtils"
+import testAggregations from "../../../__test__/fixtures/testAggregations"
+import { ItemFilterData } from "../../models/itemFilterData"
+
+describe("Item Filter Utils", () => {
+  describe("isRecapLocation", () => {
+    it("returns true for a recap location", () => {
+      expect(isRecapLocation("loc:rc2ma")).toBe(true)
+    })
+    it("returns false for a recap location", () => {
```
EdwinGuzman, line 16: Description should be for a "non-recap" location, right?

src/utils/utilsTests/itemFilterUtils.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+        "loc:mab",
+        "loc:mac",
+        "loc:spaghetti",
+      ])
+    })
+    it("replaces offsite locations with no other locations", () => {
+      const locations = ["loc:rc2ma", "loc:rcma2", "loc:rcrc"]
+      expect(combineRecapLocations(locations)).toEqual(["Offsite"])
+    })
+    it("can handle an empty array", () => {
```
EdwinGuzman, line 37: This is fine. In the more business logic text, it could be something like, "can handle the situation where there are no locations".

src/components/ItemFilters/ItemFilters.test.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+      // clicking other filter label closes the location filter
+      await userEvent.click(statusCheckboxGroupButton)
+      await userEvent.click(locationCheckboxGroupButton)
+      const applyButton = screen.getByTestId("clear-location-button")
+      expect(applyButton).toBeDisabled()
+    })
+  })
+
+  it.only("persists previously applied selection after closing filter without applying", async () => {
```
EdwinGuzman, line 99: Remove the `.only`.

src/models/itemFilterData.ts
```diff
@@ -0,0 +1,72 @@
...
+  labelsForConcatenatedValues(values: string) {
+    return Array.from(
+      new Set(
+        values.split(",").map((val: string) => `'${this.labelForValue(val)}'`)
+      )
+    ).join(", ")
+  }
+}
+
+export class LocationFilterData extends ItemFilterData {
```
EdwinGuzman, line 38: I wonder if you need tests for these classes when it's pretty much covered in the other tests. What do you think? Or perhaps a later PR?

src/components/ItemFilters/ItemFilterButtons.tsx
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <ButtonGroup
+      className="item-filter-buttons"
+      isDisabled={selectedOptions.length === 0}
+    >
+      <Button
+        data-testid={`clear-${field}-button`}
+        key={`clear-${field}-button`}
+        buttonType="text"
+        id="clear-filter-button"
```
EdwinGuzman, line 45: Hmm I initially thought this _might_ be a duplicate ID accessibility issue, but only one filter is opened at any point. Would be good to double check this.

src/components/ItemFilters/ItemFilterLabel.tsx
```diff
@@ -0,0 +1,43 @@
...
+  selectedOptions,
+  setWhichFilterIsOpen,
+  isOpen,
+}: ItemFilterLabelProps) => {
+  const fieldFormatted = field[0].toUpperCase() + field.substring(1)
+  return (
+    <Button
+      className={styles.itemFilterButton}
+      sx={{
+        borderColor: "var(--nypl-colors-ui-gray-medium)",
```
EdwinGuzman, line 25: You can use `ui.gray.medium` for the design token value.

EdwinGuzman: This only works for DS/Chakra components.

src/types/filterTypes.ts
```diff
@@ -0,0 +1,30 @@
+export type Locations = string[]
+
+export type AppliedFilters = {
```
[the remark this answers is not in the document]
charmingduchess: the location format status keys are hardcoded in `parseItemFilterQueryParams` in `itemFilterUtils.ts`, so I don't think there's a risk of things breaking. If you're opposed to those keys being hard coded, we can discuss changing that.

src/utils/itemFilterUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export const buildItemFilterQueryParams = (
+  { location, format, status }: AppliedFilters,
+  recapLocations: string
+) => {
+  const locs = location.map((loc) => {
+    if (loc === "Offsite") return recapLocations
+    else return loc
+  })
+  const location_query = location.length
```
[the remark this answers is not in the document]
charmingduchess: updated

src/utils/itemFilterUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+    ? "item_location=" + locs.join(",")
+    : ""
+  const format_query = format.length ? "&item_format=" + format.join(",") : ""
+  const status_query = status.length ? "&item_status=" + status.join(",") : ""
+
+  const query = encodeURI(`?${location_query}${format_query}${status_query}`)
+  return query.length > 3 ? query : ""
```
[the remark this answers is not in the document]
charmingduchess: that check was from when I was including the ampersand in the template literal, so good catch! im changing how that check is happening.

src/utils/utilsTests/itemFilterUtils.test.tsx
charmingduchess: yes!

apoorvakhanolkar:
My only comment is to set the dropdowns to a fixed width of 200px with a spacing of 32px between them, so that they have some breathing room.

charmingduchess:
@apoorvakhanolkar would you be okay 230px for the filter width and 16px margin? At 200px the longer labels wrap in a way that I find hard to read

apoorvakhanolkar:
> @apoorvakhanolkar would you be okay 230px for the filter width and 16px margin? At 200px the longer labels wrap in a way that I find hard to read

Yes that's OK

src/types/filterTypes.ts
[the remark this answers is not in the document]
EdwinGuzman: Gotcha, that's clearer now.

src/models/itemFilterData.ts
EdwinGuzman: I'm on the fence, too, which is why I asked. I think integration is fine since it covers a lot but now that we're including tests and I see a file without one, it makes me _feel_ like that's an issue. But it's really not.

apoorvakhanolkar (approved):

charmingduchess:
We decided to not worry about the tabbing/z index/focus issue until we get feedback from Clare about accessibility

EdwinGuzman (approved):

PR #43, opened 2023-11-07
SCC-3770 - Search Results Pagination

dgcohen:
This PR adds the following:

1. Pagination component from the DS, including a click handler that updates the router with the appropriate page in the query string.
2. A utility function that programmatically generates page offset value strings based on the current page and the page limit (e.g. Showing results 501-550)
3. Parsing of the page number in the query string into a number. This was necessary this since the QS library parses numbers in the query string as strings, and overriding this functionality globally is too cumbersome for our use case.

### DIFF
--- a/pages/search/index.tsx
+++ b/pages/search/index.tsx
@@ -1,5 +1,9 @@
 import Head from "next/head"
-import { Heading, SimpleGrid } from "@nypl/design-system-react-components"
+import {
+  Heading,
+  SimpleGrid,
+  Pagination,
+} from "@nypl/design-system-react-components"
 import { useRouter } from "next/router"
 import { parse } from "qs"
 
@@ -11,9 +15,10 @@ import { fetchResults } from "../api/search"
 import {
   mapQueryToSearchParams,
   mapElementsToSearchResultsBibs,
+  getQueryString,
 } from "../../src/utils/searchUtils"
 import { mapWorksToDRBResults } from "../../src/utils/drbUtils"
-import { SITE_NAME } from "../../src/config/constants"
+import { SITE_NAME, RESULTS_PER_PAGE } from "../../src/config/constants"
 import type SearchResultsBib from "../../src/models/SearchResultsBib"
 
 /**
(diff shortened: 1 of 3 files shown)


src/utils/searchUtils.ts
```diff
@@ -184,17 +202,18 @@ export function mapQueryToSearchParams({
   lccn,
   filters,
 }: SearchQueryParams): SearchParams {
+  const hasIdentifiers = issn || isbn || oclc || lccn
```
EdwinGuzman, line 205: Was this added because it caused an issue?

src/utils/searchUtils.ts
```diff
@@ -8,6 +8,23 @@ import type {
   SearchResultsElement,
 } from "../types/searchTypes"
 import SearchResultsBib from "../models/SearchResultsBib"
+import { RESULTS_PER_PAGE } from "../config/constants"
+
+/**
+ * getPaginationOffsetStrings
+ * Used to generate search results start and end counts on Search Results page
+ */
+export function getPaginationOffsetStrings(
```
EdwinGuzman, line 17: It feels like this could use a few tests in `searchUtils.test.tsx`.

EdwinGuzman (commented):
Looks good! Just some comments.

pages/search/index.tsx
```diff
@@ -58,7 +73,9 @@ export default function Search({ results }) {
           <>
             <Heading level="h2" mb="xl" size="heading4">
               {`Displaying ${
-                totalResults > 50 ? "1-50" : totalResults.toLocaleString()
+                totalResults > RESULTS_PER_PAGE
+                  ? `${resultsStart}-${resultsEnd}`
+                  : totalResults.toLocaleString()
               } of ${totalResults.toLocaleString()} results for keyword "${
```
charmingduchess, line 79: potentially out of scope for this, but should "keyword" in this string be dynamic to whatever search query was made?

charmingduchess: Also, this string is becoming unwield. Could it be replaced with a variable and a util to generate it?

charmingduchess (approved):
one suggestion for some abstraction, but otherwise looks good.

PR #44, opened 2023-11-07
SCC-3771 - Search Results Sorting

dgcohen:
Ticket: https://jira.nypl.org/browse/SCC-3771

This PR adds the Select field in the search results page that controls the sorting of the Search Results.

It works by simply setting the sort direction and sort by fields in SearchParams, and removing the page value if set (in case the user selected another page) before updating the router with the new params, which internally handles the re-hydrating of the Search Results.


pages/search/index.tsx
```diff
@@ -37,6 +43,11 @@ export default function Search({ results }) {
   // Map DRB Works from response to DRBResult objects
   const drbResults = mapWorksToDRBResults(drbWorks)
 
+  const handleSortChange = async (e) => {
+    const [sortBy, order] = e.target.value.split("_")
+    await replace(getQueryString({ ...searchParams, sortBy, order }))
```
EdwinGuzman, line 48: Why `replace` and not `push`?

src/utils/searchUtils.ts
```diff
@@ -191,8 +205,8 @@ export function mapQueryToSearchParams({
     contributor,
     title,
     subject,
-    sortBy: sort,
-    order: sort_direction,
+    sortBy: sort as SortKey,
```
EdwinGuzman, line 208: Why not set `sortBy` as `SortKey` in the `SearchQueryParams` interface so you don't have to do the `... as ...` type coercion? Same for `sort_direction`.

EdwinGuzman (commented):
Functionally, it works well. It's a bit odd how it takes time for the select to update. It seems that it needs to load the new results before the entire page is interactive and updated again. Not a big problem and probably not one when we can implement a loading UI while fetches are made.

EdwinGuzman (approved):

__test__/pages/search/searchResults.test.tsx
```diff
@@ -23,6 +24,20 @@ describe("Search Results page", () => {
...
+    it("renders the sort select field and updates the query string in the url on changes", async () => {
+      const query = "spaghetti"
+      await mockRouter.push(`/search?q=${query}`)
+      render(<SearchResults results={results} />)
+
+      const sortSelect = screen.getByLabelText("Sort by")
+      expect(sortSelect).toHaveValue("relevance")
+      await userEvent.selectOptions(sortSelect, "Title (A - Z)")
+      expect(sortSelect).toHaveValue("title_asc")
+
```
charmingduchess, line 36: your test implicitly tests that the page param is removed, but can it be its own expectation? just for documentation purposes.

src/utils/utilsTests/searchUtils.test.ts
```diff
@@ -30,7 +30,7 @@ describe("searchUtils", () => {
           isbn: "456",
           search_scope: "contributor",
           sort_direction: "asc",
-          sort: "spaghetti",
+          sort: "relevance",
```
charmingduchess, line 33: 🍝 
seems like the epitome of relevance to me but what do i know...

pages/search/index.tsx
```diff
@@ -37,6 +43,11 @@ export default function Search({ results }) {
   // Map DRB Works from response to DRBResult objects
   const drbResults = mapWorksToDRBResults(drbWorks)
 
+  const handleSortChange = async (e) => {
+    const [sortBy, order] = e.target.value.split("_")
```
charmingduchess, line 47: can you add a comment here describing what's happening?

charmingduchess (commented):
Going off of what Edwin mentioned about the loading taking a long time, one thing I noticed in this and the other PR is that not only does the page take a while to load, the url doesn't update until the page is loaded. I'm assuming that's something to do with getServerSideProps activity. I'm curious though if theres a way to have the url update before the page has finished loading. It seems weird as a user that there's API calls being made under the hood before the URL has updated. Maybe also just the nature of not having any loading indicator. The current DFE, however, does update the URL before the page has loaded. I get that there's a different server/api/query mechanism happening in DFE, but it feels satisfying to click submit and see the url populated with your search terms.

PR #45, opened 2023-11-20
Scc 3279/item filter mobile

charmingduchess:
- Update item filters to be responsive to mobile and tablet view. 
- eliminate the previous modal implementation. 
- add dummy search bar for styling/layout purposes.


apoorvakhanolkar (approved):

styles/components/ItemFilters.module.scss
```diff
@@ -1,30 +1,44 @@
...
 }
 
-.filterGroup {
-  display: flex;
-  width: fit-content;
+// filter group is a generic class for any filters that may exist. As of 11/22
+// there are the aggregation filters and the search by year filter
+.filterGroupMobile {
+  flex-grow: 1;
+  & .itemFilterOptionsContainer {
```
EdwinGuzman, line 18: The ampersand is not necessary here. The styles will compile to `.filterGroupMobile .itemFilterOptionsContainer`. If the usage for this is to have more specificity, you can remove the space `&.itemFilterOptionsContainer`. When compiled, the output will be `.filterGroupMobile.itemFilterOptionsContainer`.

src/components/ItemFilters/FilterGroup.tsx
```diff
@@ -0,0 +1,18 @@
+import {
+  Card,
+  CardHeading,
+  CardContent,
+} from "@nypl/design-system-react-components"
+
+const FilterGroup = (children, headingText, className, ref) => {
```
EdwinGuzman, line 7: I don't think you're using this component.

styles/components/ItemFilters.module.scss
```diff
@@ -1,30 +1,44 @@
...
+  gap: var(--nypl-space-s);
 }
 
-.filterGroup {
-  display: flex;
-  width: fit-content;
+// filter group is a generic class for any filters that may exist. As of 11/22
+// there are the aggregation filters and the search by year filter
+.filterGroupMobile {
+  flex-grow: 1;
```
EdwinGuzman, line 17: I can't figure out why but between 500-600px, both containers want to stay in the same row. You can add `width: 100%` here so that the filter container and year search container span the full width. Otherwise, this happens:
Screen Shot 2023-11-27 at 4 22 17 PM

EdwinGuzman (approved):
Looks good! I pointed out a layout issue but approving because, like you mentioned, the year search container is temporary for now. It should get updated at some point but it's not crucial for this PR.

PR #46, opened 2023-11-21
Scc 3853/url and display string

charmingduchess:
- performance work to remove extraneous rerenders and `useEffect` call
  - removing `useEffect` with dependency on `appliedFilters` and instead passing a `submitFilters` function down to filter components.
- create the filter display string (ie "34 Items matched filtered by xyz") based on the query params provided by the next router
- styling display string based on Design System components
- stray bug and styling fixes, not limited to:
  - removing space from the filter bar
  - making sure count on filter bar reflects applied and not just selected filters


src/utils/utilsTests/itemFilterUtils.test.tsx
```diff
@@ -84,55 +84,40 @@ describe("Item Filter Utils", () => {
...
-      expect(buildItemFilterQueryParams(query, recapLocations)).toBe(
+      expect(buildItemFilterQueryString(query, recapLocations)).toBe(
         "?item_location=loc:abc,loc:rc2ma,loc:rc3ma,loc:rc4ma"
       )
     })
   })
 
-  describe("buildAppliedFiltersString", () => {
-    const query = {
+  describe.only("buildAppliedFiltersString", () => {
```
EdwinGuzman, line 93: Remove the `.only`.

EdwinGuzman (changes requested):
Works well for Location and Status! For Format, it doesn't seem to have the right id/key for the last two checkboxes and renders `Filtered by format: 'undefined'`
Screen Shot 2023-11-27 at 4 03 52 PM

Also added comments for updated patterns.

charmingduchess:
> Works well for Location and Status! For Format, it doesn't seem to have the right id/key for the last two checkboxes and renders `Filtered by format: 'undefined'` Screen Shot 2023-11-27 at 4 03 52 PM
> 
> Also added comments for updated patterns.
@EdwinGuzman this is a data issue - there are dates in the format section which breaks the internal logic. i updated the fixture on a later branch to use more standard formats. paul said we should not account for bad data.

EdwinGuzman:
> @EdwinGuzman this is a data issue - there are dates in the format section which breaks the internal logic. i updated the fixture on a later branch to use more standard formats. paul said we should not account for bad data.

Great, sounds good.

src/components/ItemFilters/FiltersContainer.tsx
```diff
@@ -17,62 +24,76 @@ interface ItemFilterContainerProps {
...
     )
+    router.push("/search/advanced" + url)
   }
 
-  useEffect(() => {
-    setAppliedFilters(parseItemFilterQueryParams(query))
-  }, [query])
-
-  useEffect(tempSubmitFilters, [activeFilters, tempSubmitFilters])
+  useMemo(() => {
```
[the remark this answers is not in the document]
charmingduchess: awesome! yes this makes sense.

src/components/ItemFilters/FiltersContainer.tsx
```diff
@@ -17,62 +23,75 @@ interface ItemFilterContainerProps {
...
-    else return new ItemFilterData(agg)
-  })
-  const [activeFilters, setAppliedFilters] = useState({
-    location: [],
-    format: [],
-    status: [],
-  })
+  const router = useRouter()
+
+  const filterData = useRef(
```
EdwinGuzman, line 28: For type safety and to better see how `filterData` is structured, can you type this as `useRef<ItemFilterData[]>(`

src/components/ItemFilters/FiltersContainer.tsx
```diff
@@ -17,62 +23,75 @@ interface ItemFilterContainerProps {
...
+    appliedFilters,
+    filterData
+  )
   const ref = useRef<HTMLDivElement>(null)
   useCloseDropDown(() => setWhichFilterIsOpen(""), ref)
 
   const [whichFilterIsOpen, setWhichFilterIsOpen] = useState("")
 
-  const [tempQueryDisplay, setTempQueryDisplay] = useState("")
+  const itemsMatched = buildItemsString(router.query)
```
EdwinGuzman, line 48: What do you think about renaming the function as `buildItemsMatchedString` or is that too long?

src/utils/itemFilterUtils.ts
```diff
@@ -51,27 +51,30 @@ export const buildItemFilterQueryParams = (
   else return ""
 }
 
-export const buildAppliedFiltersString = (
-  query: BibPageQueryParams,
-  numItems = 20,
-  itemAggs: ItemFilterData[]
-) => {
+export const buildItemsString = (query, numItems = 20) => {
```
EdwinGuzman, line 54: Why is 20 the default?

charmingduchess: it's just for temporary display purposes because i don't know what kind of data is going to be passed in from the bib page. ill add a comment.

EdwinGuzman (approved):
Minor comments but the update looks good.

PR #48, opened 2023-11-27
Loading State Prototype

dgcohen:
This is a prototype for a way in which the Loading state could work in the new app.

It works by attaching setState handlers to Next's router events (routeChangeStart, routeChangeComplete etc) and showing the DFE's <SkeletonLoader/> component if loading is in progress.


EdwinGuzman (commented):
This is very nice. Can you update this to include the SearchBar component to get it on all interactive elements on the page? It'd be just for prototype purposes -- I don't think the user should make updates while the rest of the page is loading.

I haven't looked into it and we chatted briefly about it, but I worry that Next 13+ will have a different way to implement this. Let's also get Apoorva's input on this.

charmingduchess:
Looks good as is, but I'm wondering if this is a good use case for [React's Suspense component](https://react.dev/reference/react/Suspense)?

dgcohen:
> This is very nice. Can you update this to include the SearchBar component to get it on all interactive elements on the page? It'd be just for prototype purposes -- I don't think the user should make updates while the rest of the page is loading.
> 
> I haven't looked into it and we chatted briefly about it, but I worry that Next 13+ will have a different way to implement this. Let's also get Apoorva's input on this.

Just added the Loading state to the Search Form. Had to pass in a loading boolean prop, since the SearchBar currently includes the maxwidth container for the search bar.

EdwinGuzman:
> Looks good as is, but I'm wondering if this is a good use case for [React's Suspense component](https://react.dev/reference/react/Suspense)?

I agree and think that this is a good use case for suspense. Let's wait until we move to next 13 and update to react 18 before implementing suspense. For now, this looks good.

@dgcohen can you ask Clare for an accessibility review for this feature? It's functional but I don't know enough to say how accessible it is.

charmingduchess (approved):

EdwinGuzman (approved):
The update looks good. It may be missing in a few places but we can clean those up later (such as the select dropdown and the pagination).

PR #51, opened 2023-12-05
SCC-3705: Search result fieldset date input

EdwinGuzman:
Fixes JIRA ticket [SCC-3705](https://jira.nypl.org/browse/SCC-3705)

- Converts the class-based `[FieldsetDate](https://github.com/NYPL/discovery-front-end/blob/development/src/app/components/Filters/FieldsetDate.jsx)` component into a functional component.
- Uses the DS `DatePicker` to render the input fields, invalid text, and helper text.
- Adds tests but can't be fully tested just yet until the parent is created and used to track updates.


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "git.ignoreLimitWarning": true,
+  "editor.formatOnSave": true,
```
EdwinGuzman, line 3: Not part of the ticket but I found this helpful.

dgcohen (approved):
This looks good! I'm assuming the rendering of this component will we added in a separate PR?

PR #52, opened 2023-12-05
SCC-3759: Hold pages server utils

EdwinGuzman:
Resolves JIRA ticket [SCC-3759](https://jira.nypl.org/browse/SCC-3759).

- This _moves_ over and structures the Hold Pages-related functions that make API calls and manipulate the data. This PR sets up the scaffold for calling these functions.
- This PR itself could use a refactor for some API calls.
- The `User` and `Bib` classes are mocked and are empty placeholders until they are implemented. This means that this work is not functional and depends on those functions. These should be worked on separately.
- To not make this bigger than it needs to be, I did not add tests. Tests should be added for each page once they're built out.


pages/hold/request/[bibId-itemId-itemSource]/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+      { bibId: bibAndItemId },
+      {},
+      {
+        fetchSubjectHeadingData: false,
+        features: urlEnabledFeatures,
+      }
+    )
+    const { bib } = bibResponseData
+    barcode = LibraryItem.getItem(bib, itemId).barcode
+    const urlIsAeon = bib.items
```
Comment (EdwinGuzman) [inline pages/hold/request/[bibId-itemId-itemSource]/index.tsx:55]: @charmingduchess check if Aeon url logic is needed or why we need to redirect.

Comment (charmingduchess) [reply pages/hold/request/[bibId-itemId-itemSource]/index.tsx:55]: looks like it is addressing requests made in legacy from [this ticket](https://jira.nypl.org/browse/SCC-2963). Can you add a comment in there calling that out?

Comment (EdwinGuzman) [reply pages/hold/request/[bibId-itemId-itemSource]/index.tsx:55]: I don't know whether this is true or not. This part is copied over from DFE so I'd have to do more investigation before stating that's true and addresses the ticket.

PR #55, opened 2023-12-08
SCC-3897 - Search Results Electronic Resources Link

dgcohen:
Ticket: https://jira.nypl.org/browse/SCC-3897

This PR adds the ElectronicResourcesLink component that is mostly derived from the ElectronicResourcesResultsItem component in DFE, with a few minor adjustments.

### DIFF
--- a/src/components/SearchResult/ElectronicResourcesLink.tsx
+++ b/src/components/SearchResult/ElectronicResourcesLink.tsx
@@ -1,8 +1,6 @@
 import {
   Box,
   Heading,
-  Text,
-  Icon,
   Link as DSLink,
 } from "@nypl/design-system-react-components"
 
--- a/src/components/SearchResult/ElectronicResourcesLink.tsx
+++ b/src/components/SearchResult/ElectronicResourcesLink.tsx
@@ -1,0 +1,58 @@
+import {
+  Box,
+  Heading,
+  Text,
+  Icon,
+  Link as DSLink,
+} from "@nypl/design-system-react-components"
+
+import RCLink from "../RCLink/RCLink"
+import type { ElectronicResource } from "../../types/bibTypes"
+
+interface ElectronicResourcesLinkProps {
+  bibUrl?: string
+  electronicResources: ElectronicResource[]
+}
+
+/**
+ * The SearchResult component displays a single search result element.
+ */
+const ElectronicResourcesLink = ({
+  bibUrl,
+  electronicResources,
+}: ElectronicResourcesLinkProps) => {
+  return (
+    <Box>
+      <Heading level="four" size="callout" mb="xxs">
+        Available Online
+      </Heading>
+      {electronicResources.length === 1 ? (
+        <DSLink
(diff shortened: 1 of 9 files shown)


src/components/SearchResult/SearchResult.tsx
```diff
@@ -29,12 +30,18 @@ const SearchResult = ({ bib }: SearchResultProps) => {
...
       </CardHeading>
       <CardContent>
-        <Box sx={{ p: { display: "inline", marginRight: "s" } }}>
+        <Box sx={{ p: { display: "inline", marginRight: "s" } }} mb="m">
           {bib.materialType && <Text>{bib.materialType}</Text>}
           {bib.publicationStatement && <Text>{bib.publicationStatement}</Text>}
           {bib.yearPublished && <Text>{bib.yearPublished}</Text>}
           <Text>{bib.itemMessage}</Text>
         </Box>
+        {bib.electronicResources && (
```
EdwinGuzman, line 39: In the Vercel preview, all bib search results are displaying the `ElectronicResourcesLink` and that's not what true for the same searches in production. Something seems to be off.

src/components/SearchResult/ElectronicResourcesLink.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <DSLink
+          href={electronicResources[0].url}
+          target="_blank"
+          rel="noreferrer"
+        >
+          {electronicResources[0].prefLabel || electronicResources[0].url}
+        </DSLink>
+      ) : (
+        <RCLink href={`${bibUrl}#electronic-resources`}>
+          <Text
```
EdwinGuzman, line 39: Don't render `p` tags inside links.

In this case, I think this block can be a Link with DS type of "standalone" https://nypl.github.io/nypl-design-system/reservoir/v2/?path=/story/components-navigation-link--with-controls&args=type:standalone
This will render the arrow for you.

EdwinGuzman (changes requested):
One minor update but the logic check for when to display needs a closer look.

src/components/SearchResult/SearchResult.tsx
charmingduchess: I think the issue cab be addressed here https://github.com/NYPL/research-catalog/blob/5f63115adf9986c4f56817090f353cccba19cd51/src/models/SearchResultsBib.ts#L35
There should be a check for `electronicResources?.length` because an empty array is truthy.

charmingduchess (commented):
Looks good, except for the bug that has resources rendering for all bibs. I have one small ask- can you change the name of this PR to be ER or Electronic Resources Link instead of EDD?

src/components/SearchResult/SearchResult.tsx
dgcohen: Just added a helper function hasElectronicResources that checks for the length. This resolves the issue.

EdwinGuzman (approved):

charmingduchess (approved):

PR #56, opened 2023-12-11
SCC-0898 - Search Results View All Link

dgcohen:
Ticket: https://jira.nypl.org/browse/SCC-3898

This PR renders a View All x Items link that leads to the Bib page when the number of physical items exceeds 3.

Notes:

1. This should be moved into the conditional that renders the ItemTable when the two branches are merged.
2. I changed the item location endpoint type from ItemLocationEndpoint to a generic string to avoid having to cast the type for every item fixture. Not sure if there's a more elegant way to handle this in typescript, but it shouldn't make a difference in practice.

### DIFF
--- a/src/components/Layout/Layout.tsx
+++ b/src/components/Layout/Layout.tsx
@@ -49,6 +49,13 @@ const Layout = ({
                     text: "Research Catalog",
                   },
                 ]}
+                __css={{
+                  a: {
+                    _focus: {
+                      outlineColor: "ui.white",
+                    },
+                  },
+                }}
               />
               <div className={styles.researchHeadingContainer}>
                 <Heading id="heading-h1" level="h1" text="Research Catalog" />
--- a/src/components/RCLink/RCLink.tsx
+++ b/src/components/RCLink/RCLink.tsx
@@ -12,6 +12,7 @@ interface RCLinkProps {
   className?: string
   color?: string
   type?: LinkTypes
+  hasWhiteFocusRing?: boolean
 }
 
 /**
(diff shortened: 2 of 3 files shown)


src/components/SearchResult/SearchResult.tsx
```diff
@@ -29,12 +31,33 @@ const SearchResult = ({ bib }: SearchResultProps) => {
...
           {bib.materialType && <Text>{bib.materialType}</Text>}
           {bib.publicationStatement && <Text>{bib.publicationStatement}</Text>}
           {bib.yearPublished && <Text>{bib.yearPublished}</Text>}
           <Text>{bib.itemMessage}</Text>
         </Box>
+        {/* Move the code block below to the conditional for rendering the Item Table */}
+        {bib.numPhysicalItems > ITEMS_PER_SEARCH_RESULT && (
+          <CardActions>
+            <RCLink href={`${bib.url}#items-table`}>
+              <Text
```
EdwinGuzman, line 44: Similar to the other PR, don't render `p` elements inside `a` elements. This `Link` can also be of type "standalone" which you may need to propagate down from the `RCLink`.

EdwinGuzman (changes requested):
One update similar to PR #55

charmingduchess (approved):

src/components/SearchResult/SearchResult.tsx
dgcohen: Good catch, just made both links of the "standalone" type.

EdwinGuzman (approved):

PR #57, opened 2023-12-12
Add basic reverse-proxy test with documentation

nonword:
Adds basic prototype for running an nginx server that defers certain site areas to RC and certain areas to QA DFE

Note that tests are failing due to broken tests in `main`


EdwinGuzman (approved):
Tested and works locally for me. I think it could use additional routes but this is a good start.

dgcohen (approved):
Looks good! Just merged in main to fix the failing tests.

charmingduchess:
Will we need the complementary reverse proxy for DFE as well? To route back to RC for search and bib details?

PR #59, opened 2023-12-13
SCC-3904: Advanced Search page A11y and VQA fixes

EdwinGuzman:
Resolves JIRA ticket [SCC-3904](https://jira.nypl.org/browse/SCC-3904).

This fixes six main issues found on the Advanced Search page:
- Removes the `DatePicker` component for text input components. Accessibility issues were flagged in the `DatePicker` and the DS team wants to take a step back to fix them. So it's not a blocker, the `TextInput` component should be used for now.
- Adds focus to the error message when it is rendered.
- Updates the error messages to include the work "Error: ..."
- Adds a basic validation so that the "From" date cannot be bigger than the "To" date.
- Updates the button group layout for the Clear and Submit button.
- Moves the date fields to the first column for better visual layout.

### DIFF
--- a/__test__/pages/search/advancedSearchForm.test.tsx
+++ b/__test__/pages/search/advancedSearchForm.test.tsx
@@ -3,7 +3,10 @@ import { fireEvent, render, screen, act } from "@testing-library/react"
 import mockRouter from "next-router-mock"
 import userEvent from "@testing-library/user-event"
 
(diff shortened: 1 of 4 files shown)


EdwinGuzman:
Passed accessibility and VQA reviews.

dgcohen (approved):
This looks good! It would be good to eventually abstract some of this error handling, but we can address that as a follow up.

PR #60, opened 2023-12-18
Add PickupTimeEstimator testing tool

nonword:
Adds PickupTimeEstimator module over to RC with a `pages/test-pickup-times` route.

Known issues:
 - Are we okay with releasing an unadvertised (but unprotected) `pages/test-pickup-times` route? Is there a better path or access restriction to employ?
 - I'm not sure the `pages/test-pickup-times` component correctly implements route-prop interaction (i.e. form controls update state, which push a route change, which trigger re-execution of `getServerSideProps`, etc)
 - Includes several shims correcting for location-service behavior, which should be closed soon.
 - Does not yet include the [special delivery logic currently implemented in DFE](https://github.com/NYPL/discovery-front-end/pull/2137)
 - the TestPickupTimes test file is rudimentary, but perhaps that's okay for an unadvertised testing tool?


pages/test-pickup-times/index.tsx
```diff
@@ -0,0 +1,416 @@
...
+  fulfillment: string
+  holdingLocation: string
+  deliveryLocation: string
+  requestTime: string
+  currentTime: string
+  scenario: string
+}
+
+/**
+ * The Search page is responsible for fetching and displaying the Search results,
```
dgcohen, line 142: This comment is from SearchResults, would be good to have one giving a brief overview of TestPickupTimes.

pages/test-pickup-times/index.tsx
```diff
@@ -0,0 +1,416 @@
...
+
+import Layout from "../../src/components/Layout/Layout"
+
+import { BASE_URL, SITE_NAME } from "../../src/config/constants"
+import {
+  getPickupTimeEstimate,
+  overrideNow,
+} from "../../src/utils/pickupTimeEstimator"
+
+// Prepare fulfillment dropdown values:
```
dgcohen, line 26: Small nit but I think it would be good to move most of the stuff preceding the component to a separate pickupTimesUtils file.

pages/test-pickup-times/index.tsx
```diff
@@ -0,0 +1,416 @@
...
+  if (
+    currentParams.fulfillment !== fulfillment ||
+    currentParams.holdingLocation !== holdingLocation ||
+    currentParams.deliveryLocation !== deliveryLocation ||
+    currentParams.requestTime !== requestTime ||
+    currentParams.currentTime !== currentTime
+  ) {
+    if (scenario && scenarios[scenario]) {
+      Object.keys(scenarios[scenario]).forEach((key) => {
+        if (scenarios[scenario][key] !== eval(key)) {
```
dgcohen, line 197: This seems like a lot of nested if statements, would it be possible to refactor this a bit?

pages/test-pickup-times/index.tsx
```diff
@@ -0,0 +1,416 @@
...
+              </Select>
+              <HorizontalRule __css={{ margin: 0 }} />
+
+              <TextInput
+                id="holdingLocation"
+                labelText="Item Location"
+                type="text"
+                name="holdingLocation"
+                value={holdingLocation}
+                key="holdingLocation"
```
dgcohen, line 260: Are these key props being used? They shouldn't be necessary if the elements aren't being rendered inside a loop.

pages/test-pickup-times/index.tsx
nonword: Whoops. Fixed

pages/test-pickup-times/index.tsx
nonword: I moved most of the util-like things into `src/utils/testPickupTimesUtils.ts`

pages/test-pickup-times/index.tsx
nonword: Cleaned this up

pages/test-pickup-times/index.tsx
nonword: Thanks. Fixed

dgcohen (approved):
Looks great!

PR #61, opened 2024-01-03
Scc 3661/details no subjects

charmingduchess:
Preliminary PR for bib details. Included in this work is the logic to render unstyled html for all of the details except for Annotated Marc and Subject Headings. 
I created a bib page to be able to see the details. You can see them by running the app locally and accessing this endpoint: http://localhost:8080/research/research-catalog/bib. bib details are generated from static fixtures. There are some options to render within that file. You can replace the parameter on line 11 of pages/bib/index with the other imported bibs to see some other kinds of data. 
I struggled writing tests for this because it really feels like I need to test for the specific presence of certain elements when certain data is provided. I did not finish the tests yet for this reason. Any advice is welcomed!


src/models/BibDetails.ts
```diff
@@ -0,0 +1,164 @@
...
+      link: "internal",
+      label: fieldMapping.label,
+      value: value.map((v: string) => {
+        const internalUrl = `/search?filters[${fieldMapping.field}][0]=${v}`
+        return { url: internalUrl, urlLabel: v }
+      }),
+    }
+  }
+
+  get extent(): BibDetail {
```
EdwinGuzman, line 46: What's "extent" in regards to a bib?

EdwinGuzman: I don't know how to verify this business logic but I know this code is mostly from DFE. Is there a doc I can read up on dimensions and extents?

src/models/modelTests/Bib.test.ts
```diff
@@ -0,0 +1,171 @@
+import {
+  bibWithSupplementaryContent,
+  noParallels,
+  parallelsBib,
+  yiddishBib,
+} from "../../../__test__/fixtures/bibFixtures"
+import BibDetailsModel from "../BibDetails"
+
+describe("Bib model", () => {
```
EdwinGuzman, line 9: This and the name of the file should be `BibDetails`.

EdwinGuzman (commented):
Minor comments but this looks good. I know there's more work on top of this in another PR so didn't comment on everything.

package.json
```diff
@@ -55,6 +55,7 @@
     "lint-staged": "^13.2.3",
     "next-router-mock": "^0.9.10",
     "prettier": "^2.8.8",
+    "react-element-to-jsx-string": "^15.0.0",
```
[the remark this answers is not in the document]
charmingduchess: its a package i was using for debugging. similar to enzyme's .html() or .debug() methods. i imagine i'll want to use it in the future, but if you think its clutter i can remove it.

src/models/BibDetails.ts
charmingduchess: extent is like how many pages a book is, and dimensions are the size of the book, like "30cm x 40cm". I don't think this is documented anywhere besides the ticket for this work. I can probably dig it up for you if you want to see it.

seanredmond: This is mostly the number of pages with some other details. I believe it's coming from the MARC 300 field, and what gets stored here as`extent` and `dimensions` are the $a and $b subfields. All the `removeSemiColon` business is because the subfields are cataloged with punctuation (semicolons) so that they can just be joined together, but we only show $a and $b and there is extra trailing punctuation if the record has $c etc and we don't remove it

EdwinGuzman (approved):

PR #62, opened 2024-01-03
SCC-3898 - Bib Page Api Route

dgcohen:
Ticket: https://jira.nypl.org/browse/SCC-3659

This PR adds the fetchBib function, the utils it depends on, as well as the default API route handler for debugging purposes.

It also includes the skeletal Bib page component with getServerSideProps, as well as some small miscellaneous changes (e.g. renaming the generic getQueryString to getSearchQuery).

Note that there is some functionality from DFE that hasn't been moved over but has been captured in the TODO comments and if they pertain more to the Bib Page model: https://jira.nypl.org/browse/SCC-3661


pages/api/bib/[id].ts
```diff
@@ -0,0 +1,41 @@
...
+ * Default API route handler for Bib page
+ *
+ * It handles a get request for a given bib id by internally calling fetchBib
+ * and handling the response based on the status returned there.
+ *
+ * Note: This is primarily used for debugging since the Bib page fetches the
+ * result by calling fetchBib directly in getServerSideProps.
+ */
+async function handler(req: NextApiRequest, res: NextApiResponse) {
+  const id = Array.isArray(req.query.id) ? req.query.id[0] : req.query.id
```
Comment (EdwinGuzman) [inline pages/api/bib/[id].ts:16]: In what scenario can `id` potentially be an array?

pages/bib/[id].tsx
```diff
@@ -0,0 +1,59 @@
...
+  const { bib, annotatedMarc, status, redirectUrl } = await fetchBib(
+    id,
+    bibParams
+  )
+
+  switch (status) {
+    case 301:
+      return {
+        redirect: {
+          destination: redirectUrl || PATHS["404"],
```
Comment (EdwinGuzman) [inline pages/bib/[id].tsx:40]: This is simple from Nextjs and works nicely.

pages/api/bib/index.ts
```diff
@@ -0,0 +1,72 @@
...
+      if (sierraBibResponse.statusCode === 200) {
+        return {
+          status: 301,
+          redirectUrl: `${appConfig.externalUrls.circulatingCatalog}/iii/encore/record/C__R${id}`,
+        }
+      } else {
+        new Error("There was a problem fetching the bib from Sierra")
+      }
+    }
+    // The Bib api in DFE added some additional pre-processing related to holdings and dimensions
```
EdwinGuzman, line 60: @charmingduchess might've done some of this in her recent PRs.

pages/api/bib/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  standardizeBibId,
+} from "../../../src/utils/bibUtils"
+import nyplApiClient from "../../../src/server/nyplApiClient"
+import {
+  DISCOVERY_API_NAME,
+  DISCOVERY_API_SEARCH_ROUTE,
+} from "../../../src/config/constants"
+import { appConfig } from "../../../src/config/config"
+
+export async function fetchBib(
```
EdwinGuzman, line 14: Because of the file directory and its location, one would think that `/research/research-catalog/api/bib` is a valid API route, but it's not. This is the location for the function used in the `.../api/bib/:id` route and in the Bib page's `getServerSideProps` and it makes sense and it's great to reuse this function in both places.

So I think this would be better in a different directory, perhaps `src/server`? Just an idea, let me know what you think. Otherwise, this file and function is fine. It does need tests, however, which can be covered in this PR or another. Mocking `nyplApiClient` can be difficult but this was my attempt to mocking it: https://github.com/NYPL/research-catalog/pull/65/files#diff-713db81b94eaf0bbf2724f5aa23fbd2139a9ba3c70fa11bb76fcf661b3dbdbf7R3

EdwinGuzman (commented):
Some comments but biggest is about the location of the `fetchBib` function and some tests to go along with it.

pages/api/bib/index.ts
charmingduchess, line 1: The fetchBib function needs to return the subject heading results as well. [Here's where it happens in DFE](https://github.com/NYPL/discovery-front-end/blob/development/src/server/ApiRoutes/Bib.js#L31).

pages/bib/[id].tsx
```diff
@@ -0,0 +1,59 @@
...
+import Layout from "../../src/components/Layout/Layout"
+import { PATHS, SITE_NAME } from "../../src/config/constants"
+import { fetchBib } from "../api/bib"
+import { mapQueryToBibParams } from "../../src/utils/bibUtils"
+
+/**
+ * The Bib page is responsible for fetching and displaying a single Bib's details.
+ */
+export default function Bib({ bib, annotatedMarc }) {
+  console.log(bib)
```
Comment (charmingduchess) [inline pages/bib/[id].tsx:13]: Should these console.logs be removed?

pages/bib/[id].tsx
```diff
@@ -0,0 +1,59 @@
...
+        redirect: {
+          destination: redirectUrl || PATHS["404"],
+          permanent: false,
+        },
+      }
+    case 404:
+      return {
+        redirect: {
+          destination: PATHS["404"],
+          permanent: false,
```
Comment (charmingduchess) [inline pages/bib/[id].tsx:48]: what does permanent signify here?

pages/api/bib/index.ts
charmingduchess, line 1: I'm thinking that the different redirect cases (circulating vs non standard bib id for research bib) should have different status codes. Maybe [303](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303) for the circ redirects?

pages/bib/[id].tsx
Comment (charmingduchess) [reply pages/bib/[id].tsx:40]: is there a known case where redirectUrl is null or undefined?

pages/bib/[id].tsx
Comment (dgcohen) [reply pages/bib/[id].tsx:48]: it determines the status code of the redirect (307 vs 308), this simply tells web clients/search engines not to cache the redirect forever.

pages/bib/[id].tsx
Comment (dgcohen) [reply pages/bib/[id].tsx:13]: They'll be removed eventually, I just added them to bypass the unused vars warning for now.

pages/api/bib/[id].ts
Comment (dgcohen) [reply pages/api/bib/[id].ts]: Good point. Any query param is allowed to be an array, but the ID will never be an array in our usage. I cast it as a string instead.

pages/api/bib/index.ts
dgcohen: I believe that checking for research-ness first would lead to a tradeoff where non-standardized IDs would lead to 2 fetches to the discovery API (first to see if it should redirect to circ, then another when we redirect to the correct URL). We could simply fetch the standardized version and update the URL with react's router but I think I would prefer to make it a proper redirect, unless you disagree.

pages/api/bib/index.ts
dgcohen: This is an interesting observation. My initial thought was to keep the fetching functions as close as possible to the route handlers, but they technically don't need to be there. Happy to move this as well as the other fetchers to the server directory.

pages/api/bib/index.ts
dgcohen: Cool, just removed this comment.

pages/api/bib/index.ts
charmingduchess: Does there have to be an index page in this directory? Is there any use case for a /api/bib route without a bib id?

dgcohen: @charmingduchess No there isn't. As soon as we move the fetcher functions to the /server directory (as part of a follow-up ticket) we will only need the [id] file.

pages/api/bib/[id].ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+async function handler(req: NextApiRequest, res: NextApiResponse) {
+  const id = req.query.id as string
+  const bibParams = mapQueryToBibParams(req.query)
+  const { bib, annotatedMarc, status, redirectUrl } = await fetchBib(
+    id,
+    bibParams
+  )
+
+  if (req.method === "GET") {
```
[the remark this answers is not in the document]
Comment (dgcohen) [reply pages/api/bib/[id].ts:23]: Good call, added a case for a POST statement with a message saying to use a GET req.

pages/api/bib/index.ts
```diff
@@ -0,0 +1,72 @@
...
+      const sierraBibResponse = await client.get(
+        `/bibs/sierra-nypl/${id.slice(1)}`
+      )
+      if (sierraBibResponse.statusCode === 200) {
+        return {
+          status: 301,
+          redirectUrl: `${appConfig.externalUrls.circulatingCatalog}/iii/encore/record/C__R${id}`,
+        }
+      } else {
+        new Error("There was a problem fetching the bib from Sierra")
```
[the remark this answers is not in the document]
dgcohen: Good "catch", added a 404 status return in this else statement with a console.error (added one to the catch statement as as well).

pages/api/bib/[id].ts
```diff
@@ -0,0 +1,41 @@
...
+ * It handles a get request for a given bib id by internally calling fetchBib
+ * and handling the response based on the status returned there.
+ *
+ * Note: This is primarily used for debugging since the Bib page fetches the
+ * result by calling fetchBib directly in getServerSideProps.
+ */
+async function handler(req: NextApiRequest, res: NextApiResponse) {
+  const id = req.query.id as string
+  const bibParams = mapQueryToBibParams(req.query)
+  const { bib, annotatedMarc, status, redirectUrl } = await fetchBib(
```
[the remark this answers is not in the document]
Comment (dgcohen) [reply pages/api/bib/[id].ts:18]: Just added a case that returns a 404 and logs a message when the bib attribute contains a string. Chatted with @danamansana about this and I will be creating a ticket to return a more semantic error response from the Discover y API.

pages/bib/[id].tsx
Comment (dgcohen) [reply pages/bib/[id].tsx]: Good call, remove the 404 fallback from this case.

EdwinGuzman (approved):

PR #64, opened 2024-01-04
SCC-3940 - Optimize test performance

dgcohen:
After looking more closely at our tests for advanced search, which were taking about 11 seconds to complete, I discovered that the performance bottleneck was mainly coming from our frequent use of getByRole, which takes about twice as long to run as getByLabelText.

I think it makes sense to use getByRole when we need to look for a specific tag or heading level, but for the sake of integration tests where we're primarily focused on app behavior, getByLabelText is probably sufficient.

I also removed act() clauses from several tests,  which weren't contributing to the performance issues, but were nevertheless not needed.

Note that there is a test for FieldSetDate that is currently failing but that is unrelated to these changes.

### DIFF
--- a/src/components/SearchFilters/FieldsetDate.test.tsx
+++ b/src/components/SearchFilters/FieldsetDate.test.tsx
@@ -4,7 +4,7 @@ import React from "react"
 import FieldsetDate from "./FieldsetDate"
 import userEvent from "@testing-library/user-event"
 
-describe("FieldsetDate", () => {
+describe.skip("FieldsetDate", () => {
   const onDateChange = jest.fn()
 
   it("should render the basic form", () => {


EdwinGuzman (commented):
Thanks! It is faster but still ~25-30s.

Just skip the `FieldsetDate` tests for now and this is good to go.

charmingduchess (approved):
🥳 love to see it.
@dgcohen i added one more comment, if you don't want to include it with this work, no worries.

__test__/pages/search/advancedSearchForm.test.tsx
charmingduchess, line 1: Though not strictly part of the scope of this ticket, could you take a look at the commented out test in this file? I think this test was timing out due to await/act business in combination with typing user input.

dgcohen: Yeah i can definitely take a look as part of a separate ticket.

PR #65, opened 2024-01-04
SCC-3665: Server auth log in through cookie parsing and JWT verification

EdwinGuzman:
Resolves [SCC-3665](https://jira.nypl.org/browse/SCC-3665)

- Uses the current DFE `initializePatronTokenAuth` and `getPatronData` functions as reference:
  - `initializePatronTokenAuth`: https://github.com/NYPL/discovery-front-end/blob/development/src/server/routes/auth/index.js
  - `getPatronData`: https://github.com/NYPL/discovery-front-end/blob/development/src/server/routes/api/index.js
- Includes an additional function for `getPatronData` to get the 268 field for notification preference.
- Adds tests for these functions which didn't exist before.
- Right now, the patron data is only available on the homepage by calling `getPatronData` in the homepage `getServerSideProps`. Every page would need to make this function call again because I'm not setting the patron data to a global object -- and maybe we don't want to?

**Implementation notes**
In DFE, both `initializePatronTokenAuth` and `getPatronData` run on every request to every page. The patron decoded information from the cookie is set in Express' `request` object. This is then used in `getPatronData` to fetch `/patrons/${userId}` and the returned patron data is set in Redux.

For this implementation, a nextjs middleware is set with defined paths on where to run. For now it's the homepage and the search pages for testing. Instead of setting a property on Nextjs' `request` object with the patron data, it is being sent through Nextjs headers. These headers are then read in `getPatronData` and used in the `/patrons/${userId}` fetch call.

The headers are `"x-is-token-valid"` and `"x-user-id"`. From what I understand, these headers are "non-standard" and cannot be read by browsers. I was not able to view them in the browser or elsewhere but I do need someone to verify this. Otherwise, this implementation will have to change. I'm also not sure how secure sending these values through headers is.

**How to test**
There's some information in the readme. Make sure to test on `local.nypl.org:8080/research/research-catalog`.
- Log into qa or prod DFE.
- Start the app locally and open the homepage.
- The logged in patron data will be logged to the console.
- **This won't work on Vercel as expected due to the bounds of the `nyplIdentityPatron` cookie**

**Other updates**
- For now, removed next's `Link` component usage. It was throwing an error.
- Upgraded to Next13 for its middleware implementation.
- Upgraded React to v18 to be compatible with Next13 (using this with DS is ok but does output warnings).
- Adds the `jose` package for JWT verification. The middleware uses the edge runtime and not node (this is annoying), so the `jsonwebtoken` package could not be used.
- Removed `act` calls because those were making tests fail.


src/components/RCLink/RCLink.tsx
```diff
@@ -30,22 +30,23 @@ const RCLink = ({
...
-        className={className}
-        fontWeight={active && "bold"}
-        {...rest}
-        __css={
-          hasWhiteFocusRing && {
-            _focus: {
-              outlineColor: "ui.white",
-            },
-          }
+    // <Link href={href} passHref>
```
EdwinGuzman, line 33: This is temporary. The Nextjs and DS Link components aren't working together.

charmingduchess: ruh roh. is that something to do with the middleware?

src/components/SubNav/SubNav.tsx
charmingduchess, line 1: Are these BASE_URL updates to do with the link issue mention earlier?

src/config/config.ts
```diff
@@ -58,4 +58,13 @@ export const appConfig = {
   nonRecapClosedLocations: parseLocations(
     process.env.NEXT_PUBLIC_NON_RECAP_CLOSED_LOCATIONS
   ),
+  publicKey: `-----BEGIN PUBLIC KEY-----
```
charmingduchess, line 61: can this have a more descriptive name? like jwtPublicKey?

charmingduchess:
@EdwinGuzman Thanks for a thorough yet concise description of the PR! I'm going to work through this but I might need a walk thru some time next week.

src/server/getPatronData.ts
```diff
@@ -0,0 +1,105 @@
...
+              barcodes,
+              emails,
+              loggedIn: true,
+              moneyOwed,
+              homeLibraryCode,
+              patronType,
+              expirationDate,
+              phones,
+              noticePreference,
+            }
```
nonword, line 89: I think `patron` needs to be returned?

middleware.ts
```diff
@@ -0,0 +1,31 @@
...
+    },
+  })
+
+  return response
+}
+
+// EG: This is a bit annoying but the catch-all regex did not work for me.
+// We might have to manually include all routes individually.
+export const config = {
+  matcher: ["/", "/search/:path*"],
```
nonword, line 30: Should "/research/research-catalog" be among the routes for testing?

src/components/RCLink/RCLink.tsx
EdwinGuzman: With the DS. I'll update this independently.

src/components/SubNav/SubNav.tsx
EdwinGuzman: Yes. I commented out the Nextjs `Link` component which already knew that the base url was `/research/research-catalog`. So now this needs to be added manually.

middleware.ts
EdwinGuzman: Yes, I'll add it. Next is doing `/` based on the base url in its config which is already `/research/research-catalog`. So it knows the right "home" route, but it doesn't hurt to add.

src/server/getPatronData.ts
charmingduchess: I'm not sure that we actually want to return patron data here. The plan for My Account is to do fetches from Sierra instead of platform.

EdwinGuzman: Hmm true. Then maybe all this PR needs to do is the set whether the parsed and jwt verified token is valid since the patron id is there and this entire function is not needed.

middleware.ts
```diff
@@ -0,0 +1,31 @@
+// write middleware file for nextjs
+import type { NextRequest } from "next/server"
```
dgcohen, line 2: Small nit but I believe these 2 lines can be consolidated as

`import { NextResponse, type NextRequest } from "next/server"`

middleware.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Middleware to check and parse the nyplIdentiyPatron cookie and then verify
+ * it through JWT. The decoded patron information is then set in Nextjs'
+ * request headers.
+ */
+export async function middleware(request: NextRequest) {
+  const requestHeaders = new Headers(request.headers)
+  const patronTokenResponse = await initializePatronTokenAuth(request)
+
+  requestHeaders.set("x-is-token-valid", `${patronTokenResponse?.isTokenValid}`)
+  requestHeaders.set("x-user-id", `${patronTokenResponse?.decodedPatron?.sub}`)
```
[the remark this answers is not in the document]
EdwinGuzman: I was trying to not call this function on every `getServerSideProps` function/page but it seems we might need to. Not that this is bad, but I figured it could be done once in the middleware. I'll try again to make it part of a the local req or extend it instead of the custom headers.

dgcohen (approved):
This looks great as a first step. Really appreciate the thorough annotations and the work you put in mocking the async request.

Not 100% confident evaluating the security of the http headers, will have to read more about it independently, but this looks great for now.

src/server/getPatronData.ts
```diff
@@ -0,0 +1,105 @@
...
+              id,
+              names,
+              emails,
+              moneyOwed,
+              homeLibraryCode,
+              patronType,
+              expirationDate,
+              phones,
+            } = response.data
+            const barcodes = response.data.barCodes
```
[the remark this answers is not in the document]
EdwinGuzman: true, but now we don't need this file so I'm going to delete it.

EdwinGuzman:
@charmingduchess @nonword @dgcohen updated based on feedback, thanks!

charmingduchess (approved):
Thanks a lot Edwin! Super quick turn around. Much appreciated 🙏🏻

nonword (approved):
This looks good. I wonder if a wrapper function like `isPatronLoggedIn` may be a clearer name to expose to pages, but I understand this is just to get the basic wiring down

PR #67, opened 2024-01-09
SCC-3790 - Adobe Analytics Page View Tracking

dgcohen:
Ticket: https://jira.nypl.org/browse/SCC-3790

This PR adds virtual page view tracking for Adobe Analytics derived from the configuration in DFE.

One notable difference is that the data layer object doesn't need to be cleared on every route change in Next.js since it gets cleared automatically. This could mean that we can use standard page view events instead of "virtual" page views in the future, but it's currently sending the same events as DFE.

There are a few other minor changes, mostly based on the formatting of the route in Next.js vs vanilla js, but in most cases this led to simplification of the code. Also note the removal of the SHEP routes in the event name generator (this will have to be reimplemented when we introduce Enhanced Browse).


__test__/pages/search/advancedSearchForm.test.tsx
```diff
@@ -93,7 +93,7 @@ describe("Advanced Search Form", () => {
       screen.getByLabelText("Language"),
       "Azerbaijani"
     )
-    const keywordInput = screen.getByLabelText("Keywords")
+    const keywordInput = screen.getByLabelText("Keyword")
```
dgcohen, line 96: This just fixes a failing text in main. Not related to this pr.

pages/_document.tsx
```diff
@@ -17,6 +18,18 @@ export default function Document() {
       </style>
       <Head>
         <meta name="description" content="Research Catalog | NYPL" />
+        <script async src={appConfig.adobeEmbedUrl[appConfig.environment]} />
```
EdwinGuzman, line 21: Should this use the nextjs `NextScript` component?

pages/_document.tsx
```diff
@@ -17,6 +18,18 @@ export default function Document() {
...
+        <script
+          dangerouslySetInnerHTML={{
+            __html: `
+              // First define the global variable for the entire data layer array
+              window.adobeDataLayer = window.adobeDataLayer || [];
+              // Then push in the variables required in the Initial Data Layer Definition
+              window.adobeDataLayer.push({disable_page_view: true});
+              ${appConfig.environment}
```
EdwinGuzman, line 29: Why is this here?

src/utils/appUtils.ts
```diff
@@ -0,0 +1,100 @@
...
+  return ADOBE_ANALYTICS_RC_PREFIX + pageName
+}
+
+/**
+ * Tracks a virtual page view to Adobe Analytics on page navigation.
+ */
+export const trackVirtualPageView = (pathname = "", queryParams = "") => {
+  const adobeDataLayer = window["adobeDataLayer"] || []
+  const route = pathname.toLowerCase()
+  adobeDataLayer.push({
```
EdwinGuzman, line 95: I was checking DRB's implementation and I forgot about this but just want to verify if this should also clear out the array before pushing new values: https://github.com/NYPL/sfr-bookfinder-front-end/blob/5bac95879e71242daeab7c7b9d384fcb327d6244/src/lib/adobe/Analytics.ts#L30

EdwinGuzman (commented):
Some questions but looks good overall.

pages/_document.tsx
dgcohen: Good catch, had that in there for debugging and forgot to remove

src/utils/appUtils.ts
dgcohen: That was the original approach, but leaving that in actually causes an additional page view event to fire. Next is clearing the data layer stack on every page view on its own so I believe that is no longer needed.

dgcohen: Actually, the double firing turned out to be caused by something else. Readded the page view clearing logic.

pages/_app.tsx
```diff
@@ -1,8 +1,51 @@
...
+  //
+  //   // If the component is unmounted, unsubscribe
+  //   // from the event with the "off" method
+  //   return () => {
+  //     router.events.off("routeChangeComplete", handleRouteChange)
+  //   }
+  // }, [router.events])
+
+  // Track page view events to Adobe Analytics
+  useEffect(() => {
```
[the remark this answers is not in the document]
dgcohen: Some context: I originally had that but it was causing a double trigger of the useEffect().

The script above this that is commented out addresses this by sending an initial page view event on app load, and subsequent page views via a useEffect with router.events in the dependency array. What is uncommented is a solution from DRB's implementation that works as long as there is no re-rendering of the App component via state changes.

I'm more in favor of the explicit solution but we're set up some alerting to ensure the number of page views are aligned for each method when we move this to production.

pages/_document.tsx
dgcohen: Good call, I replaced this with a <Script> tag.

pages/_app.tsx
```diff
@@ -1,10 +1,67 @@
...
+    trackVirtualPageView(router.asPath)
+  })
+
   return (
     <>
+      <Script async src={appConfig.adobeEmbedUrl[appConfig.environment]} />
+      <Script id="adobeDataLayerDefinition">
```
EdwinGuzman, line 55: I noticed this was moved from _document to _app. I think scripts set and called in _document only get called once. I don't think that's true in _app so this doesn't seem right. It's one reason why the header/footer are set in _document but do verify this.

EdwinGuzman:
@avertrees

EdwinGuzman (approved):

PR #68, opened 2024-01-11
Scc 3938/bib details styling

charmingduchess:
Add styles to bib details. 
Most additions are overrides for DS List component.Are there any DS List props for not rendering the default html styling but leaving the DS CSS? noStyling seems to remove all styles including DS, but I just want to remove the numbering and indentation from the `<ol>` elements.


EdwinGuzman:
> Are there any DS List props for not rendering the default html styling but leaving the DS CSS? noStyling seems to remove all styles including DS, but I just want to remove the numbering and indentation from the <ol> elements.

No, there isn't a prop for this. I believe `noStyling` was intended to work that way and then the consuming app would add overrides but this makes me think it should have an update. I'll take a closer look at this PR and for visibility, tagging @bigfishdesign13

src/components/BibPage/BibDetail.tsx
```diff
@@ -0,0 +1,142 @@
...
+  details: BibDetail[] | LinkedBibDetail[]
+  heading?: string
+}
+
+const BibDetails = ({ details, heading }: BibDetailsProps) => {
+  return (
+    details?.length > 0 && (
+      <>
+        {heading && <Heading level="three">{heading}</Heading>}
+        <List
```
EdwinGuzman, line 26: The `List` component allows a `title` prop that renders an h2 heading.

The latest DS version [2.1.4](https://www.npmjs.com/package/@nypl/design-system-react-components/v/2.1.4) allows for the title to be a custom component. So this can be:
```
<List
  title={heading && <Heading level="three">{heading}</Heading>}
  ...
```

Can you try this default `List` style for the heading and the row of dt/dds? This would mean removing the `sx={{ borderBottom: "none" }}` rule.

I'd like to see what Apporva/Marty think about this update on the bib page. If it doesn't work out and we want the older styles, that's fine but we should try to use the default DS styles.

src/components/BibPage/BibDetail.tsx
```diff
@@ -0,0 +1,142 @@
...
+    )
+  )
+}
+
+const buildDetailElement = (field: BibDetail) => {
+  return (
+    <>
+      <dt>{field.label}</dt>
+      <dd>
+        <List data-testId={`${field.label}`} type="ol">
```
EdwinGuzman, line 56: Regarding the question in the PR's description: why the use of `ol`? The current page uses `ul` and I'm not sure if the values for each field should be numbered.

styles/components/BibDetails.module.scss
```diff
@@ -0,0 +1,14 @@
+@import "../utils/mixins";
+
+.bibDetails {
+  dl > dt,
+  dl > dd {
+    border: none;
```
EdwinGuzman, line 6: Can you leave the default border to see what Apoorva thinks?

src/components/BibPage/BibDetail.tsx
charmingduchess: I'm using ol because the order matters for the primary/parallel matches. but maybe i'm misinterpreting the use case for an ol. happy to use whatever!

styles/components/BibDetails.module.scss
charmingduchess: I spoke with @apoorvakhanolkar earlier about this and she requested that it be removed.

charmingduchess: Here are screenshots of the two different styling options. @apoorvakhanolkar can you confirm which is preferable?
Screen Shot 2024-01-12 at 9 36 45 AM
Screen Shot 2024-01-12 at 9 36 35 AM

charmingduchess: I think the border looks classy 😎 . Looks much better now with the headings too, imo.

apoorvakhanolkar: I prefer it without the borders! Just looks a bit cleaner to me.

EdwinGuzman (approved):
Looks good! I'd wait on Diego's review but I'm curious, would the order of merge create conflicts with this and the Bib page route PR?

src/models/BibDetails.ts
```diff
@@ -0,0 +1,209 @@
...
+        let detail
+        if (fieldMapping.field === "supplementaryContent")
+          detail = this.supplementaryContent
+        else if (fieldMapping.field === "creatorLiteral")
+          detail = this.buildInternalLinkedDetail(fieldMapping)
+        else detail = this.buildStandardDetail(fieldMapping)
+        if (!detail) return
+        return detail
+      })
+      .filter((f) => f)
```
dgcohen, line 114: Out of curiosity, what are these filters doing?

src/utils/bibModelPreprocessing.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * interleaveParallel(arr1, arr2)
+ * Given two arrays, returns the elements interleaved, with falsey elements removed.
+ * Also combines data from matching elements when necessary.
+ * Example: interleaveParallel ([1, 2, null, 3], [5,6,7,8,9]) =>
+ * [1,5,2,6,7,3,8,9].
+ * Assumes that arr2 is at least as long as arr1.
+ *
+ * @param {array} arr1
```
dgcohen, line 63: Don't think we need to be using JS doc type annotations for this project since we're using Typescript, but let me know if you disagree.

src/models/BibDetails.ts
charmingduchess: filtering out null values. If a bib doesn't have a supplementary content, for example, we don't want there to be a null value in the array of details.

src/utils/bibModelPreprocessing.ts
charmingduchess: True! this was just an artifact from DFE.

charmingduchess:
@EdwinGuzman  it's possible. is there part of the code you're concerned about in particular? i did create and then rename a `pages/bib.tsx` file, but i don't think it'll be too hard to parse out if there's a conflict.

EdwinGuzman:
Nothing in particular that I'm concerned about, but it would be good to consider -- someone may end up with merge conflicts and that's okay but hope nothing gets overridden accidentally.

src/components/BibPage/BibDetail.tsx
```diff
@@ -0,0 +1,129 @@
...
+              return buildPlainTextElement(detail as BibDetail)
+            }
+          }
+        )}
+      </List>
+    )
+  )
+}
+
+const buildDetailElement = (label: string, listChildren: ReactNode[]) => {
```
dgcohen, line 51: Small nit, but what do you think of renaming these to DetailElement, PlainTextElement and CompoundSubjectHeadingElement? I know they're technically functions but I'd rather keep the naming consistent with other react elements.

src/components/BibPage/BibDetail.tsx
```diff
@@ -0,0 +1,129 @@
...
+      </dd>
+    </>
+  )
+}
+
+const buildPlainTextElement = (field: BibDetail) => {
+  const values = field.value.map((val: string, i: number) => {
+    const stringDirection = displayRtl(val)
+    return (
+      <li dir={stringDirection} key={i}>
```
dgcohen, line 68: This key might need a prefix to avoid duplicates.

src/components/BibPage/BibDetail.tsx
```diff
@@ -0,0 +1,129 @@
...
+    </li>
+  ))
+  return buildDetailElement(field.label, values)
+}
+
+const buildSingleSubjectHeadingElement = (subjectHeadingUrls: Url[]) => {
+  const urls = subjectHeadingUrls.reduce((linksPerSubject, url: Url, index) => {
+    const divider = (
+      // this span will render as > in between the divided subject heading links
+      <span data-testid="divider" key={`divider-${index}`}>
```
dgcohen, line 94: Can this be achieved with CSS (pseudoselector with a content attribute?). Not sure if it can but curious.

src/components/BibPage/BibDetail.tsx
```diff
@@ -0,0 +1,129 @@
...
+    }
+    return linksPerSubject
+  }, [] as React.JSX.Element[])
+  return urls
+}
+
+const buildLinkedElement = (field: LinkedBibDetail) => {
+  const internalOrExternal = field.link
+  const values = field.value.map((urlInfo: Url, i) => {
+    return <li key={i}>{linkElement(urlInfo, internalOrExternal)}</li>
```
dgcohen, line 112: This key might also need a prefix/more unique value

src/models/BibDetails.ts
```diff
@@ -0,0 +1,202 @@
...
+      { field: "notes", label: "Notes" },
+      { field: "tableOfContents", label: "Contents" },
+      { field: "shelfMark", label: "Call Number" },
+      { field: "isbn", label: "ISBN" },
+      { field: "issn", label: "ISSN" },
+      { field: "oclc", label: "OCLC" },
+      { field: "lccn", label: "LCCN" },
+    ]
+      .map((fieldMapping) => {
+        let detail
```
dgcohen, line 78: I believe you don't need to instantiate and assign to a variable, you can just return from within the if statements.

src/models/BibDetails.ts
```diff
@@ -0,0 +1,202 @@
...
+        const filterQueryForSubjectHeading = "/search?filters[subjectLiteral]="
+        // splitSubjectHeadings: ["a", "b", "c"]
+        const splitSubjectHeadings = subject.split(" -- ")
+        return splitSubjectHeadings.map((heading, index) => {
+          const urlWithFilterQuery = `${filterQueryForSubjectHeading}${stackedSubjectHeadings[index]}`
+          const subjectHeadingUrl = {
+            url: urlWithFilterQuery,
+            urlLabel: heading,
+          }
+          return subjectHeadingUrl
```
dgcohen, line 182: Again you probably don't need to instantiate a variable, you can just return the line above.

src/models/BibDetails.ts
```diff
@@ -0,0 +1,202 @@
...
+    if (!holdings) return []
+    return [
+      { label: "Location", field: "location" },
+      { label: "Format", field: "format" },
+      { label: "Call Number", field: "shelfMark" },
+      { label: "Library Has", field: "holdingStatement" },
+      { label: "Notes", field: "notes" },
+    ]
+      .map((fieldMapping) => {
+        const detail = this.buildHoldingDetail(fieldMapping)
```
dgcohen, line 29: You can also return this directly without instantiating a variable.

src/models/BibDetails.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  constructSubjectHeadingsArray(subject: string) {
+    let currentArrayString = ""
+
+    return subject.split(" -- ").map((urlString, index) => {
+      const dashDivided = index !== 0 ? " -- " : ""
+      currentArrayString = `${currentArrayString}${dashDivided}${urlString}`
+
+      return currentArrayString
```
dgcohen, line 199: Same comment about unneeded variable instantiation.

src/utils/bibDetailUtils.ts
```diff
@@ -0,0 +1,7 @@
+export const displayRtl = (value: string) => {
```
dgcohen, line 1: I think these functions could benefit from comments and more explicit naming. I also don't mind them moving to the bibUtils file to cut down on the number of utils files.

src/utils/bibModelPreprocessing.ts
```diff
@@ -0,0 +1,101 @@
...
+   * getNoteType(note)
+   * Construct label for a note by adding the word 'Note'
+   */
+  const getNoteType = (note: Note) => {
+    const type = note.noteType || ""
+    return type.toLowerCase().includes("note") ? type : `${type} (note)`
+  }
+
+  if (!note) {
+    return notesGroupedByNoteType
```
dgcohen, line 17: You can move this towards the top of the function and return an empty object when it's not set. This makes it more legible/explicit since I think this implies that notesGroupedByNoteType contains something other than an empty object at this point.

src/utils/bibModelPreprocessing.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const getGroupedNotes = (bib: SearchResult) => {
+  const note = bib?.note?.length ? bib.note : null
+  let notesGroupedByNoteType = {}
+
+  /**
+   * getNoteType(note)
+   * Construct label for a note by adding the word 'Note'
+   */
+  const getNoteType = (note: Note) => {
```
dgcohen, line 11: I think this should be defined outside of this function so it's not instantiated every time it runs.

src/utils/bibModelPreprocessing.ts
dgcohen: It also might not need to be instantiated at all if you return the note.filter.reduce statement.

src/utils/bibModelPreprocessing.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  return notesGroupedByNoteType
+}
+
+/**
+ * combineMatching(el1, el2)
+ * Combines properties from matching (i.e. parallel) elements as necessary
+ * Right now, this is only needed to add the 'noteType' in case of parallel notes
+ * @param {string} el1
```
dgcohen, line 42: Would be good to replace all JS doc annotations with typescript.

src/components/BibPage/BibDetail.tsx
charmingduchess: sounds good to me!

src/models/BibDetails.ts
charmingduchess: ah yes thats leftover from debugging

charmingduchess:
@dgcohen thanks for the close eye to the functions!

dgcohen (approved):
Looks great!

PR #69, opened 2024-01-12
SCC-3943: Integrate special delivery schedules

nonword:
Integrates @danamansana 's [logic for adjusting delivery estimates for special scholar rooms](https://github.com/NYPL/discovery-front-end/pull/2137)


src/utils/pickupTimeEstimator.ts
```diff
@@ -117,30 +122,58 @@ export const getPickupTimeEstimate = async (
...
+    arrivalAtHoldshelf = await addOnsiteTravelDuration(
+      destinationServiceTime,
+      deliveryLocationId
     )
+    rationale.push({ time: arrivalAtHoldshelf, activity: "Onsite travel time" })
+  }
+
+  if (hasSpecialDeliverySchedule) {
+    let specialDeliveryRationale
+    ;({ arrivalAtHoldshelf, rationale: specialDeliveryRationale } =
```
dgcohen, line 161: You shouldn't need the semicolon and parens here.

src/utils/pickupTimeEstimator.ts
```diff
@@ -200,27 +233,128 @@ export const addOnsiteTravelDuration = async (serviceTime, locationId) => {
...
+    return new Date(date.getTime() - offsetMs + specialFudgeMs).toISOString()
+  }
+
+  const rationale = []
+
+  if (secondFloorScholarRooms.includes(locationId)) {
+    hasSpecialDeliverySchedule = true
+    getFirstHour = (time) => {
+      const day = time.getDay()
+      return day === 0 ? 14 : 10
```
dgcohen, line 283: I think it might make sense to have a constant that maps day keys to labels just to make things more legible. I also wouldn't mind having closing times in named constants either.

Something like

const DAYS = {
   MONDAY: 0
   TUESDAY: 1
}
etc

src/utils/pickupTimeEstimator.ts
nonword: I get "Declaration or statement expected. This '=' follows a bloc...ent, you might need to wrap the whole assignment in parentheses" without them

dgcohen (approved):
Looks great, thanks for the changes!

PR #73, opened 2024-01-19
rearrange bibdetails types and preprocessing

charmingduchess:
- rename and move SearchResult type to Bib type. I think this will clear up some of the confusion around the types/models. In reality, DAPI returns the same kinds of objects whether its the resources/bibid route or as a search result
- extract bib details types into their own file. They really refer to types that are only used by the bib details components.
- move bib detail preprocessing into the BibDetails class. 
- add owning institution detail to BibDetails' bottom details


src/models/BibDetails.ts
```diff
@@ -113,6 +117,100 @@ export default class BibDetailsModel {
...
+  getNoteType = (note: Note) => {
+    const type = note.noteType || ""
+    return type.toLowerCase().includes("note") ? type : `${type} (note)`
+  }
+
+  /**
+   * Combines properties from matching (i.e. parallel) elements as necessary
+   * Right now, this is only needed to add the 'noteType' in case of parallel notes
+   */
+  combineMatching(primaryValue: string, parallelValue: string | Note) {
```
dgcohen, line 155: Small nit, but I'm thinking combineMatching, interleaveParallel and matchParallel's could use more descriptive fields? Even if it's just interleaveParallelFields etc?

dgcohen:
Thanks so much for this! This looks way better. Just one small suggestion about naming but otherwise I think this is a big improvement.

charmingduchess:
@dgcohen good 2 go?

dgcohen (approved):
Looks great! Thanks

PR #74, opened 2024-01-23
Scc 3973/annotated marc

charmingduchess:
- add annotated marc details to the bottom details
  - including logic to remove any glaring duplicates
- refactor groupedNotes to return data structures that match other details
- remove groupedNotes from the field mappings array in bottomDetails 
- add owner detail
- incorporate bibDetails into the bib page for dynamic bib data action 😎 
- updated default environment to "development" instead of "production"


src/models/BibDetails.ts
```diff
@@ -3,18 +3,39 @@ import type {
...
-  constructor(bib: Bib) {
+  bib: Bib
+  annotatedMarcFields: AnnotatedMarcField[]
+  constructor(bib: Bib, annotatedMarc?: AnnotatedMarc) {
     this.bib = this.matchParallelToPrimaryValues(bib)
+    this.annotatedMarcFields = annotatedMarc?.fields
   }
 
-  get holdingsDetails() {
+  get annotatedMarcDetails(): AnyBibDetail[] {
```
dgcohen, line 21: Could annotatedMarcDetails be changed from a getter function to a class property? I was thinking the code below could be moved into a function that parses the annotatedMarc attribute once on instantiation rather than each time the getter is called (see getItemsFromResult in SearchResultsBib)

src/server/nyplApiClient/index.ts
```diff
@@ -8,8 +8,7 @@ interface KMSCache {
   clientId: string
   clientSecret: string
 }
-
-const appEnvironment = process.env.APP_ENV || "production"
+const appEnvironment = process.env.APP_ENV || "development"
```
dgcohen, line 11: I believe we still want the fallback to be production, is this here for local testing?

src/server/nyplApiClient/index.ts
```diff
@@ -47,7 +46,7 @@ const nyplApiClient = async (options = { apiName: "platform" }) => {
   }
 
   const baseUrl = appConfig.apiUrls[apiName][appEnvironment]
-
+  console.log({ apiName, appEnvironment })
```
dgcohen, line 49: This console log should be removed.

src/models/BibDetails.ts
```diff
@@ -3,18 +3,39 @@ import type {
...
+  Url,
+  SubjectHeadingDetail,
+  AnnotatedMarc,
+  AnyBibDetail,
 } from "../types/bibDetailsTypes"
 
-interface BibDetails extends Bib {
-  groupedNotes: object
-}
 export default class BibDetailsModel {
```
dgcohen, line 13: I know this went in previously, but could you change the name of the class to BibDetails to follow the naming convention of the other models?

src/models/BibDetails.ts
dgcohen: Furthermore, I don't think we would need annotatedMarcFields to be a class property since it could just be passed as an argument in the preprocessor function.

src/models/BibDetails.ts
```diff
@@ -3,18 +3,39 @@ import type {
...
+          return this.buildExternalLinkedDetail("Connect to:", urlValues)
+        } else {
+          const fieldValues = values.map((val) => val.content)
+          return this.buildDetail(label, fieldValues)
+        }
+      }
+    )
+  }
+
+  get holdingsDetails(): BibDetail[] {
```
dgcohen, line 38: I think my note about annotatedMarcDetails can apply to the "details" getters as well.

dgcohen (changes requested):
Looks good! Just a few notes.

src/server/nyplApiClient/index.ts
charmingduchess: I must have misunderstood our conversation. I thought we agreed on this change but maybe it was just temporarily for debugging.

src/models/BibDetails.ts
charmingduchess: updated!

src/server/nyplApiClient/index.ts
charmingduchess: i returned it to production

dgcohen (approved):
Thanks! Looks great.

PR #77, opened 2024-01-26
SCC-3968 - Fetch subject headings in Bib fetch

dgcohen:
This PR adds a function fetchBibSubjectHeadings that fetches a bib's subject headings with a timeout of 4 seconds to match DFE. It then assigns the subject headings to the bib on success.

I also changed the references to standardizedId in the bib fetcher to id, since the value of id will never be "non-standardized" after the redirect to the standardized version. This behavior will be safeguarded once tests are added for the fetchers, but I think this makes the code a bit more legible for now.


EdwinGuzman (approved):
This looks good and works locally for me.

PR #78, opened 2024-01-26
SCC-3791: Adding tests for all DRB related components and functions.

EdwinGuzman:
This results [SCC-3791](https://jira.nypl.org/browse/SCC-3791)

- This adds unit tests for two DRB components.
- Removes a DRB-related component that seemed like a duplicate but was not being used.
- Adds unit tests for DRB utility functions.
- Adds unit test for the main search function and the logic used for the `/api/drb`
- This also fixes the rendering issue in the sidebar on the search results page [SCC-3882](https://jira.nypl.org/browse/SCC-3882)

I also attempted some very basic error handling for the `/api/drb` endpoint:

Screen Shot 2024-01-26 at 5 04 38 PM

I tried to make it resolve it as a "problem detail" but we should chat about how we want to structure error objects returned from the API endpoints.


src/models/modelTests/DRBResult.test.ts
```diff
@@ -0,0 +1,33 @@
...
+describe("DRBResult model", () => {
+  const model = new DRBResult(works[0] as DRBWork)
+
+  it("get basic DRB information for a work", () => {
+    expect(model.id).toEqual("84306f57-3285-403a-bff4-3d8286d98183")
+    expect(model.title).toEqual("Anatomy of the cat")
+    // Authors still contains all the additional properties but are
+    // not used and not defined in the model or Author interface
+    expect(model.authors).toEqual([
+      { name: "Reighard, Jacob (1861-)", lcnaf: "", primary: "true", viaf: "" },
```
EdwinGuzman, line 17: I noticed that even though the `Author` interface is defined as `{ name: string }`, all these values are still passed to the object. This is also true for the general model object. Perhaps we can be pickier and `Pick<....>` only the properties we want from the API data response.

pages/search/index.tsx
```diff
@@ -78,7 +78,7 @@ export default function Search({ results }) {
         activePage="search"
         sidebar={
           <>
-            {totalResults && (
+            {totalResults > 0 ? (
```
EdwinGuzman, line 81: @dgcohen this fixes https://jira.nypl.org/browse/SCC-3882

charmingduchess (approved):

dgcohen (approved):
This looks good!

PR #79, opened 2024-01-29
SCC-3838 - Fix empty search query

dgcohen:
This PR fixes an issue where an empty search would yield no results.

This was caused by the query string falling back to "" instead of "?" when search params are absent. I also set a default q value of "" when it is undefined so that it displays as the search term instead of "undefined".

### DIFF
--- a/pages/api/search.ts
+++ b/pages/api/search.ts
@@ -25,7 +25,6 @@ export async function fetchResults(
   searchParams: SearchParams
 ): Promise<SearchResultsResponse | Error> {
   const { q, field, filters } = searchParams
-
   // If user is making a search for bib number (i.e. field set to "standard_number"),
   // standardize the bib ID and pass it as the search keywords
   const keywordsOrBibId = field === "standard_number" ? standardizeBibId(q) : q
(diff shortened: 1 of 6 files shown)


EdwinGuzman (commented):
I think you don't want the `?` to go into the url. Production DFE goes to `/research/research-catalog/search` but in this PR the url goes to `/research/research-catalog/search?` on empty searches.

In vercel, this request takes a long time and I consistently get 504 "This serverless function has timed out" _from the homepage_.

If you start out at the search page having already made a search and then clear the field and make another search, it _usually_ works fine.

dgcohen:
> I think you don't want the `?` to go into the url. Production DFE goes to `/research/research-catalog/search` but in this PR the url goes to `/research/research-catalog/search?` on empty searches.
> 
> In vercel, this request takes a long time and I consistently get 504 "This serverless function has timed out" _from the homepage_.
> 
> If you start out at the search page having already made a search and then clear the field and make another search, it _usually_ works fine.

Good catch! Just moved the query string fallback to the fetcher function and added a timeout increase for the function in Vercel following the instructions in their docs, lets see if it works.

EdwinGuzman (approved):
One minor update.

PR #95, opened 2024-02-08
SCC-3996: Adding notification component on the home and search result pages.

EdwinGuzman:
Resolves ticket [SCC-3996](https://jira.nypl.org/browse/SCC-3996)

This PR adds the DS `Notification` component to render a custom notification banner with hard-coded values only on the home and search result pages. This follows the same pattern as DFE and uses the same env var name for the banner content `NEXT_PUBLIC_SEARCH_RESULTS_NOTIFICATION` but needs the `NEXT_PUBLIC_` prefix to load correctly.

I added this env var to Vercel so it would render in previews.


src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * pages and that is controlled in the `Layout` component.
+ */
+const Notification = ({ notification = "" }) => {
+  if (!notification) return null
+
+  return (
+    <DSNotification
+      aria-label="Research Catalog Alert"
+      notificationHeading={
+        <Heading level="h2">New Service Announcement</Heading>
```
EdwinGuzman, line 19: Made this an h2 but it still _displays_ as an h4. This was done to meet the accessibility heading hierarchy:

Screen Shot 2024-02-08 at 5 31 14 PM

src/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,32 @@
...
+      }
+      notificationContent={
+        <div dangerouslySetInnerHTML={{ __html: notification }} />
```
charmingduchess, line 22: why are you using `dangerouslySetInnerHtml` here?

.env.example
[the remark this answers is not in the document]
EdwinGuzman: Yes. I'm going to redo some of this PR to better accommodate this.

src/components/Notification/Notification.tsx
EdwinGuzman: In case the notification is not a string and includes html. I guess we don't have notifications such as `Urgent: the library is <b>closed</b>`, so this could just be a string.

EdwinGuzman:
Updated this to include prop drilling of the env var from `getServerSideProps` to the banner notification in the UI.

src/components/Notification/Notification.tsx
charmingduchess: We actually do use html in the notification env var! Specifically the <a> tag is an example that comes to mind. That said, I don't think the notification ends up being anything but a string, given that it's historically been input in the AWS console.

ENVIRONMENTVARS.md
```diff
@@ -20,15 +20,16 @@ If an environment variable is updated, make sure to restart the server for the a
...
+| `SEARCH_RESULTS_NOTIFICATION`            | string  | "Due to winter holiday closures, the delivery time for off-site requests will be delayed..." | A string that can including html that will be rendered as a notification on the Home and Search Results pages.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 |
```
7emansell, line 32: nitpick: typo including -> include

7emansell (approved):
Lgtm except that SEARCH_RESULTS_NOTIFICATION sort of confused me as the name for that env variable (it seems like it wouldn't show up on Home/Search pages on open, only once the user has searched..?) but that might just be me

EdwinGuzman:
@7emansell I agree. It is confusing but matches the current env var name on production DFE. I don't want to change the name until we have more of this new app on production because it might be confusing when we do have a notification to display to have to update two different variable names.

src/components/Layout/Layout.tsx
charmingduchess, line 1: Can you rename `bannerNotification` to be `searchResultsNotification`? i imagine that once there is a hold request page we'll have a similar mechanism to pass the hold request notification  to that page.

charmingduchess: Ok I see now this is agnostic so it can be reused for hold request notification!

charmingduchess:
looks good! i agree with emma that the name could change (once 2ad is over and RC is default discovery layer).

charmingduchess (approved):

PR #96, opened 2024-02-09
SCC-3992: Add Subnav login/logout

7emansell:
Resolves ticket [SCC-3992](https://jira.nypl.org/browse/SCC-3992).

Adds Log out link to subnav which displays if user is logged in, adds redirect from Account page to log in, creates login and logout redirects that return user to the page they started on (or back to home if they logged out from the account/hold page).


.env.example
```diff
@@ -5,4 +5,5 @@ NEXT_PUBLIC_CLOSED_LOCATIONS=""
 NEXT_PUBLIC_RECAP_CLOSED_LOCATIONS=""
 NEXT_PUBLIC_NON_RECAP_CLOSED_LOCATIONS=""
 ADOBE_EMBED_URL=https://assets.adobedtm.com/1a9376472d37/ddf1bedfe52e/launch-4eefcc91c90e.min.js
-SHEP_API=http://[fqdn]/api/v0.1
\ No newline at end of file
+SHEP_API=http://[fqdn]/api/v0.1
+LOGIN_BASE_URL=https://[fqdn]
```
EdwinGuzman, line 9: Add this to the `ENVIRONMENTVARS.md` file. I think the existing description from DFE is still applicable: https://github.com/NYPL/discovery-front-end/blob/development/ENVIRONMENTVARS.md?plain=1#L53

src/components/SubNav/SubNav.tsx
```diff
@@ -41,14 +47,19 @@ const SubNav = ({ activePage }: SubNavProps) => {
...
             active={activePage === "account"}
             aria-current={activePage === "account" ? "page" : undefined}
             hasWhiteFocusRing
           >
             My Account
           </RCLink>
         </li>
+        {isAuthenticated && (
+          <li>
+            <Link href={logOutRedirect}>Log out</Link>
```
EdwinGuzman, line 60: The `RCLink` internally uses `Link` so you can use that instead.

src/config/config.ts
```diff
@@ -47,6 +47,12 @@ export const appConfig = {
       "https://digital-research-books-beta.nypl.org/about?source=catalog",
     circulatingCatalog: "https://nypl.na2.iiivega.com/",
     legacyCatalog: "https://legacycatalog.nypl.org/",
+    loginUrl:
+      process.env.LOGIN_BASE_URL + "/login" ||
+      "https://login.nypl.org/auth/login",
```
EdwinGuzman, line 52: Hard-code the defaults to the QA links for login and logout urls.

src/server/auth.ts
```diff
@@ -53,3 +55,43 @@ export default async function initializePatronTokenAuth(req: NextRequest) {
...
+export function getLogoutRedirect(router) {
+  const current = router.pathname
+  let backPath = current
+  // If the patron is on any hold or account page, then
+  // redirect them to the home page after logging out.
+  if (current.includes("hold") || current.includes("account")) {
+    backPath = "/"
+  }
+  return `${appConfig.externalUrls.logoutUrl}?redirect_uri=http://local.nypl.org:8080/research/research-catalog${backPath}`
+  // TODO: The home url needs to be an env variable fs
```
EdwinGuzman, line 96: DFE has this as  `${window.location.origin}`  instead of an env var but this function is not being called on the client side to update the logout url. I'd be curious about testing this because I'm not sure if hardcoding the home domain url will work in various instances (the prod aws instance and the reverse proxy instance).

This _does_ work though. We can chat about this.

src/server/auth.ts
```diff
@@ -53,3 +55,43 @@ export default async function initializePatronTokenAuth(req: NextRequest) {
...
+    console.log(redirect)
+    return redirect
+  }
+}
+
+/**
+ * Creates redirect to log out user, then return user to their current page. Requires an instance of useRouter.
+ */
+
+export function getLogoutRedirect(router) {
```
EdwinGuzman, line 87: The comment "Requires an instance of useRouter." is great. You can also make this required in the code by setting the type to `NextRouter` as `(router: NextRouter)`.

EdwinGuzman: (And import the type from "next/router")

src/server/auth.ts
```diff
@@ -53,3 +55,43 @@ export default async function initializePatronTokenAuth(req: NextRequest) {
...
+/**
+ * Creates and returns redirect url from initializePatronTokenAuth() response
+ */
+export function getLoginRedirect(req) {
+  if (
+    !req.patronTokenResponse ||
+    !req.patronTokenResponse.isTokenValid ||
+    !req.patronTokenResponse.decodedPatron ||
+    !req.patronTokenResponse.decodedPatron.sub
+  ) {
```
EdwinGuzman, line 68: This works but it _shouldn't_ work. In DFE, we set the API patron response into the Express `req` object. We can't and shouldn't do that in the Next.js `req` object. In this case `req.patronTokenResponse` will always return an empty object since it doesn't exist.

This works because in the /pages/account/index.tsx file, it is only called when a patron is not authenticated. So if they're not authenticated and because `req.patronTokenResponse` is always empty, this function will always return the url. I think that's fine and how this function should work.

Instead of returning the `redirect` string or undefined, always return the redirect url, so you can remove the logic check. This is also under the assumption that this function should only be called when a user is authenticated (add that to the function comment block).

Another approach is to pass the `isAuthenticated` value to this function and then return the string only when it's false. But I think the way you set it up where you call this function when `isAuthenticated` is already false makes more sense.

src/server/auth.ts
EdwinGuzman: For now this works but update this to an env var before this goes to QA.

src/config/config.ts
EdwinGuzman: I take this back.

I want to say to try the pattern that's used for other env vars instead where its
```
loginUrl: {
  development: ...
  production: ..
}
```

but this is complicated by the fact that it uses an env var to generate the string. This probably needs a larger discussion with other devs about what the defaults should be.

src/server/auth.ts
```diff
@@ -53,3 +55,35 @@ export default async function initializePatronTokenAuth(req: NextRequest) {
...
+ */
+export function getLoginRedirect(req) {
+  const protocol = "http"
+  const hostname = req.headers["host"]
+  const originalUrl = BASE_URL + req.url
+  const fullUrl = encodeURIComponent(`${protocol}://${hostname}${originalUrl}`)
+  const redirect =
+    //`${appConfig.externalUrls.login}?redirect_uri=${fullUrl}`
+    `${process.env.LOGIN_BASE_URL}/login?redirect_uri=${fullUrl}`
+  console.log(redirect)
```
charmingduchess, line 71: rm console.log

7emansell:
Still writing unit tests + AC for more e2e testing for Chris, but done with code (I think). Also, environment is defaulting to prod even with APP_ENV set to development, so config.ts has 'dev-login' for both login/out links

7emansell:
**Doesn't work on Vercel as well

src/server/auth.ts
```diff
@@ -53,3 +54,43 @@ export default async function initializePatronTokenAuth(req: NextRequest) {
 
   return patronTokenResponse
 }
+
+/**
+ * Creates and returns redirect url from initializePatronTokenAuth() response. Call this function only with an un-authenticated patron,
+ * i.e., patronTokenResponse.isTokenValid must be false.
+ */
+export function getLoginRedirect(req) {
+  const protocol = "http"
```
EdwinGuzman, line 63: This is fine for testing locally but it should be https since we require the secure protocol on qa/prod. Is it available in the `req` object?

src/server/auth.ts
```diff
@@ -53,3 +54,43 @@ export default async function initializePatronTokenAuth(req: NextRequest) {
...
+  return redirect
+}
+
+/**
+ * Creates redirect to log out user, then return user to their current page.
+ */
+export const useLogoutRedirect = () => {
+  const baseUrl = "/research/research-catalog"
+  // Will send user back to prod if below doesn't work
+  const [redirect, setRedirect] = useState(`https://www.nypl.org${baseUrl}`)
```
EdwinGuzman, line 79: Use the `BASE_URL` constant, similar to how you used it in the previous function. And also on line 87.

src/server/auth.ts
```diff
@@ -53,3 +54,43 @@ export default async function initializePatronTokenAuth(req: NextRequest) {
...
+    // send them back to the page they were on.
+    if (current.includes("hold") || current.includes("account")) {
+      backPath = window.location.origin + baseUrl
+    }
+    setRedirect(
+      `${
+        appConfig.externalUrls.logoutUrl[appConfig.environment]
+      }?redirect_uri=${backPath}`
+    )
+  }, [])
```
EdwinGuzman, line 94: This works on all the pages I tested, but I'm also a bit confused how this does get updated even though it gets mounted once.

src/server/auth.ts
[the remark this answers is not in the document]
EdwinGuzman: Agreed. Looks like you will need to mock a lot of object and global variables.

EdwinGuzman (commented):
Some minor comments. I agree you needs some unit tests but you don't need to test the redirect functionality from the urls -- testing that the returned url is what we want is enough.

src/server/auth.ts
charmingduchess: why are we using window here instead of the next router object?

src/server/auth.ts
7emansell: `req.headers["x-forwarded-proto"]` or `req.protocol` are available but they're always undefined locally so I'm unsure. I think there's a broader issue with testing this because the environment is not consistent (setting APP_ENV doesn't seem to affect it)

7emansell: I assume `req.protocol` would be defined on prod and it would be `https`

src/server/auth.ts
7emansell: Window allowed more specificity and makes it so we don't have to reconstruct the url from host, protocol, and pathname (which are not consistently available across all environments) but this does work with a router locally (see this commit https://github.com/NYPL/research-catalog/commit/e202e3b483ca2f2b1c42b9875f9fefa546e8776f)

src/server/auth.ts
```diff
@@ -53,3 +54,42 @@ export default async function initializePatronTokenAuth(req: NextRequest) {
 
   return patronTokenResponse
 }
+
+/**
+ * Creates and returns redirect urk. Call this function only with an un-authenticated patron,
```
EdwinGuzman, line 59: minor typo "urk"

src/server/auth.ts
```diff
@@ -53,3 +54,42 @@ export default async function initializePatronTokenAuth(req: NextRequest) {
...
+    appConfig.externalUrls.loginUrl[appConfig.environment]
+  }?redirect_uri=${fullUrl}`
+  return redirect
+}
+
+/**
+ * Creates redirect to log out user, then return user to their current page.
+ */
+export const useLogoutRedirect = () => {
+  // Will send user back to prod if below doesn't work
```
EdwinGuzman, line 77: Say that typically the below, the useEffect, won't work for the noscript users (or the users who have javascript turned off).

EdwinGuzman (approved):
Update looks good, just very minor comments.

src/server/auth.ts
dgcohen: I'm surprised to learn about the inconsistency across environments with router.pathname. We're using that elsewhere so it might be good to reconsider its use in those places.

dgcohen (approved):
This looks good! I'd like us to move isAuthenticated to a context or hook to avoid prop drilling (not sure if you already considered it) but this looks good for now.

src/server/auth.ts
charmingduchess: what environments are they not available in?

charmingduchess:
What do you think about extracting the auth call and setting prop.isAuthenticated to a method that takes `req` as an argument? It might be a pain to update every single page if our auth changes.

src/config/config.ts
```diff
@@ -47,6 +47,14 @@ export const appConfig = {
       "https://digital-research-books-beta.nypl.org/about?source=catalog",
     circulatingCatalog: "https://nypl.na2.iiivega.com/",
     legacyCatalog: "https://legacycatalog.nypl.org/",
+    loginUrl: {
+      development: "https://dev-login.nypl.org/auth/login",
+      production: "https://dev-login.nypl.org/auth/login",
```
charmingduchess, line 52: let's put the actual production login url here?

src/server/auth.ts
charmingduchess: so do we want protocol to be req.protocol if it _is_ defined?

src/server/auth.ts
charmingduchess: If it's the development environment where protocol is missing, it shouldn't be a problem for useRouter for most of the app. I think we rarely use it to determine the protocol.

src/config/config.ts
EdwinGuzman: Yes, this was temporary because we didn't have the env vars down correctly and is waiting PR #99 to be merged in.

7emansell: Updated for clarity but ^

PR #98, opened 2024-02-12
SCC-3955: Account data fetching

7emansell:
Resolves ticket [SCC-3955](https://jira.nypl.org/browse/SCC-3955).
Creates Sierra client and returns data for My Account.


src/models/MyAccount.ts
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  buildPatron(patron: SierraPatron): Patron {
+    return {
+      name: patron.names[0],
+      barcode: patron.barcodes[0],
+      expirationDate: patron.expirationDate,
+      primaryEmail: patron.emails[0],
+      emails: patron.emails,
+      primaryPhone: patron.phones[0].number,
```
EdwinGuzman, line 166: If `patron.phones` is an empty array, this will fail.

EdwinGuzman: Even if a phone number is required by the ILS for a patron account, let's assume it's not to be on the safer side.

src/models/MyAccount.ts
```diff
@@ -0,0 +1,213 @@
...
+        patron: MyAccount.getRecordId(checkout.patron),
+        title: bibDataMap[checkout.item.bibIds[0]].title,
+        isResearch: bibDataMap[checkout.item.bibIds[0]].isResearch,
+        bibId: checkout.item.bibIds[0],
+        isNyplOwned: bibDataMap[checkout.item.bibIds[0]].isNyplOwned,
+      }
+    })
+  }
+
+  static async fetchCheckouts(baseQuery: string) {
```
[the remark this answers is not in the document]
EdwinGuzman: I better understand this patter now. The two `this`s are confusing in the factory function and the constructor but it makes sense so you can ignore this comment.

It would be nice to somehow have the patron id as an instance variable so that here you could do
```
client.get(`${this.patronId}${holdsQuery}`)
```

and not have to pass `baseQuery`, but this function gets called before the instance is created. That also confused me a bit.

EdwinGuzman (changes requested):
I left minor comments.

As a general question, `sinon` is introduced but could jest be used instead? I think the tests could be rewritten to use jest mocks so that we don't introduce another npm package.

EdwinGuzman:
Also, I did not review the code that is similar to #96

src/models/MyAccount.ts
```diff
@@ -0,0 +1,213 @@
...
+            detail: entry.chargeType.display,
+            amount: entry.itemCharge,
+            date: entry.assessedDate,
+          }
+        }
+      }),
+    }
+  }
+
+  patronCookieMatchesCheckoutOrHold(cookieId: string) {
```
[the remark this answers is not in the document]
7emansell: Will be moved into api route handler

src/models/MyAccount.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  buildHolds(holds: SierraHold[], bibDataMap): Hold[] {
+    return holds.map((hold: SierraHold) => {
+      return {
+        patron: MyAccount.getRecordId(hold.patron),
+        id: MyAccount.getRecordId(hold.id),
+        pickupByDate: hold.pickupByDate,
+        canFreeze: hold.canFreeze,
+        frozen: hold.frozen,
+        status: MyAccount.getStatus(hold.status),
```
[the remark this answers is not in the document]
7emansell: @charmingduchess i think you might have said statii are not unique to holds, I can't remember– but if so I think we stick with this name

src/models/modelTests/MyAccount.test.ts
```diff
@@ -0,0 +1,183 @@
...
+        holds: holds.entries,
+        patron,
+        checkouts: checkouts.entries,
+        fines: { total: 0, entries: [] },
+      })
+      expect(account.patron).toStrictEqual({
+        name: "KAHN, VERA RUTH",
+        barcode: "23333121538324",
+        expirationDate: "2025-03-28",
+        primaryEmail: "veraruthkahn@gmail.com",
```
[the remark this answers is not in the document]
charmingduchess: good point!

src/server/sierraClient/index.ts
```diff
@@ -0,0 +1,58 @@
+import wrapper from "@nypl/sierra-wrapper"
+import aws from "aws-sdk"
+
+interface KMSCache {
+  client: any
+  key: string
+  secret: string
```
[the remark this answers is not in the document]
charmingduchess: updated to cache and use decrypted values.

src/models/MyAccount.ts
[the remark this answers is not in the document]
charmingduchess: going to update to getholdstatus. there is status on checkout, but this method is very specific to holds.

charmingduchess:
@EdwinGuzman @dgcohen 
Sorry for the delay in responding to these comments! I think I incorporated everything here. 

The refactor I mentioned at stand up for the nyplapiclient is included in this work. It does not break the tests, but we could also refactor the tests as well to be less reliant on promises and callbacks.

src/server/nyplApiClient/index.ts
```diff
@@ -46,28 +22,33 @@ const nyplApiClient = async (options = { apiName: "platform" }) => {
...
-      })
-      .catch((error) => {
-        // TODO: add this once the Winston config is added.
-        // logger.error('ERROR trying to decrypt using KMS.', error);
-        reject(error)
-      })
-  })
+  let decryptedId: string
+  let decryptedSecret: string
+  if (CACHE.secret && CACHE.secret) {
```
EdwinGuzman, line 28: `.secret` is checked twice but one should be `id`.

src/server/nyplApiClient/index.ts
```diff
@@ -46,28 +22,33 @@ const nyplApiClient = async (options = { apiName: "platform" }) => {
...
+    const nyplApiClient = new NyplApiClient({
+      base_url: baseUrl,
+      oauth_key: decryptedId,
+      oauth_secret: decryptedSecret,
+      oauth_url: appConfig.urls.tokenUrl,
+    })
+    CACHE.clients[apiName] = nyplApiClient
+    return nyplApiClient
+  } catch (error) {
+    console.error(error.message)
```
EdwinGuzman, line 50: Not necessarily needed for this PR but this should throw an error. If something goes wrong here, calling `client.get("...")` will crash

src/server/sierraClient/index.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+  try {
+    await wrapper.config({
+      key: decryptedKey,
+      secret: decryptedSecret,
+      base: base,
+    })
+
+    CACHE.client = wrapper
+    return wrapper
```
EdwinGuzman, line 44: Similar comment to `nyplApiClient`. This returns the client itself but line 21 returns the resolved promise.

EdwinGuzman (commented):
Some comments about the returned values of the clients. There's already work going on based on this branch that I don't think it's worth keeping this open and instead fixing/updating the comments in those future PRs. Thoughts? 

Otherwise, this looks good.

EdwinGuzman (approved):
Looks good. I made a comment in a commit but should've looked at the final code, so ignore that.

pages/account/index.tsx
```diff
@@ -1,18 +1,42 @@
...
+
+export default function MyAccount({
+  checkouts,
+  holds,
+  patron,
+  fines,
+  isAuthenticated,
+}: MyAccountProps) {
+  const sierraClientErrorThrown = !patron.name
+  console.log(checkouts, holds, patron, fines)
```
dgcohen, line 26: This console.log can be removed

jest.config.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
     "!<rootDir>/coverage/**",
   ],
   moduleNameMapper: {
+    sinon: "<rootDir>/node_modules/sinon/pkg/sinon.js",
```
dgcohen, line 15: Is this a holdover from testing with sinon?

pages/account/index.tsx
```diff
@@ -1,18 +1,42 @@
...
+}
+
+export default function MyAccount({
+  checkouts,
+  holds,
+  patron,
+  fines,
+  isAuthenticated,
+}: MyAccountProps) {
+  const sierraClientErrorThrown = !patron.name
```
dgcohen, line 25: Is this the only case where an error message is shown?

pages/account/index.tsx
```diff
@@ -32,7 +56,16 @@ export async function getServerSideProps({ req }) {
     }
   }
 
-  return {
-    props: { isAuthenticated },
+  const id = patronTokenResponse.decodedPatron.sub
```
dgcohen, line 59: Looks like there's a case where decoded patron can be null, which me think optional chaining should be used here. In addition, we should check for the existence of id before calling MyAccountFactory.

pages/account/index.tsx
```diff
@@ -32,7 +56,16 @@ export async function getServerSideProps({ req }) {
...
+  let props
+  try {
+    const { checkouts, holds, patron, fines } = await MyAccountFactory(id)
+    props = { checkouts, holds, patron, fines, isAuthenticated }
```
dgcohen, line 63: This can be returned directly without assigning to an intermediate 'props' var.

pages/account/index.tsx
```diff
@@ -32,7 +56,16 @@ export async function getServerSideProps({ req }) {
...
+  } catch (e) {
+    props = {
```
dgcohen, line 65: Same here

dgcohen (changes requested):
Added some comments but otherwise looks good!

pages/account/index.tsx
EdwinGuzman: True. I suggested this approach because then this function only has one return and it's at the end. It's a preference thing but whatever works best.

pages/account/index.tsx
charmingduchess: It's there so we don't get a linting error for unused vars! also still being used generally for debugging.

pages/account/index.tsx
charmingduchess: I'm going to revert it. We have a return statement higher up anyways, and I personally think it feels more explicit to have the return statement split by try/catch

PR #99, opened 2024-02-13
SCC-3998 - App config cleanup

dgcohen:
This PR makes a few changes to the way appConfig is being used.

Rather than accessing all build time environment variables directly in appConfig, we are setting the correct ENV value in the docker build and using that to access the default values in appConfig.

I also removed the NEXT_PUBLIC prefix from closed locations since we not be able to use these as runtime variables as expected. These will be passed to the components as props from getServerSideProps as part of a separate PR to allow for runtime configuration.


src/config/config.ts
```diff
@@ -1,74 +1,58 @@
 import { parseLocations } from "../utils/itemUtils"
 
 export const appConfig = {
-  environment: process.env.APP_ENV || "production",
+  environment: process.env.ENV || "production",
```
EdwinGuzman, line 4: I think `APP_ENV` is less confusing than `ENV` but not a big deal.

src/config/config.ts
```diff
@@ -1,74 +1,58 @@
...
     researchMaterialsHelp:
       "https://www.nypl.org/help/request-research-materials",
   },
   tokenUrl: "https://isso.nypl.org/",
-  closedLocations: parseLocations(process.env.NEXT_PUBLIC_CLOSED_LOCATIONS),
-  recapClosedLocations: parseLocations(
-    process.env.NEXT_PUBLIC_RECAP_CLOSED_LOCATIONS
-  ),
+  closedLocations: parseLocations(process.env.CLOSED_LOCATIONS),
+  recapClosedLocations: parseLocations(process.env.RECAP_CLOSED_LOCATIONS),
```
EdwinGuzman, line 45: This still won't work if either of these are updated through Task Definitions. This is okay to keep for now but the strategy will have to be updated in the future.

src/config/config.ts
charmingduchess, line 1: do we not have access to process.env on the client side?

src/config/config.ts
dgcohen: I agree. Can change it to APP_ENV, will just need to update the env variable name in vercel and aws.

src/config/config.ts
EdwinGuzman: It's tricky. we do if it is set in `next.config.js` but it also needs to be declared at build time in Travis.

EdwinGuzman (approved):

src/config/config.ts
dgcohen: Cool, going to keep it for now and open a ticket to sort out runtime vars.

7emansell (approved):
Lgtm

src/config/config.ts
```diff
@@ -1,76 +1,70 @@
...
     discovery: {
-      development:
-        process.env.DISCOVERY_API_BASE_URL ||
-        process.env.PLATFORM_API_BASE_URL ||
-        "https://qa-platform.nypl.org/api/v0.1",
-      production:
-        process.env.DISCOVERY_API_BASE_URL ||
-        process.env.PLATFORM_API_BASE_URL ||
-        "https://platform.nypl.org/api/v0.1",
+      development: "https://platform.nypl.org/api/v0.1",
```
charmingduchess, line 16: i think the platform url needs to match the auth url. otherwise we're going to have inconsistencies with running hold request and checkout flows. there are a bunch of records that are not the same or not present in both qa-platform and prod platform (likewise sierra). i'm having trouble putting my finger exactly on what would happen, but i don't think its a great idea. hold request renewals and cancellations will be happening in sierra test, but hold requests from RC, which happen via platform ASAIK, will be in prod. the patron will be populated with data from test, and i can see a bunch of weirdness going down.

tl;dr: beware mixing prod data with test auth
@nonword can you confirm?

nonword: Agreed. I think you prob don't want to wire up local `development` environment to use `production` endpoints anywhere. The `discovery` base URL will control where holds are placed. And will likely also be where patron records are pulled from. When patron-login is folded in, the login/logout endpoints will need to agree with where we pull patron records from. So is there a reason to wire up local development to use production endpoints for any of these resources?

dgcohen: I configured it that way to keep the settings consistent with the local config defaults before i made these changes. Makes sense the have development point to the QA endpoints, will make that change now.

charmingduchess (approved):

EdwinGuzman (approved):

PR #100, opened 2024-02-14
SCC-4002: Adding meta tags on the main _app file and dynamically on all pages

EdwinGuzman:
Resolve JIRA Ticket [SCC-4002](https://jira.nypl.org/browse/SCC-4002)

- Adds all main tags in `_app.tsx`. The dynamic tags that will be updated on every page have a `key` attribute.
- Those are updated on every page file. Next.js handles updating the value of meta tag on every page.
- These are the same values as what is currently on prod.


pages/404/redirect.tsx
```diff
@@ -2,15 +2,23 @@ import Head from "next/head"
...
     <>
       <Head>
-        <title>404 Not Found</title>
+        <meta property="og:title" content={metadataTitle} key="og-title" />
+        <meta
+          property="og:site_name"
+          content={metadataTitle}
+          key="og-site-name"
+        />
+        <meta name="twitter:title" content={metadataTitle} key="tw-title" />
```
charmingduchess, line 20: what does this twitter title tag do?

dgcohen (approved):
This looks good! Not sure if you already considered but Next allows for exporting metadata as a config object from the component files which cut down on the html and some of the repetition in the components, but we can consider doing this as a follow up optimization if we want:

https://nextjs.org/docs/app/building-your-application/optimizing/metadata

EdwinGuzman:
@dgcohen yes, I like that approach better but  I think that's only for the app router and not page router.

charmingduchess (approved):

PR #103, opened 2024-02-16
SCC-3951: Checkout renewal API route

7emansell:
Resolves ticket [SCC-3951](https://jira.nypl.org/browse/SCC-3951).

Adds API route handler for My Account (so far, only handles this one route) and internal route that triggers a POST request to Sierra to renew a checkout. 
Route error handling covers success (200), failure (500), and renewal not allowed (403). This last case could be because it's too soon to renew OR because it's not a renewable item (or many other more niche possibilities), but these will be handled on the Renew button component. 

Also adds some adjustments to the My Account model that only came up with this new test patron.


pages/api/account/account.test.ts
```diff
@@ -0,0 +1,77 @@
...
+jest.mock("../../../src/server/sierraClient")
+
+describe("checkoutRenewal", () => {
+  let mockStatus: any
+  let mockJson: any
+  let mockRes: NextApiResponse
+
+  beforeEach(() => {
+    jest.clearAllMocks()
+    ;(mockStatus = jest.fn().mockReturnThis()), (mockJson = jest.fn())
```
charmingduchess, line 14: do you need this parentheses syntax here? can it just be
```
mockStatus = jest.fn...
mockJson = jest.fn()
```
on two lines?

pages/api/account/[...id].ts
```diff
@@ -0,0 +1,47 @@
...
+  checkoutId: string
+) {
+  try {
+    const client = await sierraClient()
+    await client.post(`patrons/checkouts/${checkoutId}/renewal`)
+    res.status(200).json({ message: "Renewed!" })
+  } catch (error) {
+    if (error.response.status === 403) {
+      res.status(403).json({
+        message:
```
Comment (charmingduchess) [inline pages/api/account/[...id].ts:39]: lets pass the message directly from the response here instead of generating the message ourselves.

pages/api/account/[...id].ts
```diff
@@ -0,0 +1,47 @@
...
+
+export default async function handler(req: NextRequest, res: NextApiResponse) {
+  const patronTokenResponse = await initializePatronTokenAuth(req)
+  const patronId = patronTokenResponse.decodedPatron?.sub
+  if (!patronId) {
+    res.status(403).json({
+      message: "No authenticated patron",
+    })
+  }
+  const checkoutRenewalMatch = req.url.match(/\/checkouts\/(\d+)\/renewal$/)
```
Comment (EdwinGuzman) [inline pages/api/account/[...id].ts:21]: I think the comparison you want to make, according to the ticket, is `patronId` against the id from the `req.url`.

Some thoughts:
- should the file name have the catch all for the id? I think `/[id].ts` is okay and you should be able to do `req.query.id` for the variable `checkoutId`.
- Then you can compare `patronId` against the value from `checkoutId` to make sure we have the same patron.

Another approach can be to make this handler a POST request and instead of doing `/api/account/id`, you would pass the id as part of the body object. In that case you'd have a larger if statement for `if (request.method === "POST") {....` and `checkoutId` would be `req.body.id`.

Comment (EdwinGuzman) [reply pages/api/account/[...id].ts:21]: The POST approach might be more "secure" since we are not putting the patron's id in the url request. I vaguely remember talking about this but forgot where we landed @charmingduchess .

pages/api/account/[...id].ts
```diff
@@ -0,0 +1,47 @@
...
+  if (checkoutRenewalMatch) {
+    const checkoutId = checkoutRenewalMatch[1]
+    await checkoutRenewal(res, checkoutId)
+  }
+}
+
+export async function checkoutRenewal(
```
Comment (EdwinGuzman) [inline pages/api/account/[...id].ts:28]: It is great that this is being exported because we might be able to reuse this function elsewhere, such as in `getServerSideProps` if we need to (or server components in the far future). However, what we want from this function is the information regarding a successful renewal and not necessarily the response object.

Right now, this function is coupled with the nextjs `res` object. If it returns a boolean or error regarding whether it was successful or not, it can be more portable and you can remove the `res` argument. It would mean that the `res.status(....` calls live in the handler function but doing so gives this function only the responsibility for the API check and returning it.

EdwinGuzman (commented):
I did not review the tests yet in case it changes because of the feedback.

pages/api/account/[...id].ts
Comment (7emansell) [reply pages/api/account/[...id].ts:21]: So `patronId` is not the same thing as the `checkoutId` from the url but (after some offline discussion previously) I think it's sufficient to just check that the user is logged in before they make the request– the case that the user could be logged in and somehow viewing and renewing someone else's checkout doesn't seem possible. So the handler is really checking what you're trying to do (which will soon need to be distinguished from https://jira.nypl.org/browse/SCC-3954)

pages/api/account/checkouts/renewal/[id].ts
```diff
@@ -0,0 +1,53 @@
...
+import initializePatronTokenAuth from "../../../../../src/server/auth"
+
+/**
+ * API route handler for /api/account/checkouts/renewal/{checkoutId}
+ */
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  let responseMessage
```
Comment (charmingduchess) [inline pages/api/account/checkouts/renewal/[id].ts:12]: I'm wondering how crucial it is to have these different response messages for the different cases.  I don't see a specific use case for the other messages. They don't seem like something we'd want to display to a patron, and we haven't planned for any detailed error logging. If they are all 403 errors, maybe they can all just have the same response message - "RENEWAL NOT ALLOWED. Please contact gethelp@nypl.org for assistance."
In that case, we could just initialize `responseMessage` and `responseStatus` to 403 and that default message, and we could eliminate some repetitive lines of code.

pages/account/index.tsx
```diff
@@ -20,21 +20,52 @@ export default function MyAccount({
   patron,
   fines,
 }: MyAccountProps) {
-  console.log(checkouts, holds, patron, fines)
+  console.log("patron's checkouts", checkouts)
```
dgcohen, line 23: This console.log can be removed

src/models/MyAccount.ts
```diff
@@ -168,8 +173,12 @@ export default class MyAccount {
       expirationDate: patron.expirationDate,
       primaryEmail: patron.emails.length > 0 ? patron.emails[0] : "",
       emails: patron.emails,
-      primaryPhone: patron.phones.length > 0 ? patron.phones[0].number : "",
-      phones: patron.phones,
+      primaryPhone: patron.phones
```
dgcohen, line 176: I feel like this syntax can be simplified a bit to avoid the double ternary, maybe something like:

patron?.phones[0]?.number || ""

or

patron?.phones?.length ? patron.phones[0].number || ""

dgcohen (changes requested):
Looks good, just added a few small comments

pages/api/account/checkouts/renewal/[id].ts
```diff
@@ -0,0 +1,63 @@
...
+    const checkoutPatronId = JSON.parse(JSON.stringify(req.body))
+    /**  We check that the patron cookie matches the patron id in the request body,
+     * i.e.,the logged in user is the owner of the checkout. */
+    if (checkoutPatronId == cookiePatronId) {
+      const response = await checkoutRenewal(checkoutId)
+      responseStatus = response.status
+      responseMessage = response.message
+      // If successful, returns the renewed checkout object in the body.
+      if (response.body) {
+        responseBody = response.body.dueDate
```
Comment (EdwinGuzman) [inline pages/api/account/checkouts/renewal/[id].ts:37]: Yes, this is what I was thinking but you can be clearer  and say `response.dueDate = response.body.dueDate`. On line 14, `responseBody` is also set to an object and here (and from the test), the `dueDate` is a string.

When the client gets the response back in 
```
// New due date.
alert(responseData.body)
```

it'll be easier for anyone to understand that `responseData.body.dueDate` is the due date. Otherwise they'll see a response object like this and be confused
```
const responseData = await response.json()
console.log(responseData)
{
  status: 200,
  body: "2024-03-12T08:00:00Z"
}
```

pages/api/account/checkouts/renewal/[id].ts
```diff
@@ -0,0 +1,63 @@
...
+      }
+    } else {
+      responseStatus = 403
+      responseMessage = "Authenticated patron does not own this checkout"
+    }
+  }
+  return res.status(responseStatus).json({
```
Comment (EdwinGuzman) [inline pages/api/account/checkouts/renewal/[id].ts:44]: Let's talk about send back a better message for a GET vs POST request from the client in future PRs. There some cleanup we can do but it's not urgent for this PR.

EdwinGuzman (changes requested):
Only comment is to make the response object clearer about the due date property.

7emansell:
Merged in but weird

PR #105, opened 2024-02-20
SCC-3967 - Move data fetchers to /src/server and add tests

dgcohen:
This PR moves the data fetcher helpers that were previously in the pages directory to the src/server directory. This will ensure that the files in pages/api will only contain route handlers that invoke these functions.

I've also added elementary tests for the data fetchers including mocks. I've also slightly refactored the existing drb tests to use mockResolvedValueOnce.


EdwinGuzman (approved):
Two minor comments.

src/server/api/tests/bib.test.ts
charmingduchess, line 1: @7emansell came up with [this pattern for testing](https://github.com/NYPL/research-catalog/blob/1c3656f3643cc694ea32f5207b82f5c7af5ac6c4/pages/api/account/account.test.ts) that might be worth checking out. I think it would be good to use moving forward because it allows you to set the mocked return value in the test itself. 
```
import sierraClient from "../../../src/server/sierraClient"

jest.mock("../../../src/server/sierraClient")

describe("checkoutRenewal", () => {
  it("should return a success message if renewal is successful", async () => {
    const checkoutId = "123"
    const clientMock = jest
      .fn()
      .mockResolvedValueOnce({ status: 200, data: { message: "Renewed" } })
    ;(sierraClient as jest.Mock).mockResolvedValueOnce({ post: clientMock })

    const response = await checkoutRenewal(checkoutId)

    expect(sierraClient).toHaveBeenCalled()
    expect(clientMock).toHaveBeenCalledWith(
      `patrons/checkouts/${checkoutId}/renewal`
    )
    expect(response.status).toBe(200)
    expect(response.message).toBe("Renewed")
  })
```

src/server/api/tests/search.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        resolve({
+          get: jest
+            .fn()
+            .mockResolvedValueOnce({
+              itemListElement: [{}, {}, {}, {}],
+              totalResults: 4,
+            })
+            .mockResolvedValueOnce({
+              itemListElement: [{}, {}, {}, {}, {}, {}, {}, {}, {}, {}],
+              totalResults: 4,
```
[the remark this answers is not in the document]
dgcohen: You would think but this matches what I got from a real-world example. I think totalResults might match the search results and not the number of aggregations?

src/server/api/tests/bib.test.ts
dgcohen: Thanks! Agree about landing on a consistent solution sooner rather than later. Gonna try this out.

src/server/api/tests/search.test.ts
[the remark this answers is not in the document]
charmingduchess: item list element is the array of aggregations, and total results refers to the number of bibs that are being described.

src/server/api/tests/drb.test.ts
```diff
@@ -0,0 +1,95 @@
...
+import type { DRBResults } from "../../../types/drbTypes"
+
+jest.mock("../../nyplApiClient", () => {
+  return jest
+    .fn()
+    .mockImplementationOnce(async () => {
+      return await new Promise((resolve) => {
+        resolve({
+          get: jest.fn().mockReturnValueOnce({
+            data: {
```
charmingduchess, line 11: can this object be imported from a separate file?

src/server/api/tests/bib.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  it("should respond with a 404 status when uri is not present and sierra call responds with a non-200 status", async () => {
+    const bibResponse = (await fetchBib("b17418167")) as BibResponse
+
+    expect(bibResponse.status).toEqual(404)
+  })
+
+  it("should redirect to standardized ID page when id is non-standard", async () => {
+    const bibResponse = (await fetchBib("B17418167")) as BibResponse
+    console.log(bibResponse)
```
charmingduchess, line 117: rm console.log

charmingduchess (changes requested):
I left a couple comments to move test data and rm console.log

7emansell (approved):
Lgtm aside from console logs!

src/server/api/tests/drb.test.ts
dgcohen: Jest mocks don't seem to allow for it unfortunately.

charmingduchess (approved):

PR #109, opened 2024-02-21
SCC-3954: Settings API routes

7emansell:
Resolves ticket [SCC-3954](https://jira.nypl.org/browse/SCC-3954).

Adds API route handlers for 2 internal routes: POST settings update and POST update pin. 
Also adds buttons to my account page for testing.


pages/account/index.tsx
```diff
@@ -47,6 +47,58 @@ export default function MyAccount({
     }
   }
 
+  /** Testing settings api route */
+  async function settingsUpdate(patronId) {
+    try {
+      const response = await fetch(
+        `/research/research-catalog/api/account/settings/${patronId}`,
```
EdwinGuzman, line 54: Curious: Does this work if you remove the base `/research/research-catalog`?

pages/api/account/checkouts/renewal/checkouts.test.ts
```diff
@@ -123,24 +120,21 @@ describe("checkoutRenewal", () => {
...
-      expect(sierraClient).toHaveBeenCalled()
-      expect(clientMock).toHaveBeenCalledWith(
-        `patrons/checkouts/${checkoutId}/renewal`
-      )
-      expect(error.response.status).toBe(403)
-      expect(error.response.data.message).toBe(
-        "RENEWAL NOT ALLOWED. Please contact gethelp@nypl.org for assistance"
-      )
-    }
+    const response = await checkoutRenewal(checkoutId)
```
EdwinGuzman, line 130: Removing the try/catch and testing just what the function returns seems much nicer.

pages/api/account/update-pin/[id].ts
```diff
@@ -0,0 +1,61 @@
...
+
+export async function pinUpdate(
+  patronId: string,
+  barcode: string,
+  oldPin: string,
+  newPin: string
+) {
+  try {
+    const client = await sierraClient()
+    await client.post("patrons/validate", {
```
Comment (EdwinGuzman) [inline pages/api/account/update-pin/[id].ts:49]: Is the assumption that if this passes then the old pin is valid? So no response is expected that you need to check or verify?

pages/api/account/update-pin/[id].ts
```diff
@@ -0,0 +1,61 @@
...
+      barcode: barcode,
+      pin: oldPin,
+    })
+    await client.put(`patrons/${patronId}`, { pin: newPin })
```
Comment (EdwinGuzman) [inline pages/api/account/update-pin/[id].ts:53]: This is the similar line that `settingsUpdate` calls. I wonder if it's possible to somehow call `settingsUpdate` here but that might be more complicated and not worth doing.

Just pointing this out because now there are two places where the same Sierra endpoint is called. And really, this internal API route only want to verify the old pin before updating the patron's setting. Something we can discuss at some point with Vera but this work so let's proceed.

pages/api/account/settings/[id].ts
```diff
@@ -0,0 +1,50 @@
...
+    if (patronId == cookiePatronId) {
+      const response = await settingsUpdate(patronId, patronData)
+      responseStatus = response.status
+      responseMessage = response.message
+    } else {
+      responseStatus = 403
+      responseMessage = "Authenticated patron does not match request"
+    }
+  }
+  res.status(responseStatus).json(responseMessage)
```
Comment (EdwinGuzman) [inline pages/api/account/settings/[id].ts:36]: `responseMessage` and `responseStatus` should have default values. Try this:
- log in so your `cookiePatronId` is valid
- hit the API endpoint in your browser
- this gets returned:
Screen Shot 2024-02-23 at 3 20 25 PM

We _don't_ expect this endpoint to be hit through a GET request but it would be nice, to have a complete API, to return a better error message. This can be either a default status/message or you can check if the method is a GET request and return a "please make a post request....." message.

EdwinGuzman (changes requested):
This looks good! Just comments about the naming for the pin values and the default returned values from the APIS when a GET request is made.

Also, and not exactly relevant to this PR, but after logging in the account page breaks:
Screen Shot 2024-02-23 at 3 17 23 PM

pages/api/account/update-pin/[id].ts
Comment (7emansell) [reply pages/api/account/update-pin/[id].ts:49]: Yes based on this https://sandbox.iii.com/iii/sierra-api/swagger/index.html#!/patrons/Validate_patron_by_barcode_and_PIN_post_15, and experimentally– it'll catch as expected when the pin or barcode is incorrect or missing

pages/api/account/update-pin/[id].ts
Comment (7emansell) [reply pages/api/account/update-pin/[id].ts:53]: Right. It does seem like we could do something along the lines of, if the pin is part of the request body, then do this validation, otherwise just update the settings (all in `settingsUpdate`). I think they were separated because they're separated in the UI ("Save changes" for settings vs. "Change PIN/password" buttons)? There's already quite a lot of repetition with these handlers though so I don't know if it's worth it to just refactor this one part

pages/account/index.tsx
7emansell: It does not .

EdwinGuzman: Update this to use the `BASE_URL` constant.

7emansell:
> Also, and not exactly relevant to this PR, but after logging in the account page breaks
@EdwinGuzman looks like this is fixed on #98 (after this branch was cut, so will just be careful on that part of the merge)

EdwinGuzman (approved):
Just wait on more reviews.

pages/api/account/update-pin/[id].ts
Comment (charmingduchess) [reply pages/api/account/update-pin/[id].ts:54]: for this pr let's just combine these two endpoints into one handler. which i believe means they have to share a url, which is fine.
we can investigate combining everything later.

charmingduchess (changes requested):
Looks alright to me! Just an ask for a variable name change in the tests.

charmingduchess:
> This looks good! Just comments about the naming for the pin values and the default returned values from the APIS when a GET request is made.
> 
> Also, and not exactly relevant to this PR, but after logging in the account page breaks: 

I addressed this bug in my account data fetching PR.

charmingduchess (approved):
NB there are testing problems here that are being addressed in SCC-3953/holds-api. This PR will be merged into main after the other PR is merged into this one.

PR #111, opened 2024-02-26
SCC-3953: Holds API routes

7emansell:
Resolves ticket [SCC-3953](https://jira.nypl.org/browse/SCC-3953).

Adds API route handlers for 2 internal routes: POST hold update and POST cancel hold, and adds respective buttons to my account page for testing.

Also refactors tests for all account api routes so far.


pages/api/account/holds/cancel/[id].ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+  return res.status(responseStatus).json(responseMessage)
+}
+
+export async function holdCancel(holdId: string) {
+  try {
+    const client = await sierraClient()
+    await client.deleteRequest(`patrons/holds/${holdId}`)
+    return { status: 200, message: "Deleted" }
```
Comment (EdwinGuzman) [inline pages/api/account/holds/cancel/[id].ts:47]: Perhaps "Cancelled" is better than "Deleted"?

pages/api/account/holds/update/updateHold.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      "Authenticated patron does not own this hold"
+    )
+  })
+
+  it("should call holdUpdate if authentication succeeds", async () => {
+    ;(initializePatronTokenAuth as jest.Mock).mockResolvedValueOnce({
+      decodedPatron: { sub: "123456" },
+    })
+    await handler(req as NextApiRequest, res as NextApiResponse)
+    expect(holdUpdate).toHaveBeenCalled
```
EdwinGuzman, line 55: You need to invoke the function `toHaveBeenCalled()`.

pages/api/account/holds/cancel/cancelHold.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+  })
+
+  it("should return 403 if no authenticated patron", async () => {
+    req.body.patronId = "123456"
+    ;(initializePatronTokenAuth as jest.Mock).mockResolvedValueOnce({
+      decodedPatron: null,
+    })
+    await handler(req as NextApiRequest, res as NextApiResponse)
+    expect(holdCancel).not.toHaveBeenCalled
```
EdwinGuzman, line 32: Invoke the function.

EdwinGuzman (commented):
Looks good but there are a few places where the `toHaveBeenCalled` function should be invoked.

pages/account/index.tsx
```diff
@@ -94,6 +94,59 @@ export default function MyAccount({
...
         alert(`error: ${responseData}`)
       }
+    } catch (error) {
+      console.log(error)
+      alert("fetching error")
+    }
+  }
+
+  /** Testing hold update api route */
+  async function holdUpdate(patronId, holdId, freeze, pickupLocation) {
```
charmingduchess, line 104: since holdId, freeze, and pickupLocation are all being passed to sierra, can they be combined into one object here? My thinking was to have this method pass along the update request body without having to know what's necessarily in it.

charmingduchess: oh i see this is just the test button!

7emansell:
> Talked about this offline: move all the internal functions into one helpers function, similar to the tests.

@EdwinGuzman Forgot that only one default export can be mocked like this so I think we have to stick with it as is so that the handlers can continue mocking the entire file they're importing from. I do not understand why Jest does this tbh

charmingduchess:
> > Talked about this offline: move all the internal functions into one helpers function, similar to the tests.
> 
> @EdwinGuzman Forgot that only one default export can be mocked like this so I think we have to stick with it as is so that the handlers can continue mocking the entire file they're importing from. I do not understand why Jest does this tbh

Do they have to be default exports? I think the issue before was that the updateHold method was already defined in the handler upon import, not a default/named exports issue. I imagine we'd have a similar problem if they were both named exports too.  As long as they don't call each other, the api "helper methods" could all be in the same file. But they don't have to be!

EdwinGuzman (approved):
The new Holds route looks good but also the entire cleanup is very nice. Clever way to import and mock the necessary functions in the handler functions.

EdwinGuzman:
Wait for @dgcohen 's review but also, fix the conflicts.

pages/account/index.tsx
```diff
@@ -24,13 +24,13 @@ export default function MyAccount({
...
       const response = await fetch(
-        `/research/research-catalog/api/account/checkouts/renewal/${checkoutId}`,
+        `/research/research-catalog/api/account/checkouts/renew/${checkoutId}`,
         {
           method: "POST",
           headers: {
             "Content-Type": "application/json",
           },
-          body: JSON.stringify(patronId),
+          body: JSON.stringify({ patronId: patronId }),
```
dgcohen, line 33: Small nit but this and all the other same-name assignments can be shortened to:

JSON.stringify({ patronId })

pages/account/index.tsx
```diff
@@ -94,6 +94,59 @@ export default function MyAccount({
...
+    try {
+      const response = await fetch(
+        `/research/research-catalog/api/account/holds/update/${holdId}`,
+        {
+          method: "POST",
+          headers: {
+            "Content-Type": "application/json",
+          },
+          body: JSON.stringify({
+            patronId: patronId,
```
dgcohen, line 114: See my note about same-name assignments, nbd if you prefer this syntax though.

pages/api/account/helpers.ts
```diff
@@ -0,0 +1,77 @@
...
+    return { status: 200, message: `Pin updated to ${newPin}` }
+  } catch (error) {
+    return {
+      status: error.response.status,
+      message: error.response.data.message || error.response.data.description,
+    }
+  }
+}
+
+export async function updateSettings(patronId: string, patronData: any) {
```
dgcohen, line 25: I think this could use a more descriptive title (updatePatronSettings / updatePatronData?) and possibly a comment.

dgcohen (approved):
Looks good! Just a couple of minor comments.

pages/api/account/helpers.test.ts
dgcohen, line 1: I have an open PR that moves all server-side helper functions to src/server (https://github.com/NYPL/research-catalog/pull/105)

I know it hasn't been merged in yet, but would you mind moving this file there in the meantime to follow the new pattern?

PR #112, opened 2024-02-27
SCC-3614 - Feedback Box

dgcohen:
Ticket: https://jira.nypl.org/browse/SCC-3614

This PR includes the Feedback API route, client-side Context, and Feedback Form component.

I added a helper file testUtils.ts that overrides RTL's render functions and automatically wraps the components in all the app providers to avoid having to wrap the components in each test. Let me know if you disagree with this approach or if you have suggestions for preventing future confusion around this.

I'm currently not testing all the implementation details in the tests (most importantly the API handler), but I'd like to work on that as a follow up so that we can start user testing while I flesh out the tests.


EdwinGuzman (commented):
This looks great! Can you share some screenshot examples of the email that is sent?

My one concern is that closing the `FeedbackForm` box does not return focus to the button. Instead, the focus goes to the top of the page and that is not correct and I know it will be picked up during the accessibility review.

I added a [forceful focus](https://github.com/NYPL/nypl-design-system/pull/1502/files#diff-420431ef7b7b78f5e3c8898e10ff4f117576de7726d80067b3371916e98136faR156) on the button when the form is closed but I'm not sure why it's not working. Could it be a rerendering issue? This does work in Storybook but not here or in DC...

The `FeedbackForm` component is rendered in `Layout`. What if it was rendered higher up in the DOM tree? If you try this in `_app`, you need to move the `<DSProvider>` there which is okay.

pages/api/feedback.ts
```diff
@@ -0,0 +1,60 @@
...
+    aws.config.update({ region: "us-east-1" })
+
+    const fields = JSON.parse(req.body)
+
+    const fullUrl = encodeHTML(req.headers.referer)
+
+    const emailText = getFeedbackEmailText(fullUrl, fields)
+    const emailHTML = getFeedbackEmailHTML(fullUrl, fields)
+
+    const emailParams = {
```
charmingduchess, line 23: is there a way to leverage typescript to detail the required fields here? It's not immediately clear to me what part of this object is required and what is requiring it.

src/components/FeedbackForm/FeedbackForm.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { FeedbackContext } from "../../context/FeedbackContext"
+import type { FeedbackMetadataAndComment } from "../../types/feedbackTypes"
+import { BASE_URL } from "../../config/constants"
+
+/**
+ * Component that wraps the DS Feedback box. Can be opened by clicking the button rendered
+ * or by clicking on 'Contact a librarian' (located in src/app/components/Item/InformationLinks.jsx)
+ */
+const FeedbackForm = () => {
+  const [screen, setScreen] = useState("form")
```
charmingduchess, line 12: can this variable name reflect that its used to populate the view property in the feedback component? It's only used for that, correct?

src/components/ItemTable/ItemAvailability.test.tsx
```diff
@@ -57,11 +54,26 @@ describe("ItemAvailability", () => {
...
     expect(screen.getByText("- Please", { exact: false })).toBeInTheDocument()
-    screen.getByRole("button", {
-      name: "contact a librarian",
-    })
+    expect(screen.getByText("contact a librarian")).toBeInTheDocument()
     expect(
       screen.getByText("for assistance.", { exact: false })
     ).toBeInTheDocument()
   })
+  it("pre-populates the metadata in the feedback form for unavailable items", async () => {
```
charmingduchess, line 62: is there a way to test the presence of the other item metadata? for business logic documentation more than anything else.

src/components/ItemTable/ItemAvailability.tsx
```diff
@@ -13,6 +16,16 @@ interface ItemAvailabilityProps {
  * TODO: Add Feedback box, Due date, Available font styles
  */
 const ItemAvailability = ({ item }: ItemAvailabilityProps) => {
+  const feedback = useContext(FeedbackContext)
+
+  const openFeedbackBox = feedback?.onOpen
```
charmingduchess, line 21: is there a specific case where feedback would be returned undefined?

ENVIRONMENTVARS.md
charmingduchess, line 1: can you add source email and lib answers email here?

dgcohen:
> This looks great! Can you share some screenshot examples of the email that is sent?
> 
> My one concern is that closing the `FeedbackForm` box does not return focus to the button. Instead, the focus goes to the top of the page and that is not correct and I know it will be picked up during the accessibility review.
> 
> I added a [forceful focus](https://github.com/NYPL/nypl-design-system/pull/1502/files#diff-420431ef7b7b78f5e3c8898e10ff4f117576de7726d80067b3371916e98136faR156) on the button when the form is closed but I'm not sure why it's not working. Could it be a rerendering issue? This does work in Storybook but not here or in DC...
> 
> The `FeedbackForm` component is rendered in `Layout`. What if it was rendered higher up in the DOM tree? If you try this in `_app`, you need to move the `<DSProvider>` there which is okay.

Good catch, just pushed a forced focus as a quick fix.

ENVIRONMENTVARS.md
dgcohen: Good catch

src/components/FeedbackForm/FeedbackForm.test.tsx
```diff
@@ -0,0 +1,46 @@
+import React from "react"
+import userEvent from "@testing-library/user-event"
+
+import { render, screen } from "../../utils/testUtils"
+import FeedbackForm from "./FeedbackForm"
+
+describe("FeedbackForm", () => {
```
charmingduchess, line 7: do we want to test the closing/clearing item metadata action?

src/components/ItemTable/ItemAvailability.tsx
dgcohen: only if the provider wasn't wrapping the component. can remove the optional chaining.

src/components/FeedbackForm/FeedbackForm.test.tsx
charmingduchess: oh i forgot you said you haven't finished the testing in your comment. would love to see that in the future.

src/components/FeedbackForm/FeedbackForm.tsx
dgcohen: Good call, fixed.

src/components/FeedbackForm/FeedbackForm.test.tsx
dgcohen: will add this to the follow up testing ticket AC

src/components/ItemTable/ItemAvailability.test.tsx
dgcohen: will add this to the follow up testing ticket AC

charmingduchess (approved):
A couple comments, but nothing that should block testing.

dgcohen:
> This looks great! Can you share some screenshot examples of the email that is sent?
> 
> My one concern is that closing the `FeedbackForm` box does not return focus to the button. Instead, the focus goes to the top of the page and that is not correct and I know it will be picked up during the accessibility review.
> 
> I added a [forceful focus](https://github.com/NYPL/nypl-design-system/pull/1502/files#diff-420431ef7b7b78f5e3c8898e10ff4f117576de7726d80067b3371916e98136faR156) on the button when the form is closed but I'm not sure why it's not working. Could it be a rerendering issue? This does work in Storybook but not here or in DC...
> 
> The `FeedbackForm` component is rendered in `Layout`. What if it was rendered higher up in the DOM tree? If you try this in `_app`, you need to move the `<DSProvider>` there which is okay.

This is what the email looks like when testing locally:

Screenshot 2024-02-29 at 11 16 27 AM

EdwinGuzman (approved):

PR #114, opened 2024-02-28
SCC-3934: Profile tabs

7emansell:
Resolves ticket [SCC-3934](https://jira.nypl.org/browse/SCC-3934).

Adds Tabs component to My Account page.
Adds a redirect from `/account/overdues` to `/account` (for users with no fees).


src/components/MyAccount/profileTabs.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+  //   const { getByText } = render(
+  //     <ProfileTabs
+  //       patron={mockPatron}
+  //       checkouts={mockCheckouts}
+  //       holds={mockHolds}
+  //       fines={mockFines}
+  //       activePath="checkouts"
+  //     />
+  //   )
+  //   fireEvent.click(getByText("Requests"))
```
charmingduchess, line 71: Try using `await userEvent.click` `"@testing-library/user-event"` instead of fireEvent.click

src/components/MyAccount/profileTabs.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <ProfileTabs
+        patron={mockPatron}
+        checkouts={mockCheckouts}
+        holds={mockHolds}
+        fines={mockFines}
+        activePath="checkouts"
+      />
+    )
+    fireEvent.click(getByText("Requests"))
+    expect(useRouter().push).toHaveBeenCalledWith(
```
charmingduchess, line 72: This test is checking internal implementation of url updating. I think it should actually test what the url ends up being.

pages/account/[[...index]].tsx
```diff
@@ -231,14 +240,28 @@ export async function getServerSideProps({ req }) {
       },
     }
   }
+  // Parsing path from url to pass to ProfileTabs.
+  const tabsPath = req.url.split("/", -1)[2] || null
```
Comment (EdwinGuzman) [inline pages/account/[[...index]].tsx:244]: You can do this now or create a ticket for this:

This works for any string that is not part of the tabs because in `ProfileTabs` the default is the first tab -- this is good. But it can be confusing to have a working site and have a bad url, for example `/account/shouldnotwork`:

Screen Shot 2024-03-04 at 1 54 15 PM

So I think you can redirect here, similar to the code you have in 249-256, if `tabsPath` is not in one of the expected urls. This could be if there's typo in the url. I don't expect this to be a big issue since users will be using the tabs and not typing the url directly, but _if_ they do we want to make sure we cover these cases.

EdwinGuzman (approved):
Approving but added a comment for something that should be done either now or later. If now, I can rereview.

pages/account/[[...index]].tsx
Comment (7emansell) [reply pages/account/[[...index]].tsx:244]: Ticket is created and pulled into this sprint! https://jira.nypl.org/browse/SCC-4024 I'd like to make this a hook that handles all /account redirects, similar to the useLogOutRedirect, so we can follow the same testing pattern

charmingduchess (approved):

charmingduchess:
@7emansell just a smalll note. Can you capitalize profileTabs.test.tsx so it matches the casing of the file its testing?

PR #122, opened 2024-03-04
Scc 3927/account settings render

charmingduchess:
Render Account data (without being able to update it)


charmingduchess:
@7emansell I refactored the update path test to use `mockRouter` and `userEvent.click`

src/components/MyAccount/IconListElement.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const IconListElement = ({
+  icon,
+  term,
+  description,
+}: IconListElementPropType) => {
+  return (
+    <>
+      <dt className={styles.iconDt}>
+        <Icon size="large" name={icon} title={`account profile ${term} icon`} />
```
charmingduchess, line 20: @EdwinGuzman I wasn't quite sure what the title of the icon should be indicating. Does this look okay?

EdwinGuzman: It does look okay but, for example, when `term` is "Email:", it'll render as `account profile Email: icon`. The colon can potentially make it confusing for screen readers.
The other part of these `Icon`s is that they have `aria-hidden="true"` by default so they _should_ be hidden to screen readers. This is also okay because they are decorative and the text immediately after describes it.

For now, try to remove the colon and follow up with Clare.

src/components/MyAccount/profileTabs.test.tsx
```diff
@@ -1,23 +1,19 @@
...
   mockPatron,
 } from "../../../__test__/fixtures/accountFixtures"
-import { useRouter } from "next/router"
+import mockRouter from "next-router-mock"
+import userEvent from "@testing-library/user-event"
 
-jest.mock("next/router", () => ({
-  useRouter: jest.fn(),
-}))
+window.HTMLElement.prototype.scrollIntoView = jest.fn()
```
7emansell, line 13: Moved this into `jest.setup` on tabs PR, which makes this implementation of `mockRouter` unnecessary (can just use `fireEvent`)

7emansell (approved):
I think the Profile Tabs test should be left as is, and pending icon review from Clare, but otherwise lgtm

src/components/MyAccount/profileTabs.test.tsx
charmingduchess: ok! i'll remove the mock here, but I think we should be testing the url behavior and not the router.push calls.

7emansell: Ok changed to your implementation

src/components/MyAccount/IconListElement.tsx
charmingduchess: removed the colon!

charmingduchess: Checked with Clare. She confirmed what you were saying - the icon title is redundant and won't be displayed anyway.

src/components/MyAccount/IconListElement.tsx
```diff
@@ -0,0 +1,52 @@
...
+  term: string
+  description: string
+}
+
+const IconListElement = ({
+  icon,
+  term,
+  description,
+}: IconListElementPropType) => {
+  const removeTrailingColon = (term: string) => {
```
EdwinGuzman, line 17: Another option is to remove the colon from the defined string and add the colon on line 28 when the string is rendered, but nbd,

EdwinGuzman (approved):

src/components/MyAccount/IconListElement.tsx
charmingduchess: i hadnt pushed up the changes that actually removed this logic entirely.

PR #124, opened 2024-03-04
Scc 3704/refine results 2024 1 26

charmingduchess:
- add count to filter checkboxes
- transform greek language option


dgcohen (approved):
Looks good!

PR #125, opened 2024-03-04
SCC-4019: Account API documentation

7emansell:
Resolves [SCC-4019](https://jira.nypl.org/browse/SCC-4019).

Documents Sierra API endpoints for account page.

### DIFF
--- a/accountREADME.md
+++ b/accountREADME.md
@@ -1,0 +1,13 @@
+# My Account README
+
+This README details the Sierra API endpoints that now support [Research Catalog's My Account page](https://www.nypl.org/research/research-catalog/account) (instead of the webscraping previously used).
+
+## Account API Endpoints
+
+| Description            | Request method | Internal route                              | Example request body (\* is required)                            |
+| ---------------------- | -------------- | ------------------------------------------- | ---------------------------------------------------------------- |
+| Update hold request    | POST           | `/api/account/holds/update/{holdId}`        | `{"patronId"*: 123456}`                                          |
+| Cancel hold request    | POST           | `/api/account/holds/cancel/{holdId}`        | `{"patronId"*: 123456, "freeze": false, "pickupLocation": "sn"}` |
+| Update patron settings | PUT            | `/api/account/settings/{patronId}`          | `{"emails": ['new@email.com], "phones": [6466600432]}`           |
+| Update patron PIN      | PUT            | `/api/account/update-pin/{patronId}`        | `{"oldPin"*: 1234, "newPin"*: 7890, "barcode"*: 123456789}`      |
+| Renew checkout         | PUT            | `/api/account/checkouts/renew/{checkoutId}` | `{"patronId"*: 123456}`                                          |
--- a/accountREADME.md
+++ b/accountREADME.md
@@ -1,15 +1,65 @@
-# My Account README
+# My Account: Phase 1
 
-This README details the Sierra API endpoints that now support [Research Catalog's My Account page](https://www.nypl.org/research/research-catalog/account) (instead of the webscraping previously used).
+This README details the Sierra API endpoints that now support [Research Catalog's My Account page](https://www.nypl.org/research/research-catalog/account) (instead of the webscraping previously used). See [Sierra API documentation](https://sandbox.iii.com/iii/sierra-api/swagger/index.html#!/patrons) for more information.
 
-## Account API Endpoints Cheatsheet
+## Account API Endpoints
 
-| Description            | Request method | Internal route                              | Example request body (\* is required field)                      |
-| ---------------------- | -------------- | ------------------------------------------- | ---------------------------------------------------------------- |
-| Update hold request    | POST           | `/api/account/holds/update/{holdId}`        | `{"patronId"*: 123456, "freeze": false, "pickupLocation": "sn"}` |
-| Cancel hold request    | POST           | `/api/account/holds/cancel/{holdId}`        | `{"patronId"*: 123456`                                           |
-| Update patron settings | PUT            | `/api/account/settings/{patronId}`          | `{"emails": ['new@email.com], "phones": [6466600432]}`           |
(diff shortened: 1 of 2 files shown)


7emansell:
Not sure what else to include for now, since I don't want to just repeat the TAD. But open to adding more sections etc.

EdwinGuzman (commented):
Just a minor update for one request body example.

This might be too much but what do you think about keeping this table as an introduction for this document but remove the "Example request body..." column?

Then you can make each endpoint a section and have it's expected request query and/or body parameters and what the success/error response looks like. Examples of what I'm thinking https://github.com/NYPL/digital-collections?tab=readme-ov-file#individual-lane---get-apilanesslug and https://github.com/NYPL/dgx-patron-creator-service/wiki/API-V0.3

I don't want this to hold up dev work so this could be worked on at the end.

7emansell:
> This might be too much but what do you think about keeping this table as an introduction for this document but remove > the "Example request body..." column?

That sounds good to me!

accountREADME.md
```diff
@@ -0,0 +1,156 @@
...
+    newPin: 7890,
+    barcode: 123456789
+},
+
+```
+
+```
+exampleSuccessResponse: {
+    status: 200,
+    message: 'Pin updated to 7890'
```
7emansell, line 102: I anticipate changing this in the update PIN pr since I don't think it's a good idea to return the new pin in the response :)

EdwinGuzman (changes requested):
403 are usually for forbidden endpoints and the endpoints do return this when the patron is not authenticated.

For other errors, it looks like the error code from Sierra is propagated to the returned response. So there should be more exampleFailureResponse objects with multiple 4XX statuses.

accountREADME.md
```diff
@@ -0,0 +1,156 @@
...
+            callNumber: 'J SPA 551.46 R'
+        },
+        dueDate: '2024-03-15T08:00:00Z',
+        numberOfRenewals: 10,
+        outDate: '2024-02-14T22:05:04Z'
+    },
+  }
+}
+exampleFailureResponse: {
+    status: 403,
```
EdwinGuzman, line 152: Regarding the 403 response, there are multiple ones returned for this endpoint (and possibly others):

```
status: 403,
message: "No authenticated patron"
```
and
```
status: 403,
message: "Authenticated patron does not own this checkout"
```

List them all down. Ideally, I think a response status code should only have one error message but that can be discussed later.

7emansell:
@EdwinGuzman For every endpoint, getting a 403 for a patron that's not authenticated or doesn't match the item they're updating is possible. Should I collect those example responses in one place or just repeat them for each? Also, a recurring theme with Sierra work is that we can't predict every possible thing that it will return, but we know it will use code 403 for many different errors. Ideally, once logging is implemented, those errors will be recorded on Cloudwatch. What would be the best way to represent this in with the examples in the docs?

accountREADME.md
```diff
@@ -0,0 +1,181 @@
...
+    message: 'No authenticated patron'
+}
+exampleFailureResponse: {
+    status: 403,
+    message: 'Authenticated patron does not own this hold'
+}
+
+```
+
+If the request fails because it's rejected by Sierra (a common example is a checkout being renewed too many times), it will return a `4XX` status, with a message from Sierra.
```
EdwinGuzman, line 21: You also get the status from the rejected Sierra request? It's mostly likely a 400 bad request or something else? Not for this PR but either return whatever Sierra returns or set a status other than 403 for bad Sierra requests.

EdwinGuzman (approved):
Looks good. I'm being picky about the 403 for Sierra-related errors but that's a potential code update and not a readme update.

7emansell:
@EdwinGuzman we get the status from Sierra as well (ex. https://github.com/NYPL/research-catalog/blob/ae9dee52f7be3dbe65415c3857b5259bdee1e5b7/pages/api/account/helpers.ts#L23). The user is never seeing any of this, so how we handle this would really just be for dev troubleshooting, and I think it makes sense that we just pass whatever Sierra's saying straight through. I'm wary of misleading in the docs by saying it's "most likely" for this reason bc we just have seen so much unexpected stuff, there's a lot of reasons why a checkout couldn't be renewed in particular

EdwinGuzman:
Yeah, that's fair. If you catch all the errors so the user never sees it, that's great. For whatever reason a checkout couldn't be renewed, or another error happens in Sierra, it's not that step that I'm concerned about but rather that it's still 403 for the client. When there is an auth issue, 403 is fine but 400 might be better for a bad request.

`updatePin` does pass whatever error code is returned from the Sierra call, but it's not being caught so in this case this would be an uncaught error situation. What I mean happens here https://github.com/NYPL/research-catalog/blob/ae9dee52f7be3dbe65415c3857b5259bdee1e5b7/pages/api/account/update-pin/%5Bid%5D.ts#L31 there's no try/catch for this error.

If you do catch the error, what I'm getting at is that 400 might be a better status code:

```
if (patronId == cookiePatronId) {
  try {
    const response = await updatePin(patronId, barcode, oldPin, newPin)
    responseStatus = response.status
    responseMessage = response.message
  } catch (e) {
    responseStatus = 400
    // This is a made up message but it could be `e.message`
    responseMessage = "Something went wrong updating the pin in Sierra"
} else {
  responseStatus = 403
  responseMessage = "Authenticated patron does not match request"
}
```

We can go over this on Monday. This does not block this PR.

EdwinGuzman:
I'm not going to delete the comment but ignore it. This _is_ happening and now I see what you're saying, it was just hard to read and I didn't catch it right away.

You're right, that it can be misleading in the docs because Sierra can return any status and any error message. I think "4XX" is more appropriate than "403" in the docs though because we know that the patron is authenticated when the call is made, but that the Sierra call itself had a different error.

dgcohen (approved):

PR #133, opened 2024-03-06
update husky script and install tsc-files

charmingduchess:
This PR includes the installation tsc-files and replacing our npm run scripts with lint-staged. This work allows us to lint [run type checks only on staged files](https://github.com/microsoft/TypeScript/issues/27379). This really shortens the time needed to commit something and also will help me avoid a nervous breakdown. 
I've run this on another branch and it successfully commits when code has no type errors, and blocks commits when there are type errors.


charmingduchess:
Alas. Does not play nice with scss files. 😢

PR #134, opened 2024-03-06
SCC-3933: Account checkout tab

7emansell:
Resolves [SCC-3933](https://jira.nypl.org/browse/SCC-3933).

Adds checkout tab to account tabs. 

Also, out of scope, adds date formatting to account model.


7emansell:
Noting that the renewal modal still has the X to close in the corner, which is not in the design– will check with Apoorva when she gets back if this is okay, since DS Modal makes it pretty difficult to remove

pages/account/[[...index]].tsx
```diff
@@ -183,43 +183,43 @@ export default function MyAccount({
               fines={fines}
               activePath={tabsPath}
             />
-            {/** Testing renew checkout api route, with test checkout id. */}
+            {/* * Testing renew checkout api route, with test checkout id.
```
Comment (charmingduchess) [inline pages/account/[[...index]].tsx:186]: Can you remove this test checkout button?

src/components/MyAccount/CheckoutsTab.tsx
```diff
@@ -0,0 +1,243 @@
...
+          isDisabled={isButtonDisabled}
+        >
+          Renew
+        </Button>
+        <Modal {...modalProps} />
+      </>
+    )
+  }
+
+  function getDueDate(date) {
```
charmingduchess, line 166: can `getDueDate`, `getTitle` live on the Patron class so the data is passed to this component without needing a transformation?

src/components/MyAccount/CheckoutsTab.tsx
```diff
@@ -0,0 +1,243 @@
...
+        setModalProps(successModalProps)
+        onOpen()
+        localStorage.setItem(
+          `lastDisabledTime-${checkout.id}`,
+          new Date().getTime().toString()
+        )
+      } else {
+        setModalProps(failureModalProps)
+        onOpen()
+        //TO-DO: Log error console.log("error", responseData)
```
charmingduchess, line 144: is this a todo because the AC are not defined?

src/components/MyAccount/CheckoutsTab.tsx
```diff
@@ -0,0 +1,243 @@
...
+            display: "flex",
+            justifyContent: "flex-start",
+          }}
+        >
+          <span>You currently do not have any research items checked out.</span>
+        </Box>
+      )}
+      <Box
+        sx={{
+          fontSize: "desktop.body.body2",
```
charmingduchess, line 209: since these styles are going to be used for the requests tab as well, can they be exported to MyAccount.scss?

src/components/MyAccount/CheckoutsTab.tsx
charmingduchess: we dont have a logger yet

EdwinGuzman: This is client side so the logger (I'm assuming something like winston) would not work here.

7emansell: We will have a client side at some point though? Otherwise this error can be logged in the helper

src/components/MyAccount/ProfileTabs.tsx
```diff
@@ -56,6 +57,7 @@ const ProfileTabs = ({
 
   return (
     <Tabs
+      sx={{ "div[role=tabpanel]": { padding: "0px" } }}
```
EdwinGuzman, line 60: Hmm I can see why this can be tedious and annoying.

src/utils/myAccountUtils.tsx
```diff
@@ -0,0 +1,12 @@
+import { Link, Text } from "@nypl/design-system-react-components"
+
+export function getTitle(checkout) {
```
EdwinGuzman, line 3: I'm not opposed to this but this is interesting because we were trying to move away from functions that did logic and returned html. Based on what it returns, it seems fair to make it this way because otherwise there would be logic in the react component for what it should render. Is that right?

src/components/MyAccount/CheckoutsTab.tsx
```diff
@@ -0,0 +1,165 @@
...
+}) => {
+  const checkoutsHeaders = [
+    "Title",
+    "Barcode",
+    "Call number",
+    "Due back by",
+    "Manage checkout",
+  ]
+
+  function RenewButton(checkout, patron) {
```
EdwinGuzman, line 31: This component is big enough that it should be defined outside of `CheckoutsTab`. Keeping it in this file is fine or consider moving it to a new file.

EdwinGuzman, line 31: Give the args the proper types.

src/components/MyAccount/CheckoutsTab.tsx
```diff
@@ -0,0 +1,165 @@
...
+            <Link href="https://www.nypl.org/get-help/contact-us">
+              contact us
+            </Link>{" "}
+            for assistance.
+          </Text>
+        </Box>
+      ),
+      closeButtonLabel: "OK",
+      headingText: (
+        <Box className={styles.modalHeading}>
```
EdwinGuzman, line 69: The heading is not displaying.
Screen Shot 2024-03-08 at 10 29 34 AM

This might be too restrictive but we expect an `<h*>` HTML element or a DS `Heading` component. Replace `Box` with `<Heading ...><>...</></Heading>`. Some styles will need an update:
Screen Shot 2024-03-08 at 10 49 28 AM

src/components/MyAccount/CheckoutsTab.tsx
```diff
@@ -0,0 +1,165 @@
...
+      const lastDisabledTime = localStorage.getItem(
+        `lastDisabledTime-${checkout.id}`
+      )
+      if (lastDisabledTime) {
+        const twentyFourHoursAgo = new Date().getTime() - 8640000
+        if (parseInt(lastDisabledTime, 10) > twentyFourHoursAgo) {
+          setButtonDisabled(true)
+        }
+      }
+    }, [])
```
EdwinGuzman, line 86: If this is meant to only run once on load, why not include it as part of the component logic and not use the `useEffect`?

EdwinGuzman: The linter also complains that `checkout.id` is missing from the dep array.

src/components/MyAccount/CheckoutsTab.tsx
```diff
@@ -0,0 +1,165 @@
...
+        setButtonDisabled(true)
+        setModalProps(successModalProps)
+        onOpen()
+        localStorage.setItem(
+          `lastDisabledTime-${checkout.id}`,
+          new Date().getTime().toString()
+        )
+      } else {
+        setModalProps(failureModalProps)
+        onOpen()
```
EdwinGuzman, line 110: Since `onOpen` is called regardless, it can moved to after the if/else block.

src/components/MyAccount/CheckoutsTab.tsx
```diff
@@ -0,0 +1,165 @@
...
+        //TO-DO: Log error console.log("error", responseData)
+      }
+    }
+
+    return (
+      <>
+        <Button
+          sx={{
+            width: "-webkit-fill-available",
```
EdwinGuzman, line 119: Why not "100%"?

src/components/MyAccount/CheckoutsTab.tsx
7emansell: For sure but this is not happening locally for me ? Why would that be

src/utils/myAccountUtils.tsx
EdwinGuzman: It's a separation of concern conversation we can have with Vera. What I'm getting at is that this function could return just the necessary data and not the react components. Then, in `CheckoutsTab`, that's where the logic to render either a Link or Text would be. This does make the UI more tedious. 

Again, I'm not opposed to this but we should be careful with this pattern. It makes it confusing to know where the UI is rendered on a page and that's a big problem we had in DFE.

src/components/MyAccount/CheckoutsTab.tsx
EdwinGuzman: I'll test again. Maybe I did not run `npm install` to get the updated ds version?

src/utils/myAccountUtils.tsx
7emansell: Ah I see. On consideration: I do think this should either return data, and remain outside the component, or should be on the component and return components.  Cool w either @charmingduchess

src/components/MyAccount/CheckoutsTab.tsx
7emansell: It does seem weird but i thought that `localStorage` should be used inside `useEffect` since it's a non react state. Also linter doesn't like it but `checkout.id` isn't really a logical dependency i think

EdwinGuzman (approved):
I'm approving because the functionality and code look good.

I'm still concern about all the scss overrides and we briefly chatted about this, but it's not a blocker. I think we can revisit this in a later "cleanup" ticket to focus on the overrides and/or updates that had to be done. It's also something I'd like to include Marty and Apoorva about.

src/components/MyAccount/CheckoutsTab.tsx
```diff
@@ -0,0 +1,67 @@
...
+      return <Text>{checkout.title}</Text>
+    }
+  }
+
+  const checkoutsData = checkouts.map((checkout) => [
+    formatTitle(checkout),
+    checkout.barcode,
+    checkout.callNumber,
+    checkout.dueDate,
+    checkout.isResearch ? null : RenewButton(checkout, patron),
```
charmingduchess, line 40: Can the RenewButton be used more like a component? as in, turning up here as `<RenewButton checkout={checkout} patron={patron}/>`?

src/types/accountTypes.ts
```diff
@@ -70,6 +70,7 @@ export interface Checkout {
   title: string
   bibId?: string
   isNyplOwned: boolean
+  href: string
```
charmingduchess, line 73: can we call this property catalogHref for clarity?

charmingduchess (commented):
Just a couple minor notes. The refactoring looks great! Thanks for making the renew button a separate component and extracting the CSS.

charmingduchess (approved):

dgcohen (approved):

PR #140, opened 2024-03-08
Scc 4027/add dates 2

charmingduchess:
This PR incorporates the date fields into the refine search dialog and the filter tag set. 

- The date fields, when inputs are provided and the refine search filters are applied, populate the url query params. 
- Those applied filters, including the dates, populate the filter tag set in the search results view. 
- A url with date params set will populate the date fields of the refine search filters

### DIFF
--- a/src/components/RefineSearch/RefineSearch.tsx
+++ b/src/components/RefineSearch/RefineSearch.tsx
@@ -10,8 +10,7 @@ import { useState, useCallback } from "react"
 import { useRouter } from "next/router"
 
 import styles from "../../../styles/components/Search.module.scss"
-import type { DateFormName } from "../SearchFilters/FieldsetDate"
-import FieldsetDate from "../SearchFilters/FieldsetDate"
+import FieldsetDate, { type DateFormName }from "../SearchFilters/FieldsetDate"
 import SearchResultsFilters from "../../models/SearchResultsFilters"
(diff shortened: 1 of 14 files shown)


src/components/RefineSearch/RefineSearch.tsx
```diff
@@ -10,6 +10,8 @@ import { useState, useCallback } from "react"
 import { useRouter } from "next/router"
 
 import styles from "../../../styles/components/Search.module.scss"
+import type { DateFormName } from "../SearchFilters/FieldsetDate"
```
dgcohen, line 13: Small nit: you can combine these into one import statement:

import FieldsetDate, { type DateFormName } from "../SearchFilters/FieldsetDate"

src/utils/refineSearchUtils.ts
```diff
@@ -45,13 +45,24 @@ export const addLabelPropAndParseFilters = (
...
     ].map((filterValue: string): Option => {
+      // dateBefore and dateAfter fields are not based on
+      // aggregations results. Pass the year along with out
+      // transforming fieldname or finding the label
+      if (appliedFilterField.includes("date")) {
+        const labelPrefix = appliedFilterField.split("date")[1]
+        return {
+          count: null,
+          value: filterValue,
+          label: labelPrefix + " " + filterValue,
```
dgcohen, line 59: Small nit/preference: I prefer to use string interpolation for this kind of thing.

PR #142, opened 2024-03-12
Only linting on commit hook

charmingduchess:
This PR removes the check-types script from the pre-commit hook, but not from package.json. 
Husky will instead run lint-staged, which includes eslint and prettier autoformatting.


dgcohen (approved):
Looks good!

EdwinGuzman (approved):

PR #143, opened 2024-03-12
SCC-3935: Requests tab

7emansell:
Resolves [SCC-3935](https://jira.nypl.org/browse/SCC-3935).

Adds Requests tab to My Account page, pulling the tab structure from Checkouts into a shared ItemTab component. 
This is only functional for research (item level) holds (slack me for test patron)


src/components/MyAccount/ItemsTab.tsx
```diff
@@ -0,0 +1,40 @@
...
+import styles from "../../../styles/components/MyAccount.module.scss"
+
+const ItemsTab = ({
+  headers,
+  data,
+  verb,
+}: {
+  headers: string[]
+  data: any[]
+  verb: string
```
charmingduchess, line 11: is this the cancel and or freeze buttons?

src/components/MyAccount/ItemsTab.tsx
charmingduchess, line 1: I really like what you did to reuse the ItemsTab! Can you use a word besides verb? And since it is only one of two strings, can you make the type be the explicit string values? I think it would make it clearer looking at this file instead of navigating to other files to see what the options are.

src/components/MyAccount/ProfileTabs.tsx
```diff
@@ -16,16 +18,37 @@ const ProfileTabs = ({
   fines: Fine
   activePath: string
 }) => {
+  // currentHolds is a copy of the holds local to this component.
+  const [currentHolds, setCurrentHolds] = useState(holds)
+  /* handleState removes the passed hold from currentHolds, so page doesn't need to
+   * reload for the request to disappear. */
+  function handleHoldsState(hold) {
+    setCurrentHolds(
+      currentHolds.reduce((acc, item) => {
```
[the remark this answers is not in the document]
charmingduchess: Also, if this method only ever removes a hold, can the name reflect that?

src/components/MyAccount/RequestsTab/FreezeButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Button } from "@nypl/design-system-react-components"
+
+const FreezeButton = ({ hold, patron }: { hold: Hold; patron: Patron }) => {
+  const [frozen, setFrozen] = useState(hold.frozen)
+  const [isDisabled, setIsDisabled] = useState(false)
+
+  const handleFreezeClick = async () => {
+    // Disabling button while request happens.
+    setIsDisabled(true)
+    if (frozen) {
```
charmingduchess, line 12: can this if/else be eliminated by using the value of `frozen` for the property `freeze`? the `setFrozen` call too can use `!frozen` in that case.

src/components/MyAccount/RequestsTab/FreezeButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+          body: JSON.stringify({
+            patronId: patron.id,
+            freeze: false,
+            pickupLocation: hold.pickupLocation.code,
+          }),
+        }
+      )
+      setFrozen(false)
+    } else {
+      await fetch(
```
charmingduchess, line 29: I keep going back and forth about this... I'm wondering if it would be worth it to make a generic fetch call that would take a path, method, and body. or maybe the answer is to take those test methods from the [[index]] page and keep those all in a file somewhere. these options objects take up a lot of visual space in the code and don't have anything to do with what's rendered. @EdwinGuzman do you have thoughts on this too?

src/components/MyAccount/profileTabs.test.tsx
```diff
@@ -61,7 +61,7 @@ describe("ProfileTabs", () => {
         activePath="checkouts"
       />
     )
-    fireEvent.click(getByText("Requests"))
+    fireEvent.click(getByText("Requests (2)"))
```
charmingduchess, line 64: This is fine as is, but for the future, you can pass in a second options param with `{exact: false}` if it's something like this that is liable to change if we have more test cases.

src/components/MyAccount/CheckoutsTab/CheckoutsTab.tsx
```diff
@@ -0,0 +1,45 @@
...
+}) => {
+  const checkoutsHeaders = [
+    "Title",
+    "Barcode",
+    "Call number",
+    "Due back by",
+    "Manage checkout",
+  ]
+
+  function formatTitle(checkout: Checkout) {
```
charmingduchess, line 21: can this function name reflect that it's returning an html element?

src/components/MyAccount/RequestsTab/FreezeButton.tsx
EdwinGuzman: Yes, I agree there should be a function that wraps the fetch call. If it's created for this PR, great, but it can also wait for another PR.

Emma, this is similar to the `apiCall` function in digital-collections. The function for this app may need more arguments though.

src/components/MyAccount/RequestsTab/RequestsTab.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+  it("renders", () => {
+    render(
+      <ProfileTabs
+        patron={mockPatron}
+        checkouts={mockCheckouts}
+        holds={mockHolds}
+        fines={mockFines}
+        activePath="requests"
+      />
+    )
```
EdwinGuzman, line 33: Add an expectation that _something_ renders.

src/components/MyAccount/RequestsTab/RequestsTab.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+        body: JSON.stringify({
+          patronId: mockPatron.id,
+          freeze: true,
+          pickupLocation: "mal",
+        }),
+      }
+    )
+
+    const unfreezeButton = component.getAllByText("Unfreeze")[0]
+
```
EdwinGuzman, line 148: Can you add expectations here to verify that the freeze button is no longer rendered?

src/components/MyAccount/RequestsTab/RequestsTab.test.tsx
```diff
@@ -0,0 +1,169 @@
+import React from "react"
```
EdwinGuzman, line 1: These tests are really good but if possible, render the `RequestTab` component instead of `ProfileTabs`. You can use `useState` in test files so you can mock the `handleHoldsState` function.

src/components/MyAccount/RequestsTab/RequestsTab.tsx
```diff
@@ -0,0 +1,76 @@
...
+          handleHoldsState={handleHoldsState}
+          hold={hold}
+          patron={patron}
+        />
+        {hold.canFreeze && <FreezeButton hold={hold} patron={patron} />}
+      </Box>
+    ) : null,
+  ])
+
+  function formatTitle(hold: Hold) {
```
EdwinGuzman, line 50: Nitpick: move the functions up in the file before they are called.

src/components/MyAccount/profileTabs.test.tsx
EdwinGuzman: Yes, another option is `getByText(/requests/i)`

src/models/MyAccount.ts
```diff
@@ -96,7 +96,7 @@ export default class MyAccount {
           (subfield: { tag: string; subfield: string }) => subfield.tag === "a"
         ).content
         isResearch = nineTenContent.startsWith("RL")
-        isNyplOwned = nineTenContent !== "RLOTF"
+        isNyplOwned = !isResearch || nineTenContent !== "RLOTF"
```
EdwinGuzman, line 99: What does RLOTF stand for?...

styles/components/MyAccount.module.scss
```diff
@@ -72,6 +72,27 @@
     border-color: var(--nypl-colors-ui-gray-light-cool);
   }
   th:first-of-type {
-    width: 413px;
+    width: 427px;
```
EdwinGuzman, line 75: Eyeballing it?

styles/components/MyAccount.module.scss
```diff
@@ -72,6 +72,27 @@
...
+  font-weight: 510;
+  font-style: normal !important;
+}
+
+.statusBadgeReady {
+  background: rgba(238, 247, 240, 1) !important;
+  border-color: var(--nypl-colors-ui-success-primary) !important;
+  border-left: 4px solid;
+  font-size: var(--nypl-fontSizes-desktop-caption);
+  font-weight: 510;
```
EdwinGuzman, line 96: "510" is used a lot in this PR. The closest DS design token value for this is "medium" which makes to "500" -- in this case `font-weight: var(--nypl-fontWeights-medium);`. Likewise, in DS components you can pass `fontWeight: "medium"`.

Use the design token value to align with the DS value.

EdwinGuzman (changes requested):
Requesting some changes but overall this is very good! I tested cancelling holds but not freezing/unfreezing so I may need a new test account for that.

Something I'd like to discuss with you but that DOES NOT need to be done in this PR is the architecture for the cancel/renew buttons and the modal. A `Modal` is being rendered for each cancel/renew button. I'm wondering if this can be done so that there is a single Modal instance that is triggered and all the relevant data is passed to it, similar to how the `modalProps` is structured but on a different level in the DOM tree. I think I'm asking for something that may change how the components and data are structured so just a discussion would be fine to have first.

Realistically, users can only trigger one modal so we're not going to see multiple modals pop up anyway. This would be a good refactor ticket to work on if we find out that it can be done. This is not a blocker for release.

src/components/MyAccount/RequestsTab/CancelButton.tsx
```diff
@@ -0,0 +1,118 @@
...
+}) => {
+  const { onOpen: openModal, onClose: closeModal, Modal } = useModal()
+
+  function confirmModalProps(hold) {
+    return {
+      bodyContent: (
+        <Box className={styles.modalBody}>
+          <Text sx={{ marginLeft: "l", marginRight: "m" }}>
+            Your request for{" "}
+            <span style={{ fontWeight: 510 }}>{hold.title}</span> has been
```
dgcohen, line 33: Normally font weights are multiples of 100, but i think a DS variable can be used instead of a number here.

src/components/MyAccount/RequestsTab/CancelButton.tsx
```diff
@@ -0,0 +1,118 @@
...
+            canceled.
+          </Text>
+        </Box>
+      ),
+      closeButtonLabel: "OK",
+      headingText: (
+        <Heading className={styles.modalHeading}>
+          <>
+            <Icon size="large" name="errorFilled" color="ui.error.primary" />
+            <Text sx={{ marginBottom: "0px" }}>Request canceled </Text>
```
dgcohen, line 43: Small nit but px isn't necessary for 0 values.

src/components/MyAccount/RequestsTab/CancelButton.tsx
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  function checkModalProps(hold) {
+    return {
+      bodyContent: (
+        <Box className={styles.modalBody}>
+          <Text>
+            Are you sure that you want to cancel your request for{" "}
+            <span style={{ fontWeight: 510 }}>{hold.title}</span>?
```
dgcohen, line 61: Same as above.

src/components/MyAccount/RequestsTab/CancelButton.tsx
```diff
@@ -0,0 +1,118 @@
...
+            {
+              method: "POST",
+              headers: {
+                "Content-Type": "application/json",
+              },
+              body: JSON.stringify({ patronId: patron.id }),
+            }
+          )
+          const responseData = await response.json()
+          if (responseData == "Canceled") {
```
dgcohen, line 88: There's a typo here, but I think I'd prefer if we used a status code instead of a string to determine if there was an error.

src/server/tests/helpers.test.ts
```diff
@@ -21,7 +21,7 @@ const mockCheckoutResponse = {
 }
 
 describe("cancelHold", () => {
-  it("should return a success message if hold is cancelled", async () => {
+  it("should return a success message if hold is canceled", async () => {
```
dgcohen, line 24: Thought this was a typo, but google is telling me both spellings are correct haha

src/components/MyAccount/RequestsTab/CancelButton.tsx
dgcohen: Nvm about the typo (a google search corrected me), but I do think we should return a status code.

src/models/MyAccount.ts
nonword: "Research Library On The Fly". It's a code we add to OTF (aka "virtual") records, to tag them as being Research OTF records.

src/components/MyAccount/RequestsTab/CancelButton.tsx
```diff
@@ -0,0 +1,110 @@
...
+          {
+            method: "POST",
+            headers: {
+              "Content-Type": "application/json",
+            },
+            body: JSON.stringify({ patronId: patron.id }),
+          }
+        )
+        const responseData = await response.json()
+        if (responseData == "Cancelled") {
```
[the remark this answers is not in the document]
7emansell: Agreed. This is the first draft of AC for the checkout/request tabs https://docs.google.com/document/d/1J45SCj6a8T7IaWuGedG0uV21BE909ekcDsa2TdFWh7k/edit so if we don't address it in this PR, it's noted here and will definitely need some sort of solution/decision

src/server/tests/helpers.test.ts
7emansell: Ik lol I wrote all of this code just randomly switching between the british/american spelling which was not ideal– they should all be 'Canceled' now

src/components/MyAccount/ProfileTabs.tsx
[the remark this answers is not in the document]
charmingduchess: I mean `handleHoldsState`, not`setCurrentHolds`

src/components/MyAccount/RequestsTab/CancelButton.tsx
```diff
@@ -0,0 +1,110 @@
...
+const CancelButton = ({
+  handleHoldsState,
+  hold,
+  patron,
+}: {
+  handleHoldsState: (hold: Hold) => void
+  hold: Hold
+  patron: Patron
+}) => {
+  const { onOpen: openModal, onClose: closeModal, Modal } = useModal()
```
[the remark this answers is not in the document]
charmingduchess: The prop name is fine for me. I get confused with the onClose and onOpen methods from the useModal hook, because they're not programmable callbacks.

src/models/MyAccount.ts
charmingduchess: I was thinking this line could use a comment to explain to future devs

src/models/MyAccount.ts
```diff
@@ -113,11 +113,16 @@ export default class MyAccount {
         canFreeze: hold.canFreeze,
         frozen: hold.frozen,
         status: MyAccount.getHoldStatus(hold.status),
-        pickupLocation: hold.pickupLocation.name,
+        pickupLocation: hold.pickupLocation,
         title: bibDataMap[hold.record.bibIds[0]].title,
         isResearch: bibDataMap[hold.record.bibIds[0]].isResearch,
         bibId: hold.record.bibIds[0],
-        isNyplOwned: bibDataMap[hold.record.bibIds[0]].isResearch,
+        isNyplOwned: bibDataMap[hold.record.bibIds[0]].isNyplOwned,
```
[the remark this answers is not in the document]
nonword: Yes, I think, as a convenience we only render the title of the first bib when an item is a bound-with. Ideally we would display all titles because the patron may have requested an item on one bib and be surprised to see the title of a different bib in their holds/checkouts. Would be better to show all titles even crudely (e.g. `hold.record.bibIds.map((id) => bibDataMap[id].title).join('; ')`)

For the purpose of determining "is-nypl-owned", just checking the first bib is enough because an item will never contain a mix of institution bibs.

styles/components/MyAccount.module.scss
charmingduchess: sorry this didnt get communicated, but this styling is not going to be necessary once we incorporate the v3 update that includes the new status badges. So don't worry about changing this!

EdwinGuzman (approved):
I'm approving for the changes I requested but I can review again when other feedback is addressed.

.eslintrc.js
```diff
@@ -49,5 +49,6 @@ module.exports = {
         format: ["camelCase", "PascalCase", "UPPER_CASE"],
       },
     ],
+    "@typescript-eslint/ban-ts-comment": ["off"],
```
charmingduchess, line 52: I remember you saying this line didn't end up allowing the ts-ignore line. Is that still the case?

src/components/MyAccount/ItemsTab.tsx
```diff
@@ -0,0 +1,42 @@
...
+  headers: string[]
+  data: any[]
+  userAction: "requested" | "checked out"
+}) => {
+  return (
+    <>
+      {data?.length === 0 && (
+        <Box className={styles.notification}>
+          <span>
+            You currently do not have any research items {userAction}.
```
charmingduchess, line 18: I'm confirming with Apoorva to see if this is what we want to say here.

.eslintrc.js
7emansell: Yes but it's still needed to allow the `ts-nocheck` which I am using

src/components/MyAccount/ItemsTab.tsx
charmingduchess: She says:
> Good point! Changing the language to just 'items requested'

7emansell: But should the checkouts tab still specifically say "research items checked out" ?

charmingduchess: You can remove the word "research".

src/models/MyAccount.ts
charmingduchess: Nitpick on the comment - RL is the 910 content for all research materials, OTF is added specifically for virtual records.

7emansell:
This PR is now only waiting on the DS to be updated in `main`

charmingduchess (approved):
I believe the v3 DS work has [already been merged to main](https://github.com/NYPL/research-catalog/pull/148), so just clear up the merge conflicts and do the thing!

charmingduchess:
🥳

PR #144, opened 2024-03-13
SCC-3936: Fees tab

7emansell:
Resolves [SCC-3936](https://jira.nypl.org/browse/SCC-3936).

Adds Fees tab to My Account page.

### DIFF
--- a/pages/account/[[...index]].tsx
+++ b/pages/account/[[...index]].tsx
@@ -1,5 +1,9 @@
 import Head from "next/head"
-import { Button, Text } from "@nypl/design-system-react-components"
+import {
+  Button,
+  Notification,
+  Text,
+} from "@nypl/design-system-react-components"
 import Layout from "../../src/components/Layout/Layout"
 import initializePatronTokenAuth, {
   getLoginRedirect,
@@ -149,6 +153,16 @@ export default function MyAccount({
           </Text>
         ) : (
           <>
+            <Notification
+              className="feesNotification"
+              notificationType="warning"
+              showIcon
+              ariaLabel="Fees warning"
+              id="notification-id"
+              notificationContent="You have outstanding fees. Borrowing privileges will be suspended for cardholders with replacement fees totaling $100 or more.
+Fees can be paid at any New York Public Library branch in cash, U.S. Postal money order, personal check, or online through the Library website."
+            />
+
             <ProfileHeader patron={patron} />
             <ProfileTabs
               patron={patron}
--- a/pages/account/[[...index]].tsx
+++ b/pages/account/[[...index]].tsx
@@ -1,6 +1,8 @@
 import Head from "next/head"
 import {
   Button,
+  Icon,
+  Link,
   Notification,
   Text,
 } from "@nypl/design-system-react-components"
@@ -13,6 +15,7 @@ import type { Checkout, Hold, Patron, Fine } from "../../src/types/accountTypes"
 import ProfileTabs from "../../src/components/MyAccount/ProfileTabs"
 import ProfileHeader from "../../src/components/MyAccount/ProfileHeader"
 import { BASE_URL } from "../../src/config/constants"
+import styles from "../../styles/components/MyAccount.module.scss"
 
 interface MyAccountPropsType {
   checkouts?: Checkout[]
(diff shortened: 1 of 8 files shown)


pages/account/[[...index]].tsx
```diff
@@ -149,6 +156,35 @@ export default function MyAccount({
           </Text>
         ) : (
           <>
+            {fines.total > 0 && (
```
Comment (7emansell) [inline pages/account/[[...index]].tsx:159]: Is the below notification component big enough to pull out?

styles/components/MyAccount.module.scss
```diff
@@ -96,3 +119,30 @@
...
+    gap: 4px;
+    flex-direction: row;
+    > p {
+      margin-bottom: 0px;
+      color: var(--nypl-colors-ui-error-primary);
+      font-size: var(--nypl-fontSizes-desktop-body-body1);
+    }
+    > svg {
+      fill: var(--nypl-colors-ui-error-primary);
+      margin-top: 3px;
```
dgcohen, line 139: This might be a question for Apoorva, but do we want to use hardcoded pixel values in these styles or should we use design system spacing/REM values?

styles/components/MyAccount.module.scss
```diff
@@ -96,3 +119,30 @@
   font-weight: 510;
   font-style: normal !important;
 }
+
+.feesNotification {
+  margin-left: 0px !important;
```
dgcohen, line 124: Small nit/stylistic preference, but you don't need the px here.

styles/components/MyAccount.module.scss
dgcohen, line 1: Are these styling overrides here because the design deviates substantially from the DS? This might already be a conversation you're having, but if so we might want to introduce an alternative Notification or greater flexibility to the one in the DS.

pages/account/[[...index]].tsx
Comment (dgcohen) [reply pages/account/[[...index]].tsx:158]: Doesn't seem substantial enough to warrant it's own component in my opinion, what do others think?

styles/components/MyAccount.module.scss
7emansell: Yes definitely a conversation we want to have once Apoorva gets back– right now I'm just recreating the Figma as is as the starting point for that but the hope is that we can use design tokens/components as they are instead of customizing

pages/account/[[...index]].tsx
Comment (EdwinGuzman) [reply pages/account/[[...index]].tsx:158]: Agreed, this is okay. Also, this will get updated with the new `Banner` component which should be released by the end of March: https://nypl-design-system-git-dsd-1691-banner-nypl.vercel.app/?path=%2Fdocs%2Fcomponents-feedback-banner--docs

styles/components/MyAccount.module.scss
EdwinGuzman: Sorry, we should've mentioned that this would be the new `Banner` component. This recreation using the `Notification` component looks great but it will be replaced later. So don't worry about the hard-coded value for the banner/notification.

EdwinGuzman (approved):
This looks good. I was able to test this using my own account which has fees 😓 

The updated notification will be updated with the new `Banner` component but that is not ready yet.

Something I only just noticed in this PR but is also part of other tables is the empty values for certain rows/columns. I think accessibility-wise some sort of label should be added but Clare can give better guidance on that for all the tables -- each table may have a different "blank" value it should display that makes sense. So nothing you need to update now but pointing this out because it will probably need a fix later on.

This won't be easy to test until it's on QA because Clare would need to log in to test so let's not forget this once we're at that step.

styles/components/MyAccount.module.scss
7emansell: Gotcha. this and StatusBadge

dgcohen (approved):

pages/account/[[...index]].tsx
Comment (charmingduchess) [reply pages/account/[[...index]].tsx:158]: I think it can come out! especially because there's that big chunk of text, and it doesn't require a ton of props. It's relevant to the parent index page that there could be a notification displayed, but not necessarily what is rendered as a part of that notification.

styles/components/MyAccount.module.scss
charmingduchess: I would also be in favor of using DS values instead of hard coded custom values. @apoorvakhanolkar would it be okay to use the closest approximation of DS spacing values here instead?

7emansell: noting here that I do think this should be a broader conversation about the My account page designs since the decision on this will affect the styling on other components

PR #145, opened 2024-03-15
Bib level hold functionality

7emansell:
Adds handling for bib level holds, so circ items can be displayed correctly in the Requests tab.


src/models/MyAccount.ts
```diff
@@ -91,20 +99,29 @@ export default class MyAccount {
...
-      const nineTen = bibFields.varFields.find(
-        (field) => field.marcTag === "910"
-      )
-      // if we are unsure of the research ness of a bib, default to true so
-      // we don't let them renew or freeze the record
-      if (!nineTen) {
-        isResearch = true
-        isNyplOwned = false
+      // Bib level hold won't have varFields, and we know it will be circ and NYPL owned.
+      if (bibFields.varFields) {
```
charmingduchess, line 103: can this be refactored so we don't have a nested if statement? one approach to this could be to extract the 910 field parsing into its own method and return {isResearch, isNyplOwned} case by case.

src/models/MyAccount.ts
```diff
@@ -114,18 +131,25 @@ export default class MyAccount {
   buildHolds(holds: SierraHold[], bibData: SierraBibEntry[]): Hold[] {
     const bibDataMap = MyAccount.buildBibData(bibData)
     return holds.map((hold: SierraHold) => {
+      // Hold without bibIds is a bib level id.
+      const bibId = hold.record.bibIds ? hold.record.bibIds[0] : hold.record.id
```
charmingduchess, line 135: checking for bibIds to indicate if something is an item level hold works, but can you instead use `hold.recordType === 'i'` just to be little more explicit?

src/models/MyAccount.ts
```diff
@@ -114,18 +131,25 @@ export default class MyAccount {
...
         canFreeze: hold.canFreeze,
         frozen: hold.frozen,
         status: MyAccount.getHoldStatus(hold.status),
-        pickupLocation: hold.pickupLocation.name,
-        title: bibDataMap[hold.record.bibIds[0]].title,
-        isResearch: bibDataMap[hold.record.bibIds[0]].isResearch,
-        bibId: hold.record.bibIds[0],
-        isNyplOwned: bibDataMap[hold.record.bibIds[0]].isResearch,
+        pickupLocation: hold.pickupLocation,
+        title: bibDataMap[bibId].title,
```
charmingduchess, line 144: Assigning `bibDataMap[bibId]` to a variable here (ie `bibForHold`) could make it clearer what object we're looking at

charmingduchess (changes requested):
Great work!
Two smaller and one bigger thing I'd like to see changed.

src/models/MyAccount.ts
7emansell: What is `hold.recordType`? It doesn't seem to be on the holds that I'm looking at

charmingduchess: it's a string, in our case either `"i"` or `"b"`. I'm seeing it on every hold that comes back from the Sierra API.

src/models/MyAccount.ts
```diff
@@ -94,32 +94,42 @@ export default class MyAccount {
     return bibData
   }
 
+  /**
+   * getBibVarFields
+   * Reads varFields of a item-level hold's bib data to return if it is a research item,
+   * and if it is owned by NYPL.
+   */
+
+  static getBibVarFields(bibFields) {
```
charmingduchess, line 103: this snippet can be included here:
```
// Bib level hold won't have varFields, and we know it will be circ and NYPL owned.
      if (!bibFields.varFields) {
        isResearch = false
        isNyplOwned = true
      }
```
The pattern I'm going for here is pretty much like a switch - eg returning within each if statement. A little more verbose, but tends to be easier to read down the line. So the snippet above can be the first part of this method, and you can return the values straight away.

7emansell: This seems to be just recreating the nested if statement? I'm unclear what we're trying to avoid at this point

charmingduchess: this is what I'm getting at:
```
if no varfields
	  return isnypl: true isResearch: false

else if no 910 field
	  return isResearch: true isNypl: false (based on !isResearch from line 121)
	
else 
	  calulate isResearch, isNypl
	  return isResearch isNypl
```

otherwise you have these values being calculated in different places. also, as it is now, we have the info we need to return as soon as we know something doesn't have a 910 field. there's no need to couple that logic with the logic for records with 910 fields.

7emansell: Ahhhhh okay. i think i got it thank you

charmingduchess (approved):

dgcohen (approved):
Looks good to me!

src/models/modelTests/data/MyAccount.js
dgcohen, line 1: I think this type of thing should be added to __tests__/fixtures for consistency

src/models/MyAccount.ts
```diff
@@ -124,14 +148,14 @@ export default class MyAccount {
...
-        isResearch: bibDataMap[hold.record.bibIds[0]].isResearch,
-        bibId: hold.record.bibIds[0],
-        isNyplOwned: bibDataMap[hold.record.bibIds[0]].isNyplOwned,
-        catalogHref: bibDataMap[hold.record.bibIds[0]].isNyplOwned
-          ? bibDataMap[hold.record.bibIds[0]].isResearch
-            ? `https://nypl.org/research/research-catalog/bib/b${hold.record.bibIds[0]}`
-            : `https://nypl.na2.iiivega.com/search/card?recordId=${hold.record.bibIds[0]}`
+        title: bibForHold.title,
+        isResearch: bibForHold.isResearch,
+        bibId: bibId,
```
dgcohen, line 153: nbd but this can be shortened to just bibId

EdwinGuzman (approved):
The refactor and breaking out parts into smaller functions looks great.

PR #146, opened 2024-03-20
Noref rm date picker

charmingduchess:
- extract date input fields from advanced search page
- create useDateForm hook
  - the idea of this hook was to encapsulate as much as possible within the dateform component, including date range errors and invalid years
  - given DateForm props, the useDateForm hook returns the DateForm component and a validation method for the parent component to run on submit. 
  - the validation method checks for an error state, displays the error notification, and focuses on the first date input.
- replace date input fields with DateForm from hook in advanced search page and refine search dialog


pages/search/advanced.tsx
```diff
@@ -37,12 +36,12 @@ import type {
 import { getSearchQuery } from "../../src/utils/searchUtils"
 import initializePatronTokenAuth from "../../src/server/auth"
 import { appConfig } from "../../src/config/config"
+import DateForm from "../../src/components/DateForm"
```
EdwinGuzman, line 39: It looks like the `DateForm` from this file is not the one that is used within the `useDateForm` hook (it's also not used in this file). I think you should keep the component and hook in separate files, but move and rename `src/components/SearchFilters/DateForm.tsx` where the hook lives into `src/hooks/useDateForm.tsx`.

src/components/DateForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+  inputRef?: MutableRefObject<TextInputRefType>
+  dateAfter: string
+  dateBefore: string
+  debounceInterval: number
+  setDateRangeError: Dispatch<React.SetStateAction<boolean>>
+  changeHandler: (e: SyntheticEvent) => void
+  displayDateRangeErrorNotification: boolean
+}
+
+const DateForm = ({
```
EdwinGuzman, line 20: Not reviewing this yet because of my comment in `pages/search/advanced.tsx`. The component code here and in `SearchFilters/DateForm.tsx` have a few differences if you need to define the final place where it should go.

src/components/RefineSearch/RefineSearch.tsx
```diff
@@ -20,7 +21,18 @@ import {
   getQueryWithoutFilters,
 } from "../../utils/refineSearchUtils"
 import type { Aggregation } from "../../types/filterTypes"
+import DateForm from "../DateForm"
+import { useDateForm } from "../SearchFilters/DateForm"
 
+const debounceInterval = 60
```
EdwinGuzman, line 27: Consider making 20 or 60 a default in the hook so you don't always have to pass down a value.

src/components/SearchFilters/DateForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+interface DateFormHookPropsType {
+  inputRef?: MutableRefObject<TextInputRefType> // to focus on error
+  dateAfter: string
+  dateBefore: string
+  debounceInterval: number
+  changeHandler: (e: SyntheticEvent) => void
+}
+
+export const useDateForm = (dateFormProps: DateFormHookPropsType) => {
```
EdwinGuzman, line 23: Add a comment above to explain why and how it should be used, similar to the PR description.

src/components/DateForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+  inputRef,
+  dateAfter,
+  dateBefore,
+  debounceInterval,
+  changeHandler,
+  // these two params, though they sound similar are not
```
EdwinGuzman, line 26: Why not? That's good information to have as comments here.

src/components/SearchFilters/DateForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+        />
+        <TextInput
+          id="date-to"
+          labelText="End"
+          type="text"
+          name="dateBefore"
+          helperText="e.g. 2000"
+          value={dateBefore}
+          onChange={debounce((e) => changeHandler(e), debounceInterval)}
+          ref={inputRef}
```
EdwinGuzman, line 121: This works but I _think_ it should be two separate ref variables and props, one for each `TextInput`. The subtle issue might be when attempting to focus on the second input if there's an error there. I haven't tried this yet but will verify soon.

EdwinGuzman (changes requested):
This looks good. We talked about the rerendering and I think that's a good idea to document somewhere, maybe in the hook? Just to give a better lead in case issues come up later, e.g. we are aware this might rerender the text inputs on every app render, but given that it's a smaller app, we are not too concern about performance (yet?).

charmingduchess:
I'm thinking that the rerendering issue isn't about the hook, but about the fact that it's a controlled component. I'm seeing the same amount of rerenders on inputting text when I run the branch that uses the fieldsetdate.

pages/search/advanced.tsx
charmingduchess: done!

src/components/RefineSearch/RefineSearch.tsx
charmingduchess: done!

src/components/DateForm.tsx
charmingduchess: updated comment

src/components/SearchFilters/DateForm.tsx
charmingduchess: This component wasn't being used. I'm currently passing in two separate refs.

src/components/SearchFilters/DateForm.tsx
charmingduchess: done

src/components/RefineSearch/RefineSearch.tsx
```diff
@@ -78,10 +77,11 @@ const RefineSearch = ({
         )
       } else return null
     })
-    .concat(dateFieldset)
+    .concat(<DateForm {...dateFormProps} />)
 
   const handleSubmit = (e: SyntheticEvent) => {
     e.preventDefault()
+    if (validateDateRange() === false) return
```
EdwinGuzman, line 84: nitpick: you can do `!validateDateRange()` similar to how its set in `/pages/search/advanced.tsx`.

src/components/RefineSearch/RefineSearch.tsx
```diff
@@ -36,32 +45,22 @@ const RefineSearch = ({
...
+      const target = e.target as HTMLInputElement
+      // update the parent state to know about the updated dateValues
+      setAppliedFilters((prevFilters) => {
+        return {
+          ...prevFilters,
+          [target.name]: [target.value],
+        }
+      })
+    },
+    inputRefs: [dateInputRef],
```
EdwinGuzman, line 60: Why is only one set?

src/hooks/useDateForm.tsx
```diff
@@ -0,0 +1,84 @@
...
+const rangeContainsInvalidYearFormat = (dateAfter, dateBefore) => {
+  // if there is no input for date, it's valid. if there is, it must
+  // be at least 4 digits.
+  const isDateFourDigits = (date: string) => {
+    const dividedBy1000 = Math.floor(parseInt(date, 10) / 1000)
+    return dividedBy1000 > 0 && dividedBy1000 <= 1
+  }
+  const dateAfterValid = !dateAfter || isDateFourDigits(dateAfter)
+  const dateBeforeValid = !dateBefore || isDateFourDigits
+  // ^^
```
EdwinGuzman, line 82: What's this for?

src/hooks/useDateForm.tsx
```diff
@@ -0,0 +1,84 @@
...
+ * immediately.
+ * 2. props to pass into the DateForm component. Including original props, plus
+ * methods and values used within the DateForm component to manage error state
+ * and display an error message
+ *
+ */
+
+export const useDateForm = (dateFormProps: DateFormHookPropsType) => {
+  const { dateBefore, dateAfter } = dateFormProps
+  // const [dateRangeError, setDateRangeError] = useState("")
```
EdwinGuzman, line 28: If you don't need this you can remove it.

EdwinGuzman (commented):
The update looks good, just added some minor comments. I saw you have a `.todo` test and `DateForm` and `useDateForm` should have tests as well. That can be a separate ticket/PR if you want but it should be done before this gets deployed to prod.

src/components/RefineSearch/RefineSearch.tsx
charmingduchess: updated

src/components/RefineSearch/RefineSearch.tsx
charmingduchess: Hmm I was thinking there was different behavior from RefineSearch to Advanced search, but I'm seeing now not only is it the same, but the focus does automatically go to the first field, even when there is only one ref. Maybe I will switch it back to only using one ref across both components.

charmingduchess: I'm seeing some unexpected behavior around this. I'd love to run it by you on monday.

src/hooks/useDateForm.tsx
charmingduchess: it was leftover from an older comment!

src/hooks/useDateForm.tsx
charmingduchess: removed

charmingduchess:
I tried writing tests and got stuck because of the issue with the user input simulation taking too long during the tests. I will get some together for the useDateForm, but I'm not sure what i can test for the actual component without any user simulated typing.

EdwinGuzman:
> I'm seeing some unexpected behavior around this. I'd love to run it by you on monday.

Something I'm seeing that _might_ be related is if you type fast, there's a delay. It might be the debounce and not the ref but something else to look at.

charmingduchess:
I removed the debounce from the date form and updated so both times its used there are two unique input refs.

pages/search/advanced.tsx
```diff
@@ -114,6 +114,7 @@ export default function AdvancedSearch({ isAuthenticated }) {
   const handleClear = (e: SyntheticEvent) => {
     e.preventDefault()
     alert && setAlert(false)
+    dateInputRefs.map((ref) => (ref.current.value = ""))
```
EdwinGuzman, line 117: This doesn't have to be part of this PR, but what if `useDateForm` exposes an additional "clear" function that internally calls `inputRefs.forEach((ref) => (ref.current.value = ""))`? That way the function can be called instead of manually doing it here.

Using `forEach` instead of `map` since `map` generates a new array and we don't need that.

src/hooks/useDateForm.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+    })
+    it("validateDateRange returns false when end date is earlier than start date", async () => {
+      const { validateDateRange, dateFormProps } = setUpDateFormWithRefs(
+        "2000",
+        "1999"
+      )
+      render(<DateForm {...dateFormProps} />)
+      expect(validateDateRange()).toBe(false)
+    })
+    it.todo(
```
EdwinGuzman, line 58: The hook works but this is hard to test. I tried this and the asynchronous nature of `setDisplayDateRangeError` is not updating the value for `displayDateRangeError` right away. That means that the returned `dateFormWithHookProps` object does not have the latest error message. I was testing to see if i could call `setUpDateFormWithRefs` to get the "updated" value for `dateFormProps` but this doesn't work and isn't correct.

The way the hook works, it sort of depends on the `DateForm` component to rerender and get the updated value. I think you can ignore/remove this test and instead add a test to verify the error message on the screen for the `DateForm` component.

src/hooks/useDateForm.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+  rangeContainsInvalidYearFormat,
+  useDateForm,
+} from "./useDateForm"
+import { renderHook, screen, render } from "../utils/testUtils"
+import type { MutableRefObject } from "react"
+import { useRef } from "react"
+import type { TextInputRefType } from "@nypl/design-system-react-components"
+import DateForm from "../components/SearchFilters/DateForm"
+
+const setUpDateFormWithRefs = (start, end) => {
```
EdwinGuzman, line 12: Very nice pattern to reuse and extract values from the invoked hook.

EdwinGuzman (approved):
The way this is now works fine, although you should add more tests for `DateForm` eventually.

I think we said we did not want to focus on the second input if there's an error right? If the user inputs "1990" for Start and "asdf" for End, the focus will go to the first input even though the error is on the second input.

Not directly related to this PR but I noticed if there's an error or no search results, then the "Refine Search" button and tags do not render. In the following example, I searched for "cats" and low start and end dates (I didn't copy them down but they were around "1000" and "1100"). I now don't have a way to "reset" or remove the filters that returned this empty search results. Seems like a new ticket?

Screen Shot 2024-03-26 at 9 28 32 AM

charmingduchess:
I was intending to focus on the first input. I don't feel strongly either way about what should be happening, though. Let me know if you think it should change.

The issue of there being no escape from an empty search is expected behavior. Which is to say, it's happening in DFE. There would need to be some input from product about what we want to happen. Right now, we don't have a way to remove specific filters from the applied filters tag set when there are no aggregations returned (which is what happens when there are zero results). We don't get labels back from no aggregations, so we can't display meaningful labels for the tagset to display. All we can do given the discovery api response right now is to display a Clear All Filters. 
I think this is worth making a bigger change that involves updating the api call to request aggregations for the base query instead of aggregations for the query+filters.

EdwinGuzman:
Thanks, that's a great explanation. Let's leave this is behavior as it exists -- I agree this needs product decision and we are supposed to replicate existing functionality (even if it's not clear/ideal). We are now, however, in a better place to make updates and fixes.

Let's leave the bigger change for when we get input from product regarding UX and expected behaviors.

For the focusing on the first input, let's also leave that.

src/components/SearchFilters/DateForm.tsx
```diff
@@ -0,0 +1,70 @@
...
+        id="date-fieldset"
+        gridTemplateColumns="repeat(2, minmax(0, 1fr))"
+        legendText="Date"
+        display="grid"
+        gap="s"
+      >
+        <TextInput
+          id="date-from"
+          labelText="Start"
+          type="text"
```
dgcohen, line 48: Is type=text necessary for TextInput?

dgcohen (approved):
Looks good, just one minor comment. Would also like to see the ref clearing moved to its own function as Edwin suggested in his comment.

pages/search/advanced.tsx
charmingduchess: i included this update! thanks for the recommendation.

src/components/SearchFilters/DateForm.tsx
charmingduchess: nope i will remove

pages/search/advanced.tsx
EdwinGuzman: Looks good.

EdwinGuzman (approved):

PR #147, opened 2024-03-20
Scc 4027/adv search display

charmingduchess:
Includes advanced search params to "Displaying 1-50 results..." string so users know what advanced search params have been applied to their search. Advanced search options that are filters are not included in this, but they do show up in the applied filters tagset.


dgcohen (approved):
Looks great!

PR #149, opened 2024-03-22
Scc 3935/holds tab errors

charmingduchess:
UPDATE 4/19/24:
Refactored MyAccount model to be more of an account data client than souped up data structure. This was to facilitate implementing a Promise.allSettled so one failed request wouldn't break the whole account page. It also has the added benefit of making mocking simpler since the actual api client can be totally mocked without using jest shenanigans. I have not been able to run due to sierra test issues, but I'm confident once it's running any kinks will be easy to eliminate.

original PR comment:
Is this totally overkill? I recognize that this might be a little too much too late, and it is mostly for debugging. Our orignal error handling for My Account was definitely too broad, and was mostly put in place to make sure that _anything_ erroring during the patron building process wouldn't crash the app. I think this might be useful moving forward once we move to phase two. Let me know thoughts on this!
Basically what I did here is group fetches and build calls and log specifically where errors were coming from. I was not able to get the `error.lineNumber` or `fileName` to log anything but `undefined`.


src/models/MyAccount.ts
```diff
@@ -17,6 +17,13 @@ import type {
 
 let client
 
+class MyAccountModelError extends Error {
+  constructor(errorDetail, error) {
+    super()
+    this.message = `Error ${errorDetail} in ${error.filename}, line ${error.lineNumber}: ${error.message}`
```
EdwinGuzman, line 23: I think you're getting undefined for filename and lineNumber because they're not standard so it may not work on all browsers or consistently https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/lineNumber

src/models/MyAccount.ts
```diff
@@ -106,85 +113,111 @@ export default class MyAccount {
...
-        catalogHref: bibDataMap[hold.record.bibIds[0]].isNyplOwned
-          ? bibDataMap[hold.record.bibIds[0]].isResearch
-            ? `https://nypl.org/research/research-catalog/bib/b${hold.record.bibIds[0]}`
-            : `https://nypl.na2.iiivega.com/search/card?recordId=${hold.record.bibIds[0]}`
-          : null,
-      }
-    })
+    let bibDataMap
+    try {
+      bibDataMap = MyAccount.buildBibData(bibData)
```
EdwinGuzman, line 118: `buildBibData` calls a reduce function but I think the default is an empty array. So here I don't expect `buildBibData` to throw an error, no?

EdwinGuzman (commented):
I was thinking of this as trying to only catch fetch request errors. Looking it over again, you're also trying to catch any data issues because the data is not always consistent, right? It does seem like too much but you've already been encountering and updating the model based on new test accounts. So in that sense this makes update is good, but I'm not sure you will get the trace information you want unless you update the `MyAccountmodelError` and `SierraClientError` classes.

src/models/MyAccount.ts
charmingduchess: ah ok. i will remove.

src/models/MyAccount.ts
charmingduchess: I'm worried about certain responses not having varFields. If you don't specify it in the url, you don't get varFields back.

src/models/MyAccount.ts
```diff
@@ -227,15 +260,25 @@ export default class MyAccount {
...
-  )
-  const holdBibData = await MyAccount.fetchBibData(holds.entries, "record")
+  let holds, patron, fines, checkoutBibData, checkouts, holdBibData
+  try {
+    checkouts = await MyAccount.fetchCheckouts(baseQuery)
+    holds = await MyAccount.fetchHolds(baseQuery)
+    patron = await MyAccount.fetchPatron(baseQuery)
+    fines = await MyAccount.fetchFines(baseQuery)
+  } catch (e) {
+    throw new MyAccountModelError("fetching patron data", e)
```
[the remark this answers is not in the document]
charmingduchess: A Sierra client error wouldn't be thrown on a fetch error. There might be an error thrown by the sierra wrapper node module, but it will be caught in the My Account fetchers and bubble up as a MyAccountError. Sierra client only has one method and only two errors that we're throwing. SieraClientErrors only occur on instantiation.

charmingduchess:
It's less that the data isn't consistent, more that there is a lot of uncertainty. The data has a lot of properties and it's not always immediately clear which ones we can expect. Our types are based on what we copied out of the API responses, but for example, we only had item holds when we made the types originally, so then when we tested with bib holds there were mysterious errors that were hard to trace. I'm not sure what other updates I could add to make the tracing more explicit.

src/models/MyAccount.ts
EdwinGuzman: That's fair. If you throw an error, the entire page will crash. What if you can resolve a fallback for when responses don't have varFields so that the page can still render but not display that specific data?

src/models/MyAccount.ts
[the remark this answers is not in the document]
EdwinGuzman: This makes more sense, thanks.

src/models/MyAccount.ts
EdwinGuzman: At the top level, `MyAccoutFactory` is called within a try/catch and there are defaults, or rather undefined values being returned. So taking a better look, this should crash. 

But what if all the fetch calls succeed but there's an error with the data manipulation once `new MyAccount` is called. In that case, will none of the data get returned?

charmingduchess: That's a good point. Even without this work, if there are any errors thrown here, then no data is returned.

EdwinGuzman:
@charmingduchess I think you have some updates to make in this PR. Otherwise, if you want to move ahead with this, we should.

EdwinGuzman:
This is a bit hard to review because some of the files changed are already in `main`. If you can, update the base branch with the latest from `main`... but I'm worried about the other open PRs and the order they should be merged in.

I can also just review the commits.

src/models/MyAccount.ts
```diff
@@ -260,26 +296,45 @@ export default class MyAccount {
...
+    patronFetcher.getPatron(),
+    patronFetcher.getFines(),
+  ])
+  const [checkouts, holds, patron, fines] = sierraData.map((data) => {
+    return data.status === "fulfilled" ? data.value : []
+  }) as [Checkout[], Hold[], Patron, Fine]
+  return { checkouts, holds, patron, fines }
+}
+
+export const getPickupLocations = async () => {
```
EdwinGuzman, line 313: Where is this being used or is a stub?

src/models/modelTests/MyAccount.test.ts
```diff
@@ -54,34 +67,184 @@ describe("MyAccountModel", () => {
...
-      MyAccount.fetchHolds = async () => holds
-      MyAccount.fetchPatron = async () => patron
-      MyAccount.fetchFines = async () => fines
-      MyAccount.fetchBibData = async (entries, recordType) => {
-        if (recordType === "item") {
-          return checkoutBibs
-        } else return holdBibs
-      }
-      const account = await MyAccountFactory("12345")
+      MyAccount.prototype.fetchCheckouts = async () =>
```
EdwinGuzman, line 220: Does mocking the functions directly, without using the prototype, work? Just curious because the functions are assigned within the class and not as prototype functions but you can still mock it this way.

src/models/modelTests/MyAccount.test.ts
```diff
@@ -54,34 +67,184 @@ describe("MyAccountModel", () => {
...
+        Promise.resolve(checkouts)
+      MyAccount.prototype.fetchHolds = async () => Promise.resolve(holds)
+      MyAccount.prototype.fetchPatron = async () => Promise.resolve(patron)
+      MyAccount.prototype.fetchFines = async () => Promise.resolve(fines)
+      MyAccount.prototype.fetchBibData = jest
+        .fn()
+        .mockResolvedValueOnce(checkoutBibs)
+        .mockResolvedValueOnce(holdBibs)
+
+      // @ts-ignore
```
EdwinGuzman, line 230: What error are you ignoring? "12345" being a string and not a type of the client?

src/models/modelTests/MyAccount.test.ts
```diff
@@ -156,77 +322,67 @@ describe("MyAccountModel", () => {
...
-        MyAccount.getResearchAndOwnership({
-          varFields: [
-            { marcTag: "910", subfields: [{ tag: "a", content: "RLOTF" }] },
-          ],
-        })
-      ).toStrictEqual({
-        isResearch: true,
-        isNyplOwned: false,
-      })
+    it("will return other processed data if one fetch fails", async () => {
```
EdwinGuzman, line 353: Nice.

src/models/MyAccount.ts
charmingduchess: this was from account settings. its not being used because i havent decidded yet what's happening with the pickup locations fetching. but the method will be used at some point in the future. currently im just using static data for pickup locations.

src/models/modelTests/MyAccount.test.ts
charmingduchess: I generally mock instance methods with .prototype. when i remove it, i also get type errors.

src/models/modelTests/MyAccount.test.ts
charmingduchess: addressed this in the comments

EdwinGuzman (approved):

PR #157, opened 2024-04-02
SCC-4062 - Replace hardcoded vega links with config var

dgcohen:
Ticket:
https://jira.nypl.org/browse/SCC-4062

In anticipation of the changing of the vega url from https://nypl.na2.iiivega.com to borrow.nypl.org, I'm opening a small PR that replaces any hardcoded instance of the old url with the value stored in appConfig to easily switch it over.

All instances of the url in DFE are currently coming from an Env var so it should just be a matter of changing it and redeploying there when it's time to switch.

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -65,7 +65,7 @@ export const appConfig: AppConfig = {
   urls: {
     drbAbout:
       "https://digital-research-books-beta.nypl.org/about?source=catalog",
-    circulatingCatalog: "https://nypl.na2.iiivega.com",
+    circulatingCatalog: "https://borrow.nypl.org",
     legacyCatalog: "https://legacycatalog.nypl.org/",
     login: "https://login.nypl.org/auth/login",
     locations: "https://www.nypl.org/locations/",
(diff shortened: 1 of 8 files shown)


src/config/config.ts
```diff
@@ -65,7 +65,7 @@ export const appConfig: AppConfig = {
   urls: {
     drbAbout:
       "https://digital-research-books-beta.nypl.org/about?source=catalog",
-    circulatingCatalog: "https://nypl.na2.iiivega.com/",
+    circulatingCatalog: "https://nypl.na2.iiivega.com",
```
EdwinGuzman, line 68: This should be borrow.nypl.org.

EdwinGuzman (commented):
The update looks good but point to borrow.nypl.org

7emansell (approved):

src/server/api/bib.ts
```diff
@@ -61,7 +61,9 @@ export async function fetchBib(
       if (sierraBibResponse.statusCode === 200) {
         return {
           status: 307,
-          redirectUrl: `${appConfig.urls.circulatingCatalog}/iii/encore/record/C__R${id}`,
+          redirectUrl: `${
+            appConfig.urls.circulatingCatalog
+          }/search/card?recordId=${id.replace(/^b/, "")}`,
```
EdwinGuzman, line 66: For an update like this, make sure to point it out to Chris when he QAs.

EdwinGuzman (approved):

nonword (approved):

PR #160, opened 2024-04-03
SCC-3929: Password/PIN update modal

7emansell:
Resolves [SCC-3929](https://jira.nypl.org/browse/SCC-3929).

Adds password/PIN update modal to account settings tab. Button will be moved into the overall account settings form ([SCC-3928](https://jira.nypl.org/browse/SCC-3928)). 

Noting that majority of the modal styles are just placeholders for now.


src/components/MyAccount/SettingsTab/PasswordModal.tsx
```diff
@@ -0,0 +1,136 @@
...
+          <Text sx={{ marginBottom: 0 }}> PIN/PASSWORD change failed </Text>
+        </>
+      </Heading>
+    ),
+    onClose: resetModal,
+  })
+
+  return (
+    <>
+      <Button
```
7emansell, line 121: This is a `Link` in the designs, but an empty `href` forces an inconvenient re-render once the modal is open. Style of the button is to make it sort of look like a link within DS constraints, but it should probably just embrace its buttonhood

pages/account/[[...index]].tsx
```diff
@@ -198,27 +198,27 @@ export async function getServerSideProps({ req }) {
   // Parsing path from url to pass to ProfileTabs.
   const tabsPath = req.url.split("/", -1)[2] || null
   const id = patronTokenResponse.decodedPatron.sub
-  // try {
-  const { checkouts, holds, patron, fines } = await MyAccountFactory(id)
-  // Redirecting /fines if user has none.
-  if (tabsPath === "overdues" && fines.total === 0) {
+  try {
```
Comment (7emansell) [inline pages/account/[[...index]].tsx:201]: Not related to this PR, just commented out this try/catch accidentally in another

src/components/MyAccount/SettingsTab/PasswordModal.tsx
```diff
@@ -0,0 +1,136 @@
...
+        id="button"
+        buttonType="text"
+        onClick={() => {
```
EdwinGuzman, line 124: Since no argument is passed, you can also do `onClick={openModal}`

src/components/MyAccount/SettingsTab/PasswordModal.tsx
EdwinGuzman: It definitely should be a button. Links for navigation and buttons for user interaction. I think we (@bigfishdesign13) could make the "text" variant look more like text or a link. Right now, it the padding around the text content gets in the way (I noticed this issue elsewhere).

src/components/MyAccount/SettingsTab/PasswordChangeForm.tsx
```diff
@@ -0,0 +1,110 @@
...
+      formData.newPassword !== "" &&
+      formData.confirmPassword !== "" &&
+      formData.passwordsMatch
+    )
+  }
+
+  const handleInputChange = (e) => {
+    const { id, value } = e.target
+    if (id === "confirmPassword") {
+      formData.passwordsMatch = formData.newPassword === value
```
EdwinGuzman, line 38: State should not be updated directly like this. Include this as part of the `setFormData` function on line 40.

src/components/MyAccount/SettingsTab/PasswordChangeForm.tsx
```diff
@@ -0,0 +1,110 @@
...
+  updateModal: (errorMessage?: string) => void
+}) => {
+  const [formData, setFormData] = useState({
+    oldPassword: "",
+    newPassword: "",
+    confirmPassword: "",
+    passwordsMatch: true,
+  })
+
+  const validateForm = () => {
```
EdwinGuzman, line 26: Because of how the component re-renders, you can make this a boolean instead of a function and it will work just the same.

src/components/MyAccount/SettingsTab/PasswordChangeForm.tsx
```diff
@@ -0,0 +1,110 @@
...
+        barcode: patron.barcode,
+      }),
+    })
+    const errorMessage = await res.json()
+    res.status === 200 ? updateModal() : updateModal(errorMessage)
+  }
+
+  return (
+    <Form id="pw-form" gap="grid.s">
+      <FormField className={styles.formField}>
```
EdwinGuzman, line 61: It doesn't seem like there's a `.formField` css class in `MyAccount.module.scss`.

src/components/MyAccount/SettingsTab/PasswordChangeForm.tsx
```diff
@@ -0,0 +1,110 @@
...
+      formData.oldPassword !== "" &&
+      formData.newPassword !== "" &&
+      formData.confirmPassword !== "" &&
+      formData.passwordsMatch
+    )
+  }
+
+  const handleInputChange = (e) => {
+    const { id, value } = e.target
+    if (id === "confirmPassword") {
```
EdwinGuzman, line 37: This works well but you're missing something here. I'll run through the bug I found but this is where you should fix it.

- type in the old password/pin
- type in the new password/pin as 1234
- type in the confirmation password/pin as 12345
- they don't match. instead of updating the confirmation, what if a user goes back and updates the initial new password/pin field?
- type in the initial password/pin as 12345
- the error persists

src/components/MyAccount/SettingsTab/PasswordChangeForm.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const validateForm = () => {
+    return (
+      formData.oldPassword !== "" &&
+      formData.newPassword !== "" &&
+      formData.confirmPassword !== "" &&
+      formData.passwordsMatch
+    )
+  }
+
+  const handleInputChange = (e) => {
```
EdwinGuzman, line 35: A more universal way of getting HTML input values is by using the `name` attribute. What you're doing here checking against the `id` works but use `name` instead. This also means you should add `name="oldPassword"` to the `TextInput` components (but update the values accordingly).

EdwinGuzman (changes requested):
Works well but the form needs some updates.

src/components/MyAccount/SettingsTab/PasswordChangeForm.tsx
7emansell: Is it okay that the confirm password shows up as incorrect while the new password is being entered originally?

EdwinGuzman: Yes, I think so.

src/components/MyAccount/SettingsTab/PasswordChangeForm.tsx
EdwinGuzman: @7emansell just need this included for the updates.

EdwinGuzman (approved):

charmingduchess (approved):
Looks great! Nice job hackin' the modal.

7emansell:
will wait to merge this in so that vera can incorporate into account settings form

PR #161, opened 2024-04-04
SCC-4065: Search Results VQA pt1

EdwinGuzman:
Resolves JIRA ticket [SCC-4065](https://jira.nypl.org/browse/SCC-4065)

Fixes VQA issues on the top half of the Search Results page, specifically for the headings in this [doc](https://docs.google.com/document/d/1qgHzrubMwxx_UvVu_hfVhDfbPEPPbgrypHv3BohSPtI/edit#heading=h.ia82i1lm88dj):
- Search Bar
- Results Heading and Sort Dropdown
- Search Results Card - Single Item

### DIFF
--- a/src/components/ItemTable/ItemTable.tsx
+++ b/src/components/ItemTable/ItemTable.tsx
@@ -17,21 +17,25 @@ const ItemTable = ({ itemTableData }: ItemTableProps) => {
       <Table
         columnHeaders={itemTableData.tableHeadings}
         tableData={itemTableData.tableData}
+        mt="s"
         mb="s"
-        // TODO: These styles approximate those of the old app.
-        // Design will VQA the component with the DS defaults, but leaving this for reference.
-        // sx={{
-        //   tableLayout: "fixed",
-        //   width: "full",
-        //   tr: { border: "0 !important", padding: 0 },
-        //   "th, td, th > span, td > span": {
-        //     paddingTop: "xs",
-        //     paddingBottom: "xs",
-        //   },
-        // }}
+        sx={{
+          width: "unset",
+          thead: { mb: "xs" },
+          tr: { border: "0 !important" },
+          "th, td": {
+            width: "240px",
+            padding: "0",
+            marginRight: "32px",
+          },
+          "th, td, th > span, td > span": {
+            paddingTop: "0",
+            paddingBottom: "0",
+          },
+        }}
       />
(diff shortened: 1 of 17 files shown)


src/components/ItemTable/ItemTable.tsx
```diff
@@ -17,21 +17,25 @@ const ItemTable = ({ itemTableData }: ItemTableProps) => {
...
-        //   },
-        // }}
+        sx={{
+          width: "unset",
+          thead: { mb: "xs" },
+          tr: { border: "0 !important" },
+          "th, td": {
+            width: "240px",
+            padding: "0",
+            marginRight: "32px",
```
charmingduchess, line 29: We decided a while back if there's more than one line of inline styles to move it to a CSS module. Can you put this in another file?

src/components/ItemTable/ItemTable.tsx
```diff
@@ -17,21 +17,25 @@ const ItemTable = ({ itemTableData }: ItemTableProps) => {
...
-        // sx={{
-        //   tableLayout: "fixed",
-        //   width: "full",
-        //   tr: { border: "0 !important", padding: 0 },
-        //   "th, td, th > span, td > span": {
-        //     paddingTop: "xs",
-        //     paddingBottom: "xs",
-        //   },
-        // }}
+        sx={{
```
charmingduchess, line 22: We decided a while back if there's more than one line of inline styles to move it to a CSS module. Can you put this in another file? Also, can you replace 32px with nypl-space-l token?

charmingduchess (changes requested):
Looks good. I know that you're updating things that were already written, but I'm concerned about there being styling information in different places. I think if we are using the `sx` prop or CSS modules for a component, we should not _also_ be using `mb`, `maxH`, etc props. There was one specific place where I mentioned the `sx` prop looking like it could come out to a css module, and that component also is using the prop stylers. 
Beyond the great debate of inline or css module styling, I think the most important thing is that all of the styles are in one place. I think we should avoid using those prop stylers if there are styles happening elsewhere, including the `sx` prop. 
@dgcohen I'm curious your thoughts on this too.

src/components/ItemTable/ItemTable.tsx
EdwinGuzman: Updated.

charmingduchess (approved):

PR #162, opened 2024-04-05
SCC-4024: Account redirects

7emansell:
Resolves [SCC-4024](https://jira.nypl.org/browse/SCC-4024).

Redirects invalid URL paths on `/account` back to `/account`. Additionally, it adds the `domain` environment variable to fix the `getLoginRedirect` error that has recently materialized and corrects the `/checkouts` path to `/items` to match current prod.


__test__/pages/account/account.test.tsx
```diff
@@ -1,23 +1,100 @@
...
+      patron: mockPatron,
+      fines: mockFines,
+      holds: mockHolds,
+    })
+
+    const mockReq = {
+      headers: {
+        host: "local.nypl.org:8080",
+      },
+      url: "/account/spaghetti",
```
charmingduchess, line 67: 🍝

pages/account/[[...index]].tsx
```diff
@@ -198,27 +196,32 @@ export async function getServerSideProps({ req }) {
...
-  // try {
-  const { checkouts, holds, patron, fines } = await MyAccountFactory(id)
-  // Redirecting /fines if user has none.
-  if (tabsPath === "overdues" && fines.total === 0) {
+  try {
+    const { checkouts, holds, patron, fines } = await MyAccountFactory(id)
+    // Redirecting /fines if user has none, or any other paths.
+    if (
+      (tabsPath === "overdues" && fines.total === 0) ||
+      ["checkouts", "requests", "overdues", "settings", null].indexOf(
```
Comment (charmingduchess) [inline pages/account/[[...index]].tsx:204]: what is the null value for?

charmingduchess (changes requested):
Looks good. I am encountering a couple bugs:

- Neither appending overdues to any account url (including /account), nor inputting this url `http://local.nypl.org:8080/research/research-catalog/account/overdues` displays the fees tab. 
- This url `http://local.nypl.org:8080/research/research-catalog/account/settings/spaghetti` brings me to settings but leaves `spaghetti` at the end. I think we should remove anything that comes after tabs we recognize.

7emansell:
> * Neither appending overdues to any account url (including /account), nor inputting this url `http://local.nypl.org:8080/research/research-catalog/account/overdues` displays the fees tab.
> * This url `http://local.nypl.org:8080/research/research-catalog/account/settings/spaghetti` brings me to settings but leaves `spaghetti` at the end. I think we should remove anything that comes after tabs we recognize.

good catches these made me realize this is not working even 50% of the way i expected so back to WIP

charmingduchess:
I just realized that the url for the checkouts tab in prod is actually /items, not /checkouts. Can you update this PR to have that url instead?

EdwinGuzman:
>  how slow this is

What do you mean? Where do you see the slowness?

7emansell:
@EdwinGuzman try a url like `/account/settings/spaghetti`– for me testing locally, it takes a full second of loading for the page to redirect correctly

EdwinGuzman (commented):
This looks good. I think the logic is a bit confusing and I was wondering why you were initially using the `.some` along with `startsWith` but it makes sense after going through it and testing it. I was thinking you could use the array `.includes` function but it'll only get you so far...

As for optimization, I think this is a good example of first making it work, and then optimizing later. Honestly, the redirect lag is not terrible and doesn't take too long.

The only question I have before approving is that the ticket says a bad URL should redirect to the 404 page but here it's redirecting the the base `/account` page. This is fine but just want to confirm this is okay in case it was confirmed but the ticket not updated.

7emansell:
@EdwinGuzman let me know if you have any ideas about how to arrange the code more logically, I definitely see what you mean about the readability but maybe have just been looking at it too long . I'll update the ticket to match since this is the correct behavior for what would otherwise be 404s (checked in w Sean before).

charmingduchess:
I have a feeling the lag is coming from loading all the patron data before the redirect is processed, then redirecting and fetching the data again. Try moving the redirect logic before the data fetch so the data is only fetched once we know we're on the correct url.

pages/account/[[...index]].tsx
```diff
@@ -196,29 +194,52 @@ export async function getServerSideProps({ req }) {
     }
   }
   // Parsing path from url to pass to ProfileTabs.
-  const tabsPath = req.url.split("/", -1)[2] || null
+  const tabsPath = req.url.split("/").slice(2).join("/") || null
```
Comment (charmingduchess) [inline pages/account/[[...index]].tsx:197]: I wonder if a regex would make this less brittle and demonstrate more clearly what we're extracting from the url.

EdwinGuzman:
> I have a feeling the lag is coming from loading all the patron data before the redirect is processed, then redirecting and fetching the data again. Try moving the redirect logic before the data fetch so the data is only fetched once we know we're on the correct url.

I agree, this is a good suggestion. 

-----

The following is just an idea to try on a different branch. For this PR, I think the above will definitely help out.

When I was debugging, I was wondering if you can allow Next to follow through with a route and instead of redirecting remove the bad part of the url. For example:
- you go to `/account/settings/spaghetti`
- `getServerSideProps` processes this as `/account/settings` to fetch the necessary data and load the correct tab
- Next's router is used to remove `/spaghetti` from the URL path _without_ triggering a route change.

The "without triggering a route change" might not be possible but something you can look into. I think Next processes all new path updates as changes and will trigger a response if done on the server through `getServerSideProps`. It might be possible to do this on the client-side and use the `useRouter` hook to update/replace the URL without triggering a call to the history. If you try this, it might mean removing the redirect logic on the server and moving the logic to the client instead, which may not be a good trade-off.

pages/account/[[...index]].tsx
Comment (7emansell) [reply pages/account/[[...index]].tsx:197]:
```
const tabsPathRegex = /\/(?:[^\/]+\/){2}(.+)/
const match = req.url.match(tabsPathRegex)
const tabsPath = match ? match[1] : null
```
is something like this what you were thinking? to me it's kind of not clearer

7emansell:
> I have a feeling the lag is coming from loading all the patron data before the redirect is processed, then redirecting and fetching the data again. Try moving the redirect logic before the data fetch so the data is only fetched once we know we're on the correct url.

unfortunately I can't rearrange like that bc whether the user has fines or not changes the allowed urls... I'm not sure, would it be worth it to just fetch the fines and then everything else? cc @charmingduchess @EdwinGuzman

pages/account/[[...index]].tsx
Comment (charmingduchess) [reply pages/account/[[...index]].tsx:197]: i guess I was imagining something with some part of the url explicitly in it like
`/\/account\/(.+)\/*/`
my thinking is seeing the capture group next to what we expect the url to be is clearer. i also don't imagine this would happen, but this regex is a little more flexible for any kind of domain change since it's reliant on the /account part of the string. as i see it, that's the only thing that's guaranteed in the url for this page. but also, that's not a serious concern. 
if you think it's still not clearer, its fine to leave as is. i just thought it was worth exploring.

charmingduchess:
@7emansell i made some big changes to how the data fetching works, and i think my refactor will facilitate this kind of flexibility. why don't you leave it as is, and make a ticket to investigate a preliminary fines  fetch and moving the other fetches to after the redirect.

EdwinGuzman:
> make a ticket to investigate a preliminary fines fetch and moving the other fetches to after the redirect.

Yes, and it should be a spike to investigate alternatives. Will approve now but updating the regex is up to you.

EdwinGuzman (approved):

7emansell:
This [ticket](https://jira.nypl.org/browse/SCC-4099) has been created to further investigate redirects, but they work as of now

charmingduchess:
I know it's not actually related to this work, but if it's not too much of a lift, can you do something about setImmediate situation that is causing the tests to fail before you merge this in?

charmingduchess:
Can you make a ticket to deal with the setImmediate error? I want to get that addressed asap.

charmingduchess (approved):

PR #163, opened 2024-04-08
SCC-4064 - Search results Accessibility QA Pt. 2

dgcohen:
This PR includes fixes for a few accessibility issues on Search Result:

1. Adding lang code to DRB titles (note that this currently isn't possible on main Search Results because lang isn't sent from Discovery API in iso_2 format)
2. Adding correct attributes to disabled links (I extended the RC Link component but will investigate adding these to the design system)
3. Adding a CardHeading to DRB cards
4. Adding aria labels to ambiguous links such as Request buttons and DRB callouts (adds the Bib title per Clare's request).

### DIFF
--- a/src/components/DRB/DRBCard.tsx
+++ b/src/components/DRB/DRBCard.tsx
@@ -55,6 +55,7 @@ const DRBCard = ({ drbResult }: DRBCardProps) => {
         {drbResult.readOnlineUrl && (
           <DSLink
             href={drbResult.readOnlineUrl}
+            aria-label={`Read Online, ${drbResult.title}`}
             target="_blank"
             type="buttonPrimary"
             mb={drbResult.readOnlineUrl ? "s" : ""}
--- a/src/components/DRB/DRBCard.tsx
+++ b/src/components/DRB/DRBCard.tsx
@@ -31,13 +31,13 @@ const DRBCard = ({ drbResult }: DRBCardProps) => {
           fontSize="desktop.body.body2"
           display="inline-block"
           mb="s"
-          lang={drbResult.language}
+          lang={drbResult.language !== "en" ? drbResult.language : null}
         >
           {drbResult.title}
         </DSLink>
       </CardHeading>
       <CardContent>
-        {drbResult?.authors.length > 0 ? (
+        {drbResult?.authors?.length > 0 ? (
           <Text size="body2">
             By{" "}
             {drbResult.authors.map((author: Author | Agent, index: number) => (
--- a/src/components/DRB/DRBCard.tsx
+++ b/src/components/DRB/DRBCard.tsx
@@ -4,6 +4,7 @@ import {
   Text,
   Icon,
   Link as DSLink,
+  CardHeading,
 } from "@nypl/design-system-react-components"
 
 import type DRBResult from "../../models/DRBResult"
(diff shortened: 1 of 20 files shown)


src/components/RCLink/RCLink.tsx
```diff
@@ -46,6 +48,10 @@ const RCLink = ({
           },
         }
       }
+      // TODO: These were added in accessibility QA. Investigate adding these to the DS Link component.
+      role="link"
+      aria-disabled={disabled}
+      tabIndex={disabled ? -1 : 0}
```
EdwinGuzman, line 54: Check the QA doc. Clare wrote "The simplest solution is to remove the href to disable the link." I asked whether href should be completely removed or if it should be set to an empty string. She said the former, so do that instead of adding `aria-disabled` or `tabIndex`. `role="link"` should also already be a default of the anchor element so it may be redundant in this case.

src/components/DRB/DRBCard.tsx
```diff
@@ -22,18 +23,20 @@ const DRBCard = ({ drbResult }: DRBCardProps) => {
...
-      <CardContent>
+      <CardHeading level="h3" size="heading6" mb="0">
         <DSLink
           href={drbResult.url}
           target="_blank"
           isUnderlined={false}
           fontSize="desktop.body.body2"
           display="inline-block"
           mb="s"
+          lang={drbResult.language}
```
EdwinGuzman, line 34: I think this might be worth adding when the language is something besides English.

EdwinGuzman (changes requested):
The non-unique links and DRB sidebar result structure look good. But do update the disabled links. For the link language indicator, I'm unsure if it's help to have it if the language is "English".

src/components/RCLink/RCLink.tsx
dgcohen: Sorry should have mentioned that I asked Clare about it on slack and she proposed this as an alternative to removing the href.

dgcohen: @EdwinGuzman Do you think removing the href from an a tag is advisable? I've seen warnings related to this which is why I was looking for a different solution, although it would be a lot simpler than what I did here.

EdwinGuzman: Alright! No problem, can you make a note of that either in the ticket or in the document? The updated proposal seems fine but how it came about can get lost if it's not indicated somewhere.

The warnings are mostly from eslint and react-a11y right? It's "competing" solutions that are both right but removing the "href" may seem less conventional. Let's stick with this if Clare proposed it and it works just fine.

Did she also suggest toggling between tabIndex of -1 and 0? That seems like it should be -1 and null but ask her to go over this current implementation to find out.

dgcohen: Actually after taking a second look, I'm not seeing any warnings from react-a11y or eslint when the href isn't present. Not certain where I ran into this problem previously, might have been due to the lack of href making links inaccessible to screen readers, which I think is what Clare wants in this case.

I think making the href null should be ok in this case since it'll keep the code simpler and it's what Clare is recommending.

dgcohen: Confirmed with @EdwinGuzman that we can keep the solution that Claire proposed via our private slack (adding tabindex -1 and aria-disabled). I've updated the QA doc and the ticket with a brief write up on our conversation.

EdwinGuzman (approved):

src/components/DRB/DRBCard.tsx
dgcohen: Good call! Added a check to only render the language when it's not english.

EdwinGuzman (approved):

src/components/RCLink/RCLink.tsx
```diff
@@ -46,6 +48,10 @@ const RCLink = ({
           },
         }
       }
+      // TODO: These were added in accessibility QA. Investigate adding these to the DS Link component.
+      role="link"
+      aria-disabled={disabled}
+      tabIndex={disabled ? -1 : null}
```
charmingduchess, line 54: I think the alternative to -1 should be 0, which would include the link in sequential tabbing order. [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) says tabbing behavior with an absent tabIndex (I imagine null would be absent) "is determined by the user agent." I think it would be better to definitely have the link included in sequential tabbing.

charmingduchess (changes requested):

src/components/RCLink/RCLink.tsx
charmingduchess: @dgcohen not sure if you saw this comment, but once this is addressed I will approve.

charmingduchess (approved):

src/components/RCLink/RCLink.tsx
dgcohen: Not sure if you saw but i added the 0 index back in

PR #164, opened 2024-04-09
Scc 4063/search results vqa

charmingduchess:
Addresses [this ticket](https://jira.nypl.org/browse/SCC-4063)
- focus updates
  - opening refine search should focus on cancel, clicking cancel should focus on refine search button
  - applying filters, clearing filters, removing filters from the tagset focus on "Displaying n results..." heading
  - sorting results should return focus to sort by selector
- refine checkboxes layout
  - checkboxes switch to rows of 2 on mobile view, text does not overlap with checkboxes
- there was also a note about updating colors, but it looks like that will be happening DS-side

In order to make the focus work simpler, I refactored the search page. Instead of rendering the layout and the layout decides when and where to render the search form, the search form is generated on the search page and then passed into the layout. This allowed for the removal of the search results aggregations context.

### DIFF
--- a/pages/search/index.tsx
+++ b/pages/search/index.tsx
@@ -172,7 +172,7 @@ export default function Search({
                   // Heading component does not expect tabIndex prop, so we
                   // are ignoring the typescript error that pops up.
                   // @ts-expect-error
-                  tabIndex="0"
+                  tabIndex={-1}
                   mb="xl"
                   size="heading4"
                   ref={searchResultsHeadingRef}
--- a/pages/search/index.tsx
+++ b/pages/search/index.tsx
@@ -6,7 +6,7 @@ import {
   Select,
   SkeletonLoader,
 } from "@nypl/design-system-react-components"
-import { useEffect, useRef, type ChangeEvent } from "react"
+import { useEffect, useRef, useState, type ChangeEvent } from "react"
 import { useRouter } from "next/router"
 import { parse } from "qs"
 
@@ -14,6 +14,7 @@ import SearchForm from "../../src/components/SearchForm/SearchForm"
 import Layout from "../../src/components/Layout/Layout"
 import DRBContainer from "../../src/components/DRB/DRBContainer"
 import SearchResult from "../../src/components/SearchResult/SearchResult"
+import AppliedFilters from "../../src/components/SearchFilters/AppliedFilters"
 
 import { fetchResults } from "../../src/server/api/search"
 import {
@@ -35,7 +36,7 @@ import type SearchResultsBib from "../../src/models/SearchResultsBib"
 
 import useLoading from "../../src/hooks/useLoading"
 import initializePatronTokenAuth from "../../src/server/auth"
-import AppliedFilters from "../../src/components/SearchFilters/AppliedFilters"
+import a11yStyles from "../../styles/utils/a11y.module.scss"
 
 interface SearchProps {
   bannerNotification?: string
@@ -107,6 +108,13 @@ export default function Search({
     searchResultsHeadingRef?.current?.focus()
   }, [isLoading, isFreshSortByQuery])
 
+  const [headerText, setHeaderText] = useState(
+    getSearchResultsHeading(searchParams, totalResults)
+  )
+  useEffect(() => {
+    setHeaderText(getSearchResultsHeading(searchParams, totalResults))
+  }, [totalResults, searchParams])
+
   const searchForm = <SearchForm aggregations={aggs} />
   return (
     <>
@@ -120,6 +128,14 @@ export default function Search({
         <meta name="twitter:title" content={metadataTitle} key="tw-title" />
         <title key="main-title">{metadataTitle}</title>
       </Head>
+      {/* this span is here to ensure that screen readers are updated
+      when the page has finished loading. in particular, after applying a sort
+      to the results, focus remains on the selector for accessibility reasons.
+      this span ensures that a screen reader will also alert the user that
+      new search results have been delivered. */}
+      <span className={a11yStyles.screenreaderOnly} aria-live="polite">
+        {headerText}
+      </span>
       <Layout
         searchForm={searchForm}
         isAuthenticated={isAuthenticated}
@@ -177,7 +193,7 @@ export default function Search({
                   size="heading4"
                   ref={searchResultsHeadingRef}
                 >
-                  {getSearchResultsHeading(searchParams, totalResults)}
+                  {headerText}
                 </Heading>
                 <SimpleGrid columns={1} gap="grid.xl">
                   {searchResultBibs.map((bib: SearchResultsBib) => {
--- a/pages/search/index.tsx
+++ b/pages/search/index.tsx
@@ -22,6 +22,7 @@ import {
   mapElementsToSearchResultsBibs,
   getSearchQuery,
   sortOptions,
+  getFreshSortByQuery,
 } from "../../src/utils/searchUtils"
 import type {
   SearchResultsResponse,
@@ -31,7 +32,6 @@ import type {
 import { mapWorksToDRBResults } from "../../src/utils/drbUtils"
 import { SITE_NAME, RESULTS_PER_PAGE } from "../../src/config/constants"
 import type SearchResultsBib from "../../src/models/SearchResultsBib"
-import { SearchResultsAggregationsProvider } from "../../src/context/SearchResultsAggregationsContext"
 
 import useLoading from "../../src/hooks/useLoading"
 import initializePatronTokenAuth from "../../src/server/auth"
@@ -41,6 +41,7 @@ interface SearchProps {
   bannerNotification?: string
   results: SearchResultsResponse
   isAuthenticated: boolean
+  isFreshSortByQuery: boolean
 }
 
 /**
@@ -51,6 +52,7 @@ export default function Search({
   bannerNotification,
   results,
   isAuthenticated,
+  isFreshSortByQuery,
 }: SearchProps) {
   const metadataTitle = `Search Results | ${SITE_NAME}`
   const { push, query } = useRouter()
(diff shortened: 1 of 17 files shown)


EdwinGuzman:
Check the Vercel homepage. The search bar is missing and you cannot make a search from there.

src/components/MyAccount/RequestsTab/CancelButton.tsx
```diff
@@ -77,6 +75,7 @@ const CancelButton = ({
           <Text sx={{ marginBottom: 0 }}>Cancel request?</Text>
         </Box>
       ),
+      //@ts-ignore
```
dgcohen, line 78: Is this here because it's hard to determine the event type?

dgcohen (changes requested):
Not seeing the Search Form anymore on the home page, can you add that back and add a test to check that it's rendered?

pages/search/index.tsx
```diff
@@ -92,8 +96,28 @@ export default function Search({
     )
   }
 
+  const searchResultsHeadingRef = useRef(null)
+  useEffect(() => {
+    // don't focus on "Displaying n results..." if the page is not done loading
```
dgcohen, line 101: Great to see comments here!

pages/search/index.tsx
```diff
@@ -92,8 +96,28 @@ export default function Search({
...
+    if (isLoading) return
+    // keep focus on sort by selector if the last update to the query was a sort
+    if (isFreshSortByQuery) {
```
dgcohen, line 104: Small nit but can the parens be removed here?

pages/search/index.tsx
```diff
@@ -173,10 +216,13 @@ export default function Search({
            * TODO: The logic and copy for different scenarios will need to be added when
            * filters are implemented
            */
-          <Heading level="h3">No results. Try a different search.</Heading>
+          // @ts-expect-error
```
dgcohen, line 219: Out of curiosity, what is the error?

pages/search/index.tsx
```diff
@@ -104,7 +128,16 @@ export default function Search({
...
         <title key="main-title">{metadataTitle}</title>
       </Head>
+      {/* this span is here to ensure that screen readers are updated
+      when the page has finished loading. in particular, after applying a sort
+      to the results, focus remains on the selector for accessibility reasons.
+      this span ensures that a screen reader will also alert the user that
+      new search results have been delivered. */}
+      <span className={a11yStyles.screenreaderOnly} aria-live="polite">
+        {headerText}
+      </span>
```
EdwinGuzman, line 138: As discussed offline, let's make this a new ticket and refine the work and expected result.
Right now, when testing this with voiceover, the "Displaying 1-...." text was read twice but now read at all when the sort by select dropdown was updated.

src/components/Layout/Layout.tsx
```diff
@@ -70,7 +72,7 @@ const Layout = ({
                   isAuthenticated={isAuthenticated}
                   activePage={activePage}
                 />
-                {showSearch && <SearchForm />}
```
EdwinGuzman, line 73: Remove the import from line 13.

pages/search/index.tsx
```diff
@@ -92,8 +96,28 @@ export default function Search({
...
+  const [headerText, setHeaderText] = useState(
+    getSearchResultsHeading(searchParams, totalResults)
+  )
+  useEffect(() => {
+    setHeaderText(getSearchResultsHeading(searchParams, totalResults))
+  }, [totalResults, searchParams])
+
+  const searchForm = <SearchForm aggregations={aggs} />
   return (
-    <SearchResultsAggregationsProvider value={aggs}>
```
EdwinGuzman, line 96: Why the update from context to prop? Not that it's a bad thing but just curious about the update. If you don't need `SearchResultsAggregationsProvider` then remove the function/file.

EdwinGuzman (commented):
This looks good but I'll do another closer look when this PR gets updated and the hidden span is removed.

src/components/MyAccount/RequestsTab/CancelButton.tsx
charmingduchess: no this is because we are overriding the expected type for this method, which is a method with no args. we are hacking the modal by availing ourselves of the event which is actually passed in to the callback. i'll add a comment explaining.

pages/search/index.tsx
charmingduchess: we love a single line if statement 💅🏻

pages/search/index.tsx
charmingduchess: type error due to heading component not expecting tabindex. i added a comment.

pages/search/index.tsx
charmingduchess: i moved things around when i thought i'd have to pass refs down. i still think the refactor makes sense, but it might actually not work (i totally missed the search form on the home page 🤦🏻

src/components/SearchForm/SearchForm.tsx
```diff
@@ -1,7 +1,7 @@
 import { SearchBar } from "@nypl/design-system-react-components"
 import { useRouter } from "next/router"
-import type { SyntheticEvent, Dispatch, SetStateAction } from "react"
-import { useContext, useState, useEffect } from "react"
+import type { SyntheticEvent, Dispatch, SetStateAction, RefObject } from "react"
```
EdwinGuzman, line 3: Don't think you need `RefObject`, right?

pages/search/index.tsx
```diff
@@ -30,16 +33,16 @@ import type {
 import { mapWorksToDRBResults } from "../../src/utils/drbUtils"
 import { SITE_NAME, RESULTS_PER_PAGE } from "../../src/config/constants"
 import type SearchResultsBib from "../../src/models/SearchResultsBib"
-import { SearchResultsAggregationsProvider } from "../../src/context/SearchResultsAggregationsContext"
 
 import useLoading from "../../src/hooks/useLoading"
 import initializePatronTokenAuth from "../../src/server/auth"
-import AppliedFilters from "../../src/components/SearchFilters/AppliedFilters"
+import a11yStyles from "../../styles/utils/a11y.module.scss"
```
EdwinGuzman, line 39: This is no longer being used.

pages/search/index.tsx
```diff
@@ -148,8 +164,20 @@ export default function Search({
...
+                {displayAppliedFilters && (
+                  <AppliedFilters aggregations={aggs} />
+                )}
+                <Heading
+                  data-testid="search-results-heading"
+                  level="h2"
+                  // Heading component does not expect tabIndex prop, so we
+                  // are ignoring the typescript error that pops up.
+                  // @ts-expect-error
+                  tabIndex="0"
```
EdwinGuzman, line 176: You normally would not tab on the heading. For programmatic focus, `-1` is used. Check with Clare on this.

src/components/Layout/Layout.tsx
```diff
@@ -38,6 +40,10 @@ const Layout = ({
   const showSearch = activePage === "search"
   const showHeader = activePage !== "404"
   const showNotification = activePage === "" || activePage === "search"
+  // if the search page is rendering the layout, it will pass a search form
+  // with a refine search button. otherwise, a plain search form will do.
+  console.log({ searchFormWithRefineResults })
```
EdwinGuzman, line 45: Remove the log.

src/components/Layout/Layout.tsx
```diff
@@ -70,7 +76,7 @@ const Layout = ({
                   isAuthenticated={isAuthenticated}
                   activePage={activePage}
                 />
-                {showSearch && <SearchForm />}
+                {showSearch && searchForm}
```
EdwinGuzman, line 79: Why not update `searchForm` to be `showSearch || searchFormWithRefineResults || <SearchForm />`? There are now two logic checks in separate locations to render `<SearchForm />`.

src/components/MyAccount/RequestsTab/CancelButton.tsx
EdwinGuzman: Hopefully we make an update for this this sprint on the DS side.

src/components/RefineSearch/RefineSearch.tsx
```diff
@@ -97,24 +97,26 @@ const RefineSearch = ({
       pathname: "/search",
       query: updatedQuery,
     })
-    // refine search dialog closes after url is pushed
     toggleRefine()
   }
 
+  // the two buttons can share a ref because they are never rendered at the same
```
EdwinGuzman, line 103: Hmm I would think that you would need to update the `refineCancelRef.current` value to get the appropriate DOM element. Perhaps rename `refineOrCancelRef`?

pages/search/index.tsx
EdwinGuzman: It looks like the home search does work though. So is the homepage going to be updated in the future with a simliar update?

EdwinGuzman (commented):
This works well with the changes Clare suggested. Since it's still a PR, can you ask her for a review? If there are any changes, it'll be easier to update it in this PR than later. 

I requested changes but they're minor.

src/components/Layout/Layout.tsx
charmingduchess: Ok! I updated it so the search page passes aggregations instead of a SearchForm to Layout and the search form decides what happens based on that prop.

src/components/Layout/Layout.tsx
charmingduchess: done

pages/search/index.tsx
charmingduchess: updated

src/components/RefineSearch/RefineSearch.tsx
charmingduchess: Yea I was surprised that it worked as well. I renamed to include `or`.

charmingduchess:
> It looks like the home search does work though. So is the homepage going to be updated in the future with a simliar update?

with what similar update?

charmingduchess:
@EdwinGuzman 

> It looks like the home search does work though. So is the homepage going to be updated in the future with a simliar update?

with what similar update?

pages/search/index.tsx
charmingduchess: removed

pages/search/index.tsx
charmingduchess: what similar update?

src/components/SearchForm/SearchForm.tsx
charmingduchess: removed

dgcohen (approved):

pages/search/index.tsx
EdwinGuzman: Never mind. I confused this with the issue of the searchbar not rendering on the homepage.

EdwinGuzman (approved):
I know the tests were a headache but glad to see these tests in the repo.

PR #169, opened 2024-04-15
SCC-4066 - Small VQA fixes

dgcohen:
Some small fixes that I wanted to get in before the second round of VQA. Noticed the Advanced search link was no longer on the left so I added a margin: left style.

Also removed a console log in passing and nudged the advanced search link up with a negative margin per Apoorva's comment in the Search Bar section of the VQA doc.

### DIFF
--- a/styles/components/Search.module.scss
+++ b/styles/components/Search.module.scss
@@ -28,6 +28,11 @@
 .advancedSearch {
   align-self: flex-start;
   min-width: 127px;
+  line-height: 1;
+
+  @media screen and (min-width: 600px) {
+    margin-top: -8px;
+  }
 }
 
 .refineButtons {
--- a/styles/components/Search.module.scss
+++ b/styles/components/Search.module.scss
@@ -23,6 +23,10 @@
   justify-content: space-between;
   margin-top: var(--nypl-space-s);
   margin-bottom: var(--nypl-space-s);
+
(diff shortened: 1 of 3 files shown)


styles/components/Search.module.scss
```diff
@@ -28,6 +28,12 @@
 .advancedSearch {
   align-self: flex-start;
   min-width: 127px;
+  line-height: 1;
+  margin-left: auto;
+
+  @media screen and (min-width: 600px) {
+    margin-top: -8px;
+  }
```
EdwinGuzman, line 36: The margin between the searchbar and this link should be 16px which is what the parent is doing. So this doesn't seem like it's needed.

styles/components/Search.module.scss
```diff
@@ -28,6 +28,12 @@
 .advancedSearch {
   align-self: flex-start;
   min-width: 127px;
+  line-height: 1;
+  margin-left: auto;
```
EdwinGuzman, line 32: I noticed this issue and it seems it only happens on the homepage. I fixed this but it seems it was lost in the merge. So this is okay but I think I did it differently.

EdwinGuzman (changes requested):
Just remove the -8 margin.

EdwinGuzman (approved):

PR #172, opened 2024-04-17
SCC-4066 - VQA Fixes (Apoorva Second Pass)

dgcohen:
Here are some fixes based on Apoorva's latest comments in the VQA doc. Some of the changes requested in the doc should be addressed at the design system or data level, I've listed those out in a new section at the end of the VQA doc.

Here is a summary of the changes included here:

1. Extracted the Search Results Sort select into its own component.
2. Used the CSS display property to determine whether to show the Sort Select in the main content or the sidebar. Note: I originally used the DS breakpoints to conditionally render the component but noticed a lag issue where the Select would appear in the main section for a couple of seconds before moving to the sidebar.
3. General spacing/font fixes (primarily related to mobile)
4. Item location is no longer being hidden on Mobile.


__test__/pages/search/searchResults.test.tsx
```diff
@@ -62,7 +62,7 @@ describe("Search Results page", () => {
           results={{ results, aggregations: aggregationsResults }}
         />
       )
-      const selector = screen.getByLabelText("Sort by")
+      const selector = screen.getAllByLabelText("Sort by")[0]
```
charmingduchess, line 65: Do you know why there are multiple elements with this label text?

src/components/ItemTable/ItemAvailability.tsx
```diff
@@ -87,7 +87,13 @@ const ItemAvailability = ({ item }: ItemAvailabilityProps) => {
         <Button
           id="contact-librarian"
           buttonType="link"
-          sx={{ display: "inline", fontWeight: "inherit", fontSize: "inherit" }}
+          sx={{
```
charmingduchess, line 90: can you move this to a css module?

__test__/pages/search/searchResults.test.tsx
dgcohen: See my note in the PR description about having to render the Sort Select in different sections at different breakpoints.

src/components/ItemTable/ItemAvailability.tsx
dgcohen: Tried moving this into a CSS module but the specificity wasn't high enough to override the DS styles.

src/components/SearchResults/ElectronicResourcesLink.tsx
```diff
@@ -17,8 +17,12 @@ const ElectronicResourcesLink = ({
   electronicResources,
 }: ElectronicResourcesLinkProps) => {
   return (
-    <Box mt="s" mb="s">
-      <Text mb="xxs" fontSize="desktop.body.body1" fontWeight="bold">
+    <Box mt="s">
+      <Text
+        mb="xxs"
+        fontSize={{ base: "mobile.body.body1", md: "desktop.body.body1" }}
```
EdwinGuzman, line 23: @bigfishdesign13 just noticed that the "default" or "body" values for the `size` prop do not use this font size value. Should we update this in the DS?

src/components/DRB/DRBCard.test.tsx
```diff
@@ -73,6 +73,6 @@ describe("DRBCard", () => {
     const titleLink = screen.getByText(/Spanish Title/i)
 
     expect(titleLink).toBeInTheDocument()
-    expect(titleLink).toHaveAttribute("lang", "es")
```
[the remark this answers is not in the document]
dgcohen: I modified the DRB title component to use a CardHeading with a url prop instead of rendering the Link as a child. This puts the lang attribute on the surrounding h3 which is why I changed this to check the parentElement.

src/components/SearchResults/ElectronicResourcesLink.tsx
dgcohen: Yeah I noticed that the body1 and body2 variants in the Text component don't have the appropriate values set. We should fix that on the DS level so we can specify a variant and remove these overrides.

Maybe we can keep these for now and add TODO comments?

src/components/SearchResults/SearchResultsSort.tsx
```diff
@@ -0,0 +1,51 @@
...
+  pageHasResults,
+  searchParams,
+  handleSortChange,
+  mobileOnly = false,
+}: SearchResultsSortProps) => {
+  if (!pageHasResults) return
+  return (
+    <Select
+      name="sort_direction"
+      id={`search-results-sort${mobileOnly && "-mobile"}`}
```
charmingduchess, line 27: not a huge issue, but this id comes up as "search-results-sortfalse-wrapper" in the inspector, which I don't believe was the intention.

pages/search/index.tsx
charmingduchess, line 1: I'm having trouble tracing where the mobile or desktop selectors are being hidden. Can you illuminate?

charmingduchess (commented):
The styles are looking good. I have mostly clarifying questions. 
I did also notice an accessibility regression, though. After updating the the sort by selector, the focus should land back on that selector after the page loads. I did some work on that realm for the first round of VQA. There is a prop passed to the search index page, `isFreshSortByQuery`, which you should be able to use to help with that. If that prop is true, you want to focus on the selector. I'm not sure why it's not happening automatically given that it does so in the desktop version. We can also flag this to be its own ticket if it feels out of scope.

pages/search/index.tsx
dgcohen: The hiding and showing is being controlled by the display prop on the Select component in SearchResultSort.tsx.

EdwinGuzman (changes requested):
Some changes based on the doc.

charmingduchess (changes requested):
As we discussed offline, please move the rendering logic for the search results sort to be apparent from the search index page.

src/components/DRB/DRBCard.tsx
```diff
@@ -24,24 +26,31 @@ const DRBCard = ({ drbResult }: DRBCardProps) => {
...
-          isUnderlined={false}
-          fontSize="desktop.subtitle.subtitle2"
-          fontWeight="medium"
-          display="inline-block"
-          mb="xs"
-          lang={drbResult.language !== "en" ? drbResult.language : null}
-        >
-          {drbResult.title}
-        </DSLink>
+      <CardHeading
```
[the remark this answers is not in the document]
dgcohen: I originally changed this to match the markup for the main Search Results titles, but I reverted it back to how it was due to the lang issue you mentioned.

Btw, the undefined-link id is appearing in the main Search Result titles, should I fix that as part of this PR? Or should we wait to address these issues in the DS?

src/components/DRB/DRBCard.tsx
```diff
@@ -50,7 +59,10 @@ const DRBCard = ({ drbResult }: DRBCardProps) => {
                 <DSLink
```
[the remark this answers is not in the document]
dgcohen: Fixed

dgcohen:
> As we discussed offline, please move the rendering logic for the search results sort to be apparent from the search index page.

Removed the logic from the SearchResultsSort component. The show/hide functionality and the id are now being controlled from the search results index page.

src/components/DRB/DRBCard.tsx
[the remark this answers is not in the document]
EdwinGuzman: Fix it in this PR if possible.

EdwinGuzman (approved):

src/components/DRB/DRBCard.tsx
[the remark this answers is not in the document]
dgcohen: Fixed!

charmingduchess (approved):
Thanks for including my changes. Just one last request - could you leave comments above the display props on the SearchResultsSort explaining why you opted for that approach?

PR #173, opened 2024-04-18
Scc 3928/account settings update

charmingduchess:
This is not quite done yet, but I would love to get some eyes on it.
In particular, the way I decided to render the account settings display and form is a little kooky. I think it makes sense, but I would love some input on whats going on in AccountSettingsDisplayOptions.tsx and AccountSettingsTab.tsx


src/components/MyAccount/Settings/AccountSettingsButtons.tsx
```diff
@@ -0,0 +1,55 @@
...
+interface AccountSettingsButtonsPropsType {
+  currentlyEditing: boolean
+  setCurrentlyEditing: Dispatch<React.SetStateAction<boolean>>
+}
+
+const AccountSettingsButtons = ({
+  currentlyEditing,
+  setCurrentlyEditing,
+}: AccountSettingsButtonsPropsType) => {
+  const toggleCurrentlyEditing = () => setCurrentlyEditing(!currentlyEditing)
```
EdwinGuzman, line 13: Use the value from the callback argument to update the value: `setCurrentlyEditing((prevValue) => !prevValue)`

src/components/MyAccount/Settings/AccountSettingsButtons.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  const clearUpdate = () => {
```
EdwinGuzman, line 15: It's hard to tell because it's not testable today, but why not directly pass the value to the behavior of the button? So always set the value to false for the cancel button and true for the edit button?

I'm assuming that the toggle works because you can't access the edit button if the value is not already false, so calling `setCurrentlyEditing` will always set it to true. But perhaps being explicit will help read through this.

src/components/MyAccount/Settings/AccountSettingsUtils.ts
```diff
@@ -0,0 +1,121 @@
...
+    icon: "actionLockClosed",
+    term: "Pin/Password",
+    description: (_) => "****",
+  },
+] as {
+  field: string
+  icon: IconNames
+  term: string
+  description?: (any) => string
+}[]
```
EdwinGuzman, line 56: I think you can create a new type that extends `IconListElementPropType`.

src/components/MyAccount/Settings/AccountSettingsTab.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const AccountSettingsTab = ({ settingsData }: { settingsData: Patron }) => {
+  const [currentlyEditing, setCurrentlyEditing] = useState(false)
+  const [mostRecentPatronData, setMostRecentPatronData] = useState(settingsData)
+  const [modalProps, setModalProps] = useState(null)
+  const listElements = currentlyEditing ? (
+    <AccountSettingsForm patron={mostRecentPatronData} />
+  ) : (
+    <AccountSettingsDisplay patron={mostRecentPatronData} />
+  )
```
EdwinGuzman, line 31: I think this is what you flagged in the PR description. I would try to have one component that returns either the form or the display, but doing it this way prevents having to pass `currentlyEditing` into the single component. And this is also done at the parent level so this seems fine to me.

You mentioned it's a little kooky but is this also what you were referring to?

EdwinGuzman (commented):
Once ready for review, it'd be helpful to merge the latest from `main` because some changes in this PR are already in `main. Just to better see the updates.

src/components/MyAccount/Settings/AccountSettingsUtils.ts
charmingduchess: There might be a better way to do this, but I actually can't make those the same type. i'm calling this property description because it can be a method to return the description, but the IconListElement type expects a string or a react node, and combining these two types makes issues with that discrepancies. Not sure exactly how to address this. this account settings object doesn't really need a type is maybe one solution?

src/components/MyAccount/Settings/AccountSettingsTab.tsx
charmingduchess: Yea I like that it's happening in the parent. The kookiness I'm referring to is more about the business of like bending the icon list element pattern into the form. I feel like I have all of these switch statements everywhere and I'm not sure if it's legible to anyone but me.

EdwinGuzman (approved):

PR #176, opened 2024-04-19
SCC-4010: Logging for Cloudwatch

7emansell:
Resolves ticket [SCC-4010](https://jira.nypl.org/browse/SCC-4010).

- Adds [Winston](https://www.npmjs.com/package/winston) logging package and sets up NYPL logging standards. 
- Adds mock logger (which is not necessary yet but will be once we write a11y testing for pages)
- Adds a section in the README for logging
- Changes all logs in `getServerSideProps` to use the logger
- Cleans up logs a little bit– we don't need to log the patron authentication token on the homepage, and probably not on the account page either (I'll remove it once My Account work phase 1 is done)

Test by running the app locally and clicking My Account. You should see a info-level log of the patron authentication token in the terminal.

This should automatically work in AWS Cloudwatch but may need another PR to correct it if not, we'll know once we push to qa/prod.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -6,36 +6,21 @@ const initializeLogger = () => {
     //
   }
 
-  // Set default NYPL agreed upon log levels
+  // NYPL levels allowed by console
   const nyplLogLevels = {
-    emergency: 0,
-    alert: 1,
-    critical: 2,
-    error: 3,
-    warning: 4,
-    notice: 5,
-    info: 6,
-    debug: 7,
+    warn: 1,
+    info: 2,
+    debug: 3,
   }
 
   const getLogLevelCode = (levelString) => {
     switch (levelString) {
-      case "emergency":
-        return 0
-      case "alert":
+      case "warn":
         return 1
-      case "critical":
-        return 2
-      case "error":
-        return 3
-      case "warning":
-        return 4
-      case "notice":
-        return 5
       case "info":
-        return 6
+        return 2
       case "debug":
-        return 7
+        return 3
       default:
         return "n/a"
     }
@@ -56,7 +41,7 @@ const initializeLogger = () => {
   })
 
   const transports = [
-    new winston.transports.Console(),
+    new winston.transports.Console({ level: "debug" }),
     new winston.transports.File({
       filename: path.resolve(process.cwd(), "log", "rc.log"),
       // Log format space limited
--- a/logger.js
+++ b/logger.js
@@ -8,19 +8,22 @@ const initializeLogger = () => {
 
   // NYPL levels allowed by console
   const nyplLogLevels = {
-    warn: 1,
-    info: 2,
-    debug: 3,
+    error: 1,
+    warn: 2,
+    info: 3,
+    debug: 4,
   }
 
   const getLogLevelCode = (levelString) => {
     switch (levelString) {
-      case "warn":
+      case "error":
         return 1
-      case "info":
+      case "warn":
         return 2
-      case "debug":
+      case "info":
         return 3
+      case "debug":
+        return 4
       default:
         return "n/a"
     }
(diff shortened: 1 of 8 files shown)


logger.js
```diff
@@ -0,0 +1,86 @@
...
+      timestamp({
+        format: "YYYY-MM-DD hh:mm:ss.SSS A",
+      }),
+      formatter
+    ),
+    transports,
+  })
+}
+
+const isRunningOnVercel = process.env.VERCEL === "1"
```
7emansell, line 82: This was discovered by trial and error, and was new to me– next.js knows when it's in Vercel whether you create that env variable or not. this just moves the logs to the console when it's in Vercel because it can break the deployment (even if it works in prod).

pages/index.tsx
```diff
@@ -9,6 +9,7 @@ import {
   SkeletonLoader,
 } from "@nypl/design-system-react-components"
 import Head from "next/head"
+import logger from "../logger"
```
EdwinGuzman, line 12: It's not being used in this file.

pages/account/[[...index]].tsx
```diff
@@ -198,27 +199,27 @@ export async function getServerSideProps({ req }) {
   // Parsing path from url to pass to ProfileTabs.
   const tabsPath = req.url.split("/", -1)[2] || null
   const id = patronTokenResponse.decodedPatron.sub
-  // try {
-  const { checkouts, holds, patron, fines } = await MyAccountFactory(id)
-  // Redirecting /fines if user has none.
-  if (tabsPath === "overdues" && fines.total === 0) {
+  try {
```
Comment (EdwinGuzman) [inline pages/account/[[...index]].tsx:202]: I feel like I see this update in various PRs. Hope they can all be synced soon so there aren't many merge conflicts.

EdwinGuzman (approved):
Thanks!

charmingduchess (changes requested):

logger.js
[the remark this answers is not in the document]
7emansell: I agree and @EdwinGuzman do you think it's okay to remove levels for this repo before opening the PR for the overall standards?

pages/account/[[...index]].tsx
Comment (7emansell) [reply pages/account/[[...index]].tsx:202]: Yes that's on me sorry . Will pull down a clean copy after this pr

logger.js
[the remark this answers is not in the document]
EdwinGuzman: Yes, remove the levels for this repo. When the eng general PR is opened, this PR can be linked as an example.

7emansell:
Removed `emergency`, `alert`, `critical`, and `notice` log levels since they don't work with `console` or give a type error, and changed `warning` to `warn`.

EdwinGuzman (commented):
The last update looks good, but sorry, I just realized that `error` does exist https://developer.mozilla.org/en-US/docs/Web/API/console/error_static

I don't think we ever really use it but keep this for now to be consistent with the web console API.

EdwinGuzman (approved):

dgcohen (approved):
Looks great!

charmingduchess (approved):
Looks good! Just one small ask - Not sure if anyone else sees this, but github isn't colorizing `pages/index.tsx` on my view. Can you double check in vscode that there aren't any issues with formatting or extra brackets floating around?

7emansell:
> Looks good! Just one small ask - Not sure if anyone else sees this, but github isn't colorizing `pages/index.tsx` on my view. Can you double check in vscode that there aren't any issues with formatting or extra brackets floating around?

i see it too that is so weird? I double checked in vscode and it's fine there, it has .tsx extension so I'm not sure what else to do.

EdwinGuzman:
Don't think it's a big issue but could it be with the HTML encoding that is not being parsed correctly? Such as https://github.com/NYPL/research-catalog/blob/aee9eef2ddcb9118111209b6b773cbab7e104158/pages/index.tsx#L46

Shouldn't be a blocker to merge in.

PR #177, opened 2024-04-19
use banner component instead of homebaked Notification

charmingduchess:
I imported the banner component. It's not a 100% match to whats on prod. I found that if I passed in the banner as part of the main content in the Layout, I could skip some of the extra style overrides. this means that the notification is next to, not above the side bar. I think it's worth it to remove css overrides, but I will reach to to Apoorva to confirm if this change is ok.


charmingduchess:
@apoorvakhanolkar this implementation varies slightly from what we had previously. Let me know if the search results notification looks okay to you here: https://research-catalog-git-noref-import-banner-component-nypl.vercel.app/research/research-catalog/search?q=spaghetti

apoorvakhanolkar:
Hmm would definitely prefer for the banner to appear above (and not next to) the sidebar. Are there any inherent issues with the component itself that's causing the issue? If so could we talk to Marty and see it there's a solution? @charmingduchess @EdwinGuzman

EdwinGuzman:
@charmingduchess some of the changes were made in `main` directly. That's fine but it's hard to see that update in this PR. The issue Apoorva mentioned should be fixed if the banner is moved from the `contentPrimary` section to the `breakout` section (below `SearchForm`) in `Layout.tsx`.

dgcohen:
Aside from the placement of the banner everything looks good.

charmingduchess:
It's not a huge issue, but as is, if the Banner component is passed into the DS template in the breakout (as opposed to contentPrimary where it is now), the margins are overridden somewhere in the template. to me, the margins of the banner component are pretty integral to its design, so it seems strange that it's losing the margins. @EdwinGuzman is this something that could be easily fixed? As in, could the notification component be harder to override?
I wouldn't want this to hold up the Search Results release, and I could add in some custom CSS to get it out. But this component in particular feels like it shouldn't need it.

EdwinGuzman:
Can you post a screenshot of what you're seeing when it's in the breakout? There should be a max width on the banner so it's not the entire width of the template row, but I'm not sure what you mean by the margin.

EdwinGuzman:
I see what you mean now. I was wrong, there is no max width on the banner but rather on the content inside of the banner. 

What you can do (after moving it to the breakout section) is wrap the Banner in a div and set the margins there so that the banner is centered horizontally and has "l / 32px / 2rem" top and bottom margin.

EdwinGuzman:
I think this is a case where the design of the app is dictating how a component should display. I'll still discuss this with Marty.

charmingduchess:
@apoorvakhanolkar can you please confirm that this looks ok? i moved the component above the sidebar.

charmingduchess:
@apoorvakhanolkar I updated the styles so the max width is 1248px. Thanks for your feedback!

charmingduchess:
oops! i actually created a different branch for this now that there's a ticket with a number. I will close this and tag you there.

PR #178, opened 2024-04-22
[SCC-4091] Add ExternalLink component and update RCLink default href

dgcohen:
After realizing that we would have to add hasVisitedState false to all external links (as well as target="_blank" in most cases), I thought we should create a new component ExternalLink that would add these by default.

We currently have to explicitly define all the props that are propagated to the DSLink since the LinkProps currently aren't being exported from the DS. We should be able to get rid of these in both ExternalLink and RCLink once LinkProps export is added in the DS (I can discuss this with the team).

I also added the BASE_URL to the beginning of hrefs in RCLink by default since most of these will be internal and this might circumvent problems, similar to the one we ran into this week with the hold buttons where the base_url was missing. I added a prop that overrides this behavior in the cases where the base_url is already set in the href for whatever reason.


src/components/Links/ExternalLink/ExternalLink.test.tsx
```diff
@@ -15,18 +15,9 @@ describe("RCLink", () => {
       expect(link).toHaveAttribute("href", "/spaghetti")
     })
   })
-  it("should add appropriate accessibility attributes when link is disabled", async () => {
```
EdwinGuzman, line 18: This tests should still be applicable, no?

src/components/Links/ExternalLink/ExternalLink.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// TODO: once 2ad is phased out, replace with DS v3 Link which can wrap a
+// next link.
+
+/**
+ * The ExternalLink component is a utility that renders a Design System link with some defaults
+ * set for the Research Catalog to avoid duplication.
+ */
+const ExternalLink = ({
```
EdwinGuzman, line 31: This and `RCLink` share a lot of code that I'm afraid eventually it will become difficult to maintain both. Really, the only big difference is how `RCLink` handles the href and the state for the subnav (active and white focus ring). It's okay for now but if it becomes harder to maintain, or if we forget to use these and use a different link, then we should revisit this.

EdwinGuzman (changes requested):
Looks fine but the prop types should be more generic instead of listing down each style prop.

src/components/Links/ExternalLink/ExternalLink.tsx
```diff
@@ -0,0 +1,56 @@
...
+  type?: LinkTypes
+  fontSize?: string | Record<string, string>
+  fontWeight?: string
+  display?: string
+  rel?: string
+  mb?: string
+  mt?: string
+  lang?: string
+  isUnderlined?: boolean
+  disabled?: boolean
```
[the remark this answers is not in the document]
dgcohen: Good call, I can fix this in RCLink as well.

src/components/Links/ExternalLink/ExternalLink.tsx
dgcohen: I agree that they share a lot of functionality right now, but I think that will change when we change RCLink to wrap the link in the default Next Link again.

dgcohen: It's possible that they can be combined into a single component if we add props to differentiate them, however.

EdwinGuzman:
This looks good. I'm going to fix the merge conflicts and once approved @charmingduchess, we can merge it in.

charmingduchess (approved):
Looks good. Even updating the names to be complementary (External/RC) does a lot for my own understanding of the app. I wonder about the default target value being "_blank". Is that based on some DS standard around external linking? It doesn't seem like the external link is actually opening a new tab much more than it's opening in the current tab.

PR #179, opened 2024-04-23
SCC-4092 - Persist Search Filters on sort

dgcohen:
Ticket: https://jira.nypl.org/browse/SCC-4092

This issue was caused by not properly parsing the search filters as set in the query string to the structure expected by SearchParams. I added a new function mapQueryToSearchParams that maps these values correctly.

I also removed the sort_scope and per_page params from SearchQueryParams since they weren't being in mapQueryToSearchParams and I was seeing them as the expected to be in the spread variable ...queryFilters.

I also set a default value in the useState for searchScope in the SearchForm component to fix this other bug: 
https://jira.nypl.org/browse/SCC-4095


src/utils/searchUtils.ts
```diff
@@ -259,6 +260,29 @@ export const sortOptions: Record<string, string> = {
   date_desc: "Date (New to Old)",
 }
 
+/**
+ * mapQueryFiltersToSearchFilters
+ * Maps the filter query params to a SearchFilters object
+ */
+export function mapQueryFiltersToSearchFilters(
```
charmingduchess, line 267: This method is fine, but there [already is a method that does exactly this](https://github.com/NYPL/research-catalog/blob/15b7fadeda8c299d898a3348f54663975e115997/src/utils/refineSearchUtils.ts#L11). I don't care which one ends up sticking around, but we should only have one method that converts filter params to an object. Can you pick one and replace the instances of the other with it?

charmingduchess (changes requested):

src/utils/searchUtils.ts
EdwinGuzman: I am leaning towards using the existing function since it's already used elsewhere.

Because we want to try to get this out to QA by next week when Diego and Chris are back, I can update this PR to use the existing function. Sounds good, @charmingduchess ?

EdwinGuzman (approved):

src/utils/searchUtils.ts
charmingduchess: oh dang i totally missed this comment. my bad!

src/types/searchTypes.ts
```diff
@@ -80,20 +80,19 @@ export interface SearchFormAction {
   payload: SearchParams | SearchFilters | string | string[]
 }
 
+export type SearchQueryFilters = [string, string][]
```
charmingduchess, line 83: this is totally a nit, but I think this would be more accurately represented by Record<string, string>.

charmingduchess:
I think this PR could use one more once over to iron out places where collapseMultiValueQueryParams is being called on its own. If the filters are passed in from the getServerSideProps, which totally makes sense, we shouldn't be doing this calculation again in `RefineSearch.tsx`, `SearchForm.tsx` and `AppliedFilters.tsx`. I could see it becoming unclear down the line why it's happening in differently places redundantly. I'd be happy to do that if that sounds like a drag.

PR #180, opened 2024-04-23
SCC-4094 - DRB results not appearing on author searches

dgcohen:
Ticket:
https://jira.nypl.org/browse/SCC-4094

Turns out the search scope (aka "field") wasn't being mapped to the DRB equivalent on Research Catalog, as it is on Discovery Front End via the mapSearchScope map.


EdwinGuzman (approved):

EdwinGuzman (commented):
Note: This is missing the unique ID on the DRB results container task that is defined in the ticket.

charmingduchess (approved):

PR #182, opened 2024-04-24
SCC-4010: FIX Moving all serverside logs to logger

7emansell:
Missed some of the serverside logs in [original logging PR](https://github.com/NYPL/research-catalog/pull/176) so just adding them here.


EdwinGuzman (commented):
Some places must keep using console because the functions are called on both the server and the client. For some functions it makes sense but I'm not entirely sure how next bundles these files to know when it's safe to use logger or console. A very similar problem to what was encountered in DCF.

EdwinGuzman (approved):

charmingduchess (approved):

PR #184, opened 2024-04-24
SCC-4096: Fixing Date label orientation on mobile

EdwinGuzman:
Fixes Ticket [SCC-4096](https://jira.nypl.org/browse/SCC-4096)

- This, somehow, is only really apparent when viewing the Refine Search in mobile. The "Date" label displays vertically -- see screenshot in the ticket.
- I _think_ the grid layout was more forceful on mobile than on desktop so it rendered as two columns rather than as a single row.
- Also deleted the `FieldsetDate` component since that is not being used.


charmingduchess:
It looks good to me, but I'm also not seeing the issue on QA or in main. Can you share how to reproduce the original issue?

charmingduchess (approved):

PR #186, opened 2024-04-25
SCC-4093: Updating the search results display text for select dropdown values

EdwinGuzman:
Fixes [SCC-4093](https://jira.nypl.org/browse/SCC-4093)

The search results display string function generally works but journal title and standard numbers were not being picked up when selected from the dropdown. Also, the `for keyword "spaghetti"` persisted when a select form value other than keyword or "All fields' was selected. 

Now, selecting "journal title" for example, will display `for journal title "spaghetti"`.

The existing function was refactored and existing tests still pass and added new tests.


jest.setup.js
```diff
@@ -18,7 +18,7 @@ jest.spyOn(global.console, "error").mockImplementation(() => jest.fn())
 jest.spyOn(global.console, "warn").mockImplementation(() => jest.fn())
 
 // Increase timeout on tests
-jest.setTimeout(35000)
+jest.setTimeout(40000)
```
charmingduchess, line 21: 😕

charmingduchess (commented):
I'm looking into your comment about the search scopes, but in the mean time i noticed this url `http://local.nypl.org:8080/research/research-catalog/search?q=cheese&search_scope=title`
displays `Displaying 1-50 of 979 results for undefined`

EdwinGuzman:
@charmingduchess that issue should be fixed now.

pages/search/index.tsx
```diff
@@ -62,6 +61,7 @@ export default function Search({
 
   // TODO: Move this to global context
   const searchParams = mapQueryToSearchParams(query)
+  console.log({ query, searchParams })
```
dgcohen, line 64: Remove this console.log

src/utils/searchUtils.ts
```diff
@@ -287,6 +302,11 @@ export function mapQueryToSearchParams({
     contributor,
     title,
     subject,
+    // TODO: this is a "catch-all" for journal title and standard number
```
dgcohen, line 305: I like the approach of using a catch all here, though I'm not sure if we'll need more explicit typing down the road. Should be good for now though.

dgcohen (approved):
This looks good and works well on my local. Just remove the console log and this should be good to merge in.

src/utils/searchUtils.ts
```diff
@@ -284,6 +299,11 @@ export function mapQueryToSearchParams({
     q,
     field: search_scope,
     page: page ? parseInt(page) : 1,
+    // TODO: this is a "catch-all" for journal title and standard number
```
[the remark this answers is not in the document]
nonword: `?q=cheese&search_scope={SCOPE}` is secretly equivalent to `?{SCOPE}=cheese` for `SCOPE` = title, contributor, and subject. But that's not a guarantee in the future. It's possible we'll want to handle advanced search queries (e.g. searches with `title` and `contributor`) differently from "search-scoped" queries. So I agree I would avoid confusing advanced search terms with search-scoped keyword queries - particularly given that we're holding on to `search_scope` anyway (in the form of "field")

EdwinGuzman:
Offline update: Vera and I are still wary of how we are using fields and the search_scope value. It's not a blocker for this update but we will revisit this soon.

PR #189, opened 2024-04-26
SCC-4081: Failed cancel/freeze holds error handling

7emansell:
Resolves [SCC-4081](https://jira.nypl.org/browse/SCC-4081). Adds failure modal to Freeze button and Cancel button on the Requests tab.

### DIFF
--- a/src/components/MyAccount/RequestsTab/CancelButton.tsx
+++ b/src/components/MyAccount/RequestsTab/CancelButton.tsx
@@ -52,6 +52,32 @@ const CancelButton = ({
     }
   }
 
+  function failureModalProps(hold) {
+    return {
+      bodyContent: (
+        <Box className={styles.modalBody}>
+          <Text sx={{ marginLeft: "l", marginRight: "m" }}>
+            Your request for{" "}
+            <span style={{ fontWeight: "var(--nypl-fontWeights-medium)" }}>
+              {hold.title}
+            </span>{" "}
+            has not been canceled. Please try again.
+          </Text>
+        </Box>
+      ),
+      closeButtonLabel: "OK",
+      headingText: (
+        <Heading className={styles.modalHeading}>
+          <>
+            <Icon size="large" name="errorFilled" color="ui.error.primary" />
+            <Text sx={{ marginBottom: 0 }}>Failed to cancel request </Text>
+          </>
+        </Heading>
+      ),
+      onClose: closeModal(),
+    }
+  }
+
   function checkModalProps(hold) {
     return {
       bodyContent: (
@@ -93,6 +119,8 @@ const CancelButton = ({
           if (response.status == 200) {
             // Open next modal to confirm request has been canceled.
             setModalProps(confirmModalProps(hold))
+          } else {
+            setModalProps(failureModalProps)
           }
         } else {
           closeModal()
--- a/src/components/MyAccount/RequestsTab/FreezeButton.tsx
+++ b/src/components/MyAccount/RequestsTab/FreezeButton.tsx
@@ -1,15 +1,47 @@
 import { useState } from "react"
 import type { Hold, Patron } from "../../../types/myAccountTypes"
-import { Button } from "@nypl/design-system-react-components"
+import {
+  Box,
+  Button,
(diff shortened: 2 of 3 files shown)


src/components/MyAccount/RequestsTab/CancelButton.tsx
```diff
@@ -52,6 +52,32 @@ const CancelButton = ({
...
+            </span>{" "}
+            has not been canceled. Please try again.
+          </Text>
+        </Box>
+      ),
+      closeButtonLabel: "OK",
+      headingText: (
+        <Heading className={styles.modalHeading}>
+          <>
+            <Icon size="large" name="errorFilled" color="ui.error.primary" />
```
7emansell, line 72: not a fan of how the icon is the same for whether the cancel action has failed or succeeded (I guess bc cancel = bad even if that's what you're trying to do?) I think one of these should be changed to the info icon

charmingduchess: i don't think that cancelling a hold is negative in the sense of the design system. you should use the same icons as the other success/failure modals.

src/components/MyAccount/RequestsTab/FreezeButton.tsx
```diff
@@ -1,15 +1,47 @@
...
+      <Heading className={styles.modalHeading}>
+        <>
+          <Icon size="large" name="errorFilled" color="ui.error.primary" />
+          <Text sx={{ marginBottom: 0 }}> Freezing this hold failed </Text>
+        </>
+      </Heading>
+    ),
+    onClose: () => {
+      closeModal()
+      setFrozen(!frozen)
```
EdwinGuzman, line 36: When updating a variable based on its current value, use the callback function syntax:
```
setFrozen((prevValue) => !prevValue)
```

This is mostly to cover any race conditions or rendering issues since the setState function is asynchronous.

src/components/MyAccount/RequestsTab/FreezeButton.tsx
```diff
@@ -1,15 +1,47 @@
...
+          Please try again.
+        </Text>
+      </Box>
+    ),
+    closeButtonLabel: "OK",
+    headingText: (
+      <Heading className={styles.modalHeading}>
+        <>
+          <Icon size="large" name="errorFilled" color="ui.error.primary" />
+          <Text sx={{ marginBottom: 0 }}> Freezing this hold failed </Text>
```
EdwinGuzman, line 30: There's extra spacing around the text.

src/components/MyAccount/RequestsTab/FreezeButton.tsx
```diff
@@ -1,15 +1,47 @@
...
+        </>
+      </Heading>
+    ),
+    onClose: () => {
+      closeModal()
+      setFrozen((frozen) => !frozen)
+      setIsDisabled(false)
+    },
```
EdwinGuzman, line 38: This may need a bit more work. I tested this locally by removing the conditional mentioned in the description. Then I intentionally messed up the fetch request url in the `handleFreezeClick` function. Because it's wrong, the `response.status` is 404 (I removed "a" from `.../holds/update/...`).

When I click to freeze the hold, the modal popups correctly and shows the error message. But when I click "OK", the button text reads "Unfreeze". I think it's happening here where setFrozen is happening optimistically.

7emansell: Hmm am I missing some step in the logic here or is it just that when the **failure** modal closes, the frozen state of the button should never change? So this line should just be removed period

EdwinGuzman: yeah, I think it's basically just remove this line but test it in case there are scenarios I did not test (I only really tested the scenario I wrote)

7emansell: Ok- removed that line and made the test of this case more specific

EdwinGuzman (approved):

charmingduchess (approved):
Looks good!

PR #190, opened 2024-04-29
SCC-4061: Adding global context for search query params, only storing q for now

EdwinGuzman:
## Ticket:

- JIRA ticket [SCC-4061](https://jira.nypl.org/browse/SCC-4061)

- Replaces the `searchKeywords="TODO"` with the `q` search url query param.
- I added a global context and attempted to store _all_ query params as an object but it causes an infinite loop. I'm not sure why and I'm trying to debug this issue but for now, we don't need all the values in the object. That is why I only passed `q`.
- The other option is to prop drill the search query value from `Search` to `SearchResult` to `ItemTable` to `RequestButtons`, which is two more levels down than what I would like.

## How has this been tested?

Locally, the request button URLs should have `?searchKeyworks=...` in the url. This won't work in the app because the hold pages do not exist (regular and EDD). This will work in the reverse proxy state.

- N/A

### Checklist:


dgcohen (approved):
This looks great! Not sure whether we'll need params other than q in the context down the road, but I think we can worry about the infinite loop issue when/if we do.

EdwinGuzman:
Not yet but it's not urgent and can wait post release.

EdwinGuzman:
I also still want to figure out the proper (or better) way to handle the infinite loop with context and state.

As far as I got: the update in this PR, using just the string, works because it's a primitive so React's logic check on every render such as `"spaghetti" === "spaghetti"` makes it render only when the string is updated.

React does a shallow copy of objects, however, which follows the same js logic:

```
const a = { q: "spaghetti"  }
const b = { q: "spaghetti" }

a === b // false
```

So because of this, for a searchParams object such as 

```
{
  q: "spaghetti",
  search_scope: "title",
  subject: undefined,
  ...
}
```
even if the next object is the same on the next render, React thinks it's different so it'll trigger another rerender. I'm pretty sure this is what caused the rerender. I can try wrapping it in a useMemo but tbh that's annoying and seems wrong. So I'll take another look but we should also be clear on what we want in the global context.

charmingduchess:
I get that it's not urgent, but if you're going to have to do extra work to support this for release, isn't it worth confirming that we need it?

EdwinGuzman:
> I get that it's not urgent, but if you're going to have to do extra work to support this for release, isn't it worth confirming that we need it?

I think we now need it for the "Back to search results" link on the bib page. Otherwise, the bib page won't know that it comes from the search page.

charmingduchess:
Are we using this PR? @EdwinGuzman

PR #193, opened 2024-05-03
WIP SCC-4108: Fix how url query params are read when making searches AND displaying …

EdwinGuzman:
…the creatorLiteral filter in the search display string

## Ticket:

- JIRA ticket [xxx](url)

-

-

### Checklist:

### DIFF
--- a/pages/search/index.tsx
+++ b/pages/search/index.tsx
@@ -7,7 +7,6 @@ import {
 } from "@nypl/design-system-react-components"
 import { useEffect, useRef, type ChangeEvent } from "react"
 import { useRouter } from "next/router"
-import { parse } from "qs"
 
 import Layout from "../../src/components/Layout/Layout"
 import DRBContainer from "../../src/components/DRB/DRBContainer"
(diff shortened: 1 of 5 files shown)


pages/search/index.tsx
```diff
@@ -235,12 +234,10 @@ export default function Search({
  * relevant search results on the server side (via fetchResults).
  *
  */
-export async function getServerSideProps({ resolvedUrl, req }) {
+export async function getServerSideProps({ resolvedUrl, req, query }) {
   const bannerNotification = process.env.SEARCH_RESULTS_NOTIFICATION || ""
 
-  // Remove everything before the query string delineator '?', necessary for correctly parsing the 'q' param.
-  const queryString = resolvedUrl.slice(resolvedUrl.indexOf("?") + 1)
-  const results = await fetchResults(mapQueryToSearchParams(parse(queryString)))
```
EdwinGuzman, line 243: The `qs` package was introduced to help parse the request URL query which is helpful but the `collapseMultiValueQueryParams` function (which is called by `mapQueryToSearchParams`)  does not expect a fully parsed object.

The first object below is the type of keys it expects when it parses the queryParams. The second object is what `qs`'s "parse' function returns.

I decided to _not_ use the "parse" function because nextjs returns the same object in `getServerSideProps` and through the `useRouter` function. Additionally, "parse" cannot be used for the object that `useRouter` returns.

Screenshot 2024-05-03 at 11 11 24 AM

dgcohen (approved):
This looks good  to me. If restructuring the filters as set in Next router's query params works and there are eventually adequate integration tests, I don't see a reason why we would need to keep the qs library, and I'm always happy to get rid of a dependency.

PR #200, opened 2024-05-09
HOTFIX - Applied Filters bug on Production

dgcohen:

- This is a Hotfix for a 500 error on production that occurs when applied filters with undefined attributes are passed into buildTagsetData. The root cause of the undefined filter attributes should be found, but this will filter those values out for now preventing the page from breaking.
- For now I am returning null in the map for filter fields with undefined values, which i then have to filter out before finally flattening the output of buildTagsetData.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2312,6 +2312,21 @@
         "glob": "7.1.7"
       }
     },
+    "node_modules/@next/swc-darwin-arm64": {
+      "version": "13.4.19",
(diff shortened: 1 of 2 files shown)


EdwinGuzman (approved):
Great, this works for me.

dgcohen:
Closing since this will be fixed at the root with this PR: https://github.com/NYPL/research-catalog/pull/201

PR #201, opened 2024-05-09
Subject literal applied filter bug

charmingduchess:
## Ticket:

Fixes bug where this url http://local.nypl.org:8080/research/research-catalog/search?q=shakespeare&filters[subjectLiteral][0]=Shakespeare%2C+William%2C+1564-1616+--+Characters.&filters[materialType][0]=resourcetypes%3Anot was leading to a 500 error.

- adds special handling for subjectLiteral filters. Subject literals can have multiple facets, and a user can select all or some of those to search by from the bib page. the aggregations results, however, only return the combined facets. The code was returning undefined because the aggregations response had `shakespeare -- characters -- songs and music`, but not `shakespeare -- characters`, like the search param did. there was no match, so `addLabelPropAndParseFilters` was returning undefined for subjectLiteral in this case, causing problems down the line.
- this solution keeps the subject literal tag visible and removable, instead of hiding it.

## How has this been tested?

Accessing the url above. 

-

### Checklist:

- [x ] I updated the CHANGELOG with the appropriate information and JIRA ticket number (if applicable).
- [x ] I have added relevant accessibility documentation for this pull request.


src/components/SearchFilters/appliedFilterUtils.ts
```diff
@@ -31,6 +32,56 @@ export const buildAppliedFiltersValueArrayWithTagRemoved = (
...
+      // aggregations results. Pass the year along with out
+      // transforming fieldname or finding the label
+      if (appliedFilterField.includes("date")) {
+        const labelPrefix = appliedFilterField.split("date")[1]
+        return {
+          count: null,
+          value: filterValue,
+          label: `${labelPrefix} ${filterValue}`,
+        }
+      }
```
charmingduchess, line 67: I moved this method, so the diff is not easily apparent, but the code I added is here, line 68-73

dgcohen (approved):

EdwinGuzman (approved):
Thanks for looking into this as well!

PR #206, opened 2024-05-10
hotfix Fix missing aggregation applied filter bug

charmingduchess:
You can verify with this url:
https://research-catalog-g6mc0u7vi-nypl.vercel.app/research/research-catalog/search?q=spaghetti&filters%5BmaterialType%5D%5B0%5D=resourcetypes%3Amix&filters%5BmaterialType%5D%5B1%5D=resourcetypes%3Atxt&filters%5Blanguage%5D%5B0%5D=lang%3Aeng&filters%5Blanguage%5D%5B1%5D=lang%3Adut&filters%5BsubjectLiteral%5D%5B0%5D=Cooking+%28Pasta%29

or choose your own combination of mixed material, another material type with a much higher count, and a sprinkling of other fields.

This query includes resourcetypes:mix, but the aggregations, and hence the applied filter tag set, do not.


EdwinGuzman (approved):
Looks good, but a bit confused on the different between this and the hotfix that was already merged into `production`. I would have imagined that very little should show up as new changes.

In any case, don't merge it today (Friday) and we'll work on deploying this next week. Thank you!

charmingduchess:
@EdwinGuzman I'm not sure why the diff looks bigger than i think it needs to be. The actual change is only line 87.

dgcohen (approved):

PR #207, opened 2024-05-10
SCC-4099: Account redirect refactor

7emansell:
## Ticket:

Resolves JIRA spike [SCC-4099](https://jira.nypl.org/browse/SCC-4099)

- Tried some refactoring of how data is fetched and how redirect logic is phrased
- Found that I couldn't do any better performance-wise :)
- Leaves mildly refactored refactor logic

## How has this been tested?

Locally, using `console.time` and `console.timeLog`

### Checklist:


EdwinGuzman (approved):
The minor refactor looks good. I think as we discussed before, the API call to fetch the patron data gets called twice, once to get all the data and again if there was a redirect.

Let's leave this but after this phase, there can be a spike to see if a client-side redirect is any better.

PR #213, opened 2024-05-16
SCC-4074: My Account phone/email validation

7emansell:
## Ticket:

Resolves JIRA ticket [SCC-4074](https://jira.nypl.org/browse/SCC-4074)

- Adds form validation to phone numbers and emails

## How has this been tested?

Locally– try setting notification preference to Phone/Email then submitting an empty string for the respective field. 

### Checklist:


src/components/MyAccount/Settings/AccountSettingsDisplayOptions.tsx
```diff
@@ -29,7 +30,50 @@ export const AccountSettingsDisplay = ({ patron }: { patron: Patron }) => {
...
+}: {
+  patron: Patron
+  onValidateForm: (any) => void
+}) => {
+  const [formData, setFormData] = useState({
+    primaryPhone: patron.phones[0]?.number,
+    email: patron.emails[0],
+  })
+
+  const validateForm = useCallback(() => {
```
charmingduchess, line 45: Hmmm I'm on the fence about this pattern. My first instinct was that this logic can be in the `useEffect` below, but I also feel like it's tidy to have it in a method. Something feels redundant in the fact that both hooks have `formData` as a dependency and then the `useEffect` has `validateForm` as a dependency.
@EdwinGuzman what do you think?

whether this ends up as a method or something else, can you change the name to something like `isFormValid`?

charmingduchess (changes requested):
Looks good! Thanks for rolling with the punches on the AC here.

src/components/MyAccount/Settings/AccountSettingsTab.tsx
```diff
@@ -27,12 +28,72 @@ const AccountSettingsTab = ({ settingsData }: { settingsData: Patron }) => {
   const [mostRecentPatronData, setMostRecentPatronData] = useState(settingsData)
   const [modalProps, setModalProps] = useState(null)
   const [isLoading, setIsLoading] = useState(false)
+
+  const { onOpen: openModal, onClose: closeModal, Modal } = useModal()
+
+  const successModalProps = {
```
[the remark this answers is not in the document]
7emansell: the `useModal` function handlers were declared in this file, and the modal props declared in another file weren't calling an imported `onClose` . It seemed to make more sense and also better follow DS patterns to just have them together

src/components/MyAccount/Settings/AccountSettingsDisplayOptions.tsx
```diff
@@ -29,7 +30,50 @@ export const AccountSettingsDisplay = ({ patron }: { patron: Patron }) => {
...
+        formData.primaryPhone !== "" && phoneRegex.test(formData.primaryPhone)
+      )
+    } else if (patron.notificationPreference == "Email") {
+      return formData.email !== ""
+    } else return true
+  }, [formData])
+
+  useEffect(() => {
+    setIsFormValid(isFormValid())
+  }, [formData, isFormValid, setIsFormValid])
```
charmingduchess, line 58: Does `formData` need to be in this dependency array.

src/components/MyAccount/Settings/AccountSettingsDisplayOptions.tsx
```diff
@@ -29,7 +30,50 @@ export const AccountSettingsDisplay = ({ patron }: { patron: Patron }) => {
...
+  patron: Patron
+  setIsFormValid: Dispatch<React.SetStateAction<boolean>>
+}) => {
+  const [formData, setFormData] = useState({
+    primaryPhone: patron.phones[0]?.number,
+    email: patron.emails[0],
+  })
+
+  const isFormValid = useCallback(() => {
+    const phoneRegex = /^\d+$/
```
charmingduchess, line 46: This regex would exclude a number formatted like 123-456-7890. I think we should allow people to submit in that format, because it is easier to see when you're typing it out.

7emansell: Allow them to submit in that format but then store without the dashes ?

7emansell: @charmingduchess most recent fix allows dashes, but kind of opens up a whole can of worms with validation (for example, now someone can just submit a bunch of dashes, which will save as the empty string. We would need to require a certain number of numbers in the input to block that) so lmk what you think

src/components/MyAccount/Settings/AccountSettingsDisplayOptions.tsx
```diff
@@ -29,7 +30,50 @@ export const AccountSettingsDisplay = ({ patron }: { patron: Patron }) => {
...
+  patron: Patron
+  setIsFormValid: Dispatch<React.SetStateAction<boolean>>
+}) => {
+  const [formData, setFormData] = useState({
+    primaryPhone: patron.phones[0]?.number,
+    email: patron.emails[0],
+  })
+
+  const isFormValid = useCallback(() => {
+    const phoneRegex = /^[0-9-]+$/
```
charmingduchess, line 46: hm okay. how about we strip out non number characters and verify that the leftover string is 10 digits.

charmingduchess: alternately, you can use this regex `(\d{3})[^0-9]{1,2}?(\d{3})[^0-9]?(\d{4})` which matches these variations:
123-456-7890
1234567890
(123) 456-7890
1 (234) 567 8900

charmingduchess: Actually I think the move is to have place holder text like this:
`000 000 0000` 
And then verify that someone has submitted 10 digits minus whitespace.

src/components/MyAccount/Settings/AccountSettingsDisplayOptions.tsx
```diff
@@ -100,6 +144,8 @@ export const AccountSettingsForm = ({ patron }: { patron: Patron }) => {
               id="phone-text-input"
               labelText="Update phone number"
               showLabel={false}
+              onChange={handleInputChange}
+              placeholder="000-000-0000"
```
charmingduchess, line 148: I'm not seeing the place holder show up when I delete the input value. How are you seeing it?

7emansell: Hmm I am. When the text input is entirely cleared there's nothing?

charmingduchess: i didn't have the latest changes 🤦🏻‍♀️ it's working great!

src/components/MyAccount/ProfileTabs.tsx
```diff
@@ -37,7 +37,9 @@ const ProfileTabs = ({
       content: checkouts ? (
         <CheckoutsTab checkouts={checkouts} patron={patron} />
       ) : (
-        <Text>There was an error accessing your checkouts.</Text>
+        <Text sx={{ marginTop: "s" }}>
```
charmingduchess, line 40: Diego mentioned a preference for the css as prop pattern ie `mt="s"`. Can you update this PR with that?

PR #220, opened 2024-05-22
SCC-4131: Fixing accessibilities issues on My Account tabs

EdwinGuzman:
## Ticket:

- JIRA ticket [SCC-4131](https://jira.nypl.org/browse/SCC-4131)

- Fixes four issues found in the accessibility review for the My Account Checkouts and Requests tabs.
- Adds item titles to various buttons' `aria-label` attribute, updates the link placement in a sentence, and updates the focus management for the Freeze/Unfreeze button.

## How has this been tested?

Locally.

- Fixes issues found.

### Checklist:

### DIFF
--- a/src/components/MyAccount/RequestsTab/FreezeButton.tsx
+++ b/src/components/MyAccount/RequestsTab/FreezeButton.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react"
+import { createRef, useEffect, useState } from "react"
 import type { Hold, Patron } from "../../../types/myAccountTypes"
 import {
   Box,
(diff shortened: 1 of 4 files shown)


src/components/MyAccount/RequestsTab/FreezeButton.tsx
```diff
@@ -38,6 +39,12 @@ const FreezeButton = ({ hold, patron }: { hold: Hold; patron: Patron }) => {
     },
   }
 
+  useEffect(() => {
+    if (isDisabled === false) {
```
EdwinGuzman, line 43: This only really works when the initial state for `isDisabled` is `null`. It's then set to either true or false but the initial `null` value helps _not_ focus on the button when the page loads.

7emansell (approved):

src/components/MyAccount/RequestsTab/FreezeButton.tsx
charmingduchess: Could this focus invocation happen after line 69?

charmingduchess (changes requested):

src/components/MyAccount/RequestsTab/FreezeButton.tsx
EdwinGuzman: I tried it. It didn't work. I think it's because when the function is declared, the `buttonRef` is still null. So unfortunately it has to be in an `useEffect`.

PR #221, opened 2024-05-22
SCC-4137 - Fix title search bug on main search form

dgcohen:
## Ticket:

- JIRA ticket [SCC-4137](https://jira.nypl.org/browse/SCC-4137)

- Adds the logic derrived from DFE that removes advanced search query parameters when the search_scope is "all" (main search with "All" selected or Advanced Search.

## How has this been tested?

I've tested it locally and it has been tested by Sean, Nicole and Edwin on Vercel

### DIFF
--- a/src/utils/utilsTests/searchUtils.test.ts
+++ b/src/utils/utilsTests/searchUtils.test.ts
@@ -51,6 +51,53 @@ describe("searchUtils", () => {
         })
       ).toBe(true)
     })
+    it("includes advanced search query params when field is set to 'all'", () => {
+      const testQuery =
+        "?q=shel%20silverstein&contributor=shel silverstein&title=the giving tree&subject=books"
+      expect(
+        checkQueryParamsEquality(testQuery, {
+          q: "shel silverstein",
+          title: "the giving tree",
+          contributor: "shel silverstein",
+          subject: "books",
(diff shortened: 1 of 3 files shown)


src/utils/utilsTests/searchUtils.test.ts
```diff
@@ -51,6 +51,53 @@ describe("searchUtils", () => {
...
+      ).toBe(true)
+    })
+    it("clears advanced search query params when field param is anything other than 'all'", () => {
+      const titleQuery = "?q=shel%20silverstein&search_scope=title"
+      expect(
+        checkQueryParamsEquality(titleQuery, {
+          q: "shel silverstein",
+          title: "the giving tree",
+          contributor: "shel silverstein",
+          subject: "books",
```
EdwinGuzman, line 74: I'm confused. I thought that when there's a field other than "all", then title, contributor, and subject would be clear and not in the returned query string?

EdwinGuzman: Ah, never mind. I was not reading the helper test function correctly.

EdwinGuzman (approved):

dgcohen:
Closing because we will be releasing this as a hotfix

PR #224, opened 2024-05-22
SCC-4133: My Account general VQA updates

EdwinGuzman:
## Ticket:

- JIRA ticket [SCC-4133](https://jira.nypl.org/browse/SCC-4133)

- Fixes VQA comments for the profile header and general page. See the ticket for more and the related QA doc for more questions and comments.

## How has this been tested?

Locally.
Screenshot 2024-05-22 at 5 26 12 PM

- N/A

### Checklist:


dgcohen (approved):
Looks good on a first glance. Just missing a description of the updates in the changelog.

EdwinGuzman:
Ah, I did not update the changelog because these are fixes and not new features that are not yet deployed.

src/models/MyAccount.ts
```diff
@@ -249,9 +249,13 @@ export default class MyAccount {
         notificationPreferenceMap[patron.fixedFields["268"].value]
       return {
         notificationPreference,
-        name: patron.names[0],
+        name: MyAccount.formatPatronName(patron.names[0]),
```
charmingduchess, line 252: When we rolled out the holdshelf emails we opted to leave patron names as LAST, FIRST because it's really hard to parse names automatically (I'm sure you've seen [this blog post](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/)). I imagine we would want to do the same here?

EdwinGuzman: Yes, I was definitely worried about this. Can you check the thread regarding this in the QA doc? https://docs.google.com/document/d/1rx8pHljNYFhItXtpEjaKPB9MHtqTfdxeEs7Nw8QpEuM/edit

I got the impression that it was possible based on confirmation emails that are sent. Can we add a middle name and second last name to a test patron to check this update?

PR #227, opened 2024-05-28
Scc 3930/fetch pickup locations

charmingduchess:
## Ticket:

- JIRA ticket [SCC-3930](https://jira.nypl.org/browse/SCC-3930)

- adds the ability to update the pickup location of an item.
 - clicking change location triggers a modal with a select containing possible locations to change to
 - clicking cancel returns to original state
 - clicking confirm triggers skeleton loader until response is returned
 - location is updated in local state to newly selected location

## How has this been tested?

Manual testing in My Account running locally. Updating locations and refreshing the page to make sure the data actually made it to Sierra.

-

### Checklist:


src/components/MyAccount/RequestsTab/UpdateLocation.tsx
```diff
@@ -0,0 +1,196 @@
...
+    ),
+    closeButtonLabel: "Cancel location change",
+    confirmButtonLabel: "Confirm location",
+    onCancel: () => {
+      closeModal()
+    },
+    onConfirm: async () => {
+      setModalProps({
+        ...confirmLocationChangeModalProps(selected),
+        bodyContent: <SkeletonLoader showImage={false} />,
```
7emansell, line 87: what do you think about the heading also being set to have the skeleton loader?

src/components/MyAccount/RequestsTab/UpdateLocation.tsx
```diff
@@ -0,0 +1,196 @@
...
+      (loc) => loc.code.trim() !== pickupLocation.code.trim()
+    ),
+  ]
+  const confirmLocationChangeModalProps = (selected: SierraCodeName) => ({
+    type: "confirmation",
+    bodyContent: (
+      <Box className={styles.modalBody}>
+        <Select
+          value={selected.code}
+          onChange={(e: React.FormEvent<HTMLInputElement>) => {
```
7emansell, line 51: Does this make the whole modal flash on change bc of the Button Group error ?

src/components/MyAccount/RequestsTab/UpdateLocation.tsx
```diff
@@ -0,0 +1,196 @@
...
+    <>
+      <Button
+        sx={{ paddingLeft: 0 }}
+        size="small"
+        onClick={openModal}
+        id={`update-pickup-location-${key}`}
+        buttonType="text"
+      >
+        <Icon name="editorMode" align="left" size="small"></Icon>
+        <Text className={styles.changeLocation}>Change location</Text>
```
7emansell, line 190: nit but I think the text of this button needs to be bigger

7emansell (commented):
Works great just a couple visual questions

src/components/MyAccount/RequestsTab/RequestsTab.tsx
```diff
@@ -5,12 +5,16 @@ import type { Hold, Patron } from "../../../types/myAccountTypes"
 import ItemsTab from "../ItemsTab"
 import CancelButton from "./CancelButton"
 import FreezeButton from "./FreezeButton"
+import UpdateLocation from "./UpdateLocation"
+import { filteredPickupLocations as pickupLocations } from "../../../../__test__/fixtures/processedMyAccountData"
```
dgcohen, line 9: The should probably be in constants.ts instead of fixtures if they're being used outside of tests.

pages/api/account/holds/update/[id].ts
```diff
@@ -19,16 +19,17 @@ export default async function handler(
     return res.status(responseStatus).json(responseMessage)
   }
 
-  if (req.method !== "POST") {
-    responseMessage = "Please make a POST request to this endpoint."
+  if (req.method !== "PUT") {
```
Comment (dgcohen) [inline pages/api/account/holds/update/[id].ts:22]: Purely out of curiosity, why did the method need to be changed?

src/components/MyAccount/RequestsTab/UpdateLocation.tsx
```diff
@@ -0,0 +1,196 @@
...
+      )
+      if (response.status == 200) {
+        // Open next modal to confirm request has been canceled.
+        setModalProps(successModalProps(selected) as DefaultModalProps)
+      } else setModalProps(failureModalProps as DefaultModalProps)
+    },
+    headingText: (
+      <Heading className={styles.modalHeading}>
+        <Text sx={{ marginBottom: 0 }}>
+          {" "}
```
dgcohen, line 107: Are the whitespace characters here necessary?

src/components/MyAccount/RequestsTab/UpdateLocation.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  const [modalProps, setModalProps] = useState<BaseModalProps>(
+    confirmLocationChangeModalProps(pickupLocation) as ConfirmationModalProps
+  )
+
+  const successModalProps = (newLocation: SierraCodeName) => ({
+    type: "default",
+    bodyContent: (
+      <Box className={styles.modalBody}>
+        <Text sx={{ marginLeft: "l" }}>
```
dgcohen, line 122: Very small nit, i slightly prefer the margin props ("ml", "mb" etc) since they're available and a bit more concise but i'll leave that up to you.

src/components/MyAccount/RequestsTab/UpdateLocation.tsx
```diff
@@ -0,0 +1,196 @@
...
+          Your item will be available for pickup at the {newLocation.name}{" "}
+          Library.
+        </Text>
+      </Box>
+    ),
+    closeButtonLabel: "OK",
+    headingText: (
+      <Heading className={styles.modalHeading}>
+        <>
```
dgcohen, line 131: Don't think you need a fragment here since these are wrapped in a <Heading />

dgcohen (approved):
Looks good! Just a few minor comments.

pages/api/account/holds/update/[id].ts
Comment (charmingduchess) [reply pages/api/account/holds/update/[id].ts:22]: we had originally thought that we weren't able to make put requests from the front end, so this was written to listen for posts. but it turns out we can, so it makes sense for the request from the front end to match the one on the backend.

src/components/MyAccount/RequestsTab/UpdateLocation.tsx
charmingduchess: I'm pretty sure it's because the entire modal props has to be updated so the modal is rerendering. It's the same as when you click confirm and it flashes before showing you success or failure.

src/components/MyAccount/RequestsTab/UpdateLocation.tsx
charmingduchess: I think the heading is necessary to populate some aria attributes, so I don't want to remove it here.

src/components/MyAccount/RequestsTab/UpdateLocation.tsx
charmingduchess: Nope

src/components/MyAccount/RequestsTab/UpdateLocation.tsx
charmingduchess: I think the heading only takes one child, so it has to be wrapped. I could make the icon be a child of the text, but this pattern is repeated many times throughout my account. I'll make a separate ticket to make that adjustment and centralize the styles for those modals.

src/components/MyAccount/RequestsTab/RequestsTab.tsx
charmingduchess: Ah yes. That was fixed elsewhere but this was cut from a branch before that. This data actually isn't going to be constant when it's deployed. I'm going to leave it in `src/utils/accountUtils` because thats where it is in main right now.

src/components/MyAccount/RequestsTab/UpdateLocation.tsx
charmingduchess: update the size

PR #229, opened 2024-05-29
SCC-3964 - Render Item Table in Bib Details

dgcohen:
## Ticket:

- JIRA ticket [SCC-3964](https://jira.nypl.org/browse/SCC-3964)

- Adds the Bib model parent class that includes shared functionality between Search Results and Bib Page.
- SearchResultsBib now extends Bib class.
- Renders ItemTable component on Bib Page without pagination, filters, or View All Link (upcoming).
- Renames 'bib' to 'bibResult' when the type matches Discovery API response.
- Removes ItemLocationEndpoint type and replaces it with a string type since it was of little value and was forcing us to frequently type cast bib fixture data.
- Adds the classnames package to allow for easy conditional setting of classes for styling purposes. Used here to style the Item Table differently in Search Results than on the Bib Page.
- Removes the breakpoint specific rendering of Item Table data as we will no longer be selectively hiding content on mobile.
- Changed isBibPage to inSearchResult since I found most of the overrides to be relevant to the search results version of the item table. This is a stylistic preference so let me know if you prefer the isBibPage setting.

## How has this been tested?

- Added tests for Bib model, SearchResultsBib model, and ItemTable in Bib Page.

## Concerns

- Note that the holdingsDetails section of the Bib page is currently not rendering so I have not added tests for this section yet. I believe this issue is coming from the BibDetails model. I'll open a ticket to address this.

### Checklist:


charmingduchess:
`SearchResultsBib` and `BibResult` have very similar names. Can the types indicate somehow that they are the model type vs the untransformed Discovery API response?

charmingduchess (commented):
Overall looks great. 
I think there needs to be some logic, potentially in the item table model, to determine which properties are displayed on the bib page in mobile and tablet view. We don't display all of them for mobile.

src/components/ItemTable/ItemTable.tsx
```diff
@@ -1,10 +1,12 @@
 import { Box, Table } from "@nypl/design-system-react-components"
+import classNames from "classnames/bind"
 
 import type ItemTableData from "../../models/ItemTableData"
-import RequestButtons from "./RequestButtons"
-import ItemAvailability from "./ItemAvailability"
+import StatusLinks from "./StatusLinks"
 import styles from "../../../styles/components/ItemTable.module.scss"
 
+const cx = classNames.bind(styles)
```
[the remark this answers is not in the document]
dgcohen: Yeah exactly. It just helps to remove the verbose concatenation.

src/components/SearchResults/SearchResult.tsx
```diff
@@ -26,19 +22,6 @@ interface SearchResultProps {
  * The SearchResult component displays a single search result element.
  */
 const SearchResult = ({ bib }: SearchResultProps) => {
-  const { isLargerThanLarge: isDesktop } = useNYPLBreakpoints()
```
EdwinGuzman, line 29: Great, if the app can rely less on `useNYPLBreakpoints` then there'll be fewer flashes when the desktop layout renders.

src/models/ItemTableData.ts
```diff
@@ -56,18 +54,18 @@ export default class ItemTableData {
...
   }
 
   showAccessColumn(): boolean {
-    return this.isBibPage && this.isDesktop
+    return !this.inSearchResult
   }
 
   volumeColumnHeading(): string {
-    return this.isArchiveCollection ? "Vol/Date" : "Container"
+    return this.isArchiveCollection ? "Container" : "Vol/Date"
```
EdwinGuzman, line 69: Why this update? Or where is it documented?

src/types/itemTypes.ts
```diff
@@ -20,16 +20,14 @@ export interface SearchResultsItem {
 export interface ItemLocation extends JSONLDValue {
   customerCode?: string
   prefLabel?: string
-  endpoint?: ItemLocationEndpoint
+  endpoint?: string
 }
 
 export type ItemLocationKey = "Schwarzman" | "Performing" | "Schomburg"
 
-export type ItemLocationEndpoint = "schwarzman" | "lpa" | "schomburg"
```
EdwinGuzman, line 28: Why was this removed?

EdwinGuzman: Hmm I see it was to reduce the type casting but it makes it clear what the values are.

styles/components/ItemTable.module.scss
```diff
@@ -8,13 +8,32 @@
...
-      // Meant to be 240 col width + 32 (gap) = 272px
-      width: 272px;
-      padding: 0;
+    td,
+    td:first-of-type,
+    th:first-of-type {
+      padding: var(--nypl-space-s) 0;
+
+      // Fix for extra whitespace added by non-table cell styling on span coming from DS
+      // TODO: Bring this issue up with DS team and potentially fix there
```
EdwinGuzman, line 18: Did you bring this up to Marty already?

EdwinGuzman (changes requested):
I need to go through this again but the `Bib` to `BibResults` update is a bit confusing. It's initially clear why, but then when I see it used in some components that I would not expect it to be used and that makes it confusing.

- The column order does not match production. In production, the order is Status, Vol/Date, Format, Access, Call Number, and Item Location. In this PR, the order is Status, Vol/Date, Format, Call Number, Access, Item Location.
- There's no space between some columns:
Screenshot 2024-06-04 at 11 05 14 AM

src/models/ItemTableData.ts
dgcohen: I change this because the values were the reverse in DFE:

`const volColumnHeading = isArchiveCollection ? archiveColumnText : volText;`

src/types/itemTypes.ts
dgcohen: I agree that it's good to have them clearly defined but thought it was a bit of a headache to have to add it on every fixture and elsewhere. The values are still enforced in the locationEndpointsMap as well as the key type (ItemLocationEndpoint) but if you'd rather have the values for the endpoints type enforced I can add that back in.

styles/components/ItemTable.module.scss
dgcohen: Not yet, I'll ask him about it.

__test__/pages/bib/bibPage.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    expect(screen.getByTestId("Published By")).toHaveTextContent(
+      "Mansfield, Ohio : Urban Spaghetti, [1999?-"
+    )
+  })
+
+  it("renders the bib page item table when there are physical items in the bib", () => {
+    expect(screen.getByTestId("bib-details-item-table")).toBeInTheDocument()
+  })
+
+  it("renders the bottom bib details", () => {
```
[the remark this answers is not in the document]
dgcohen: I agree, however those are auto generated by the DS

src/components/ItemTable/ItemTable.tsx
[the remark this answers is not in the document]
charmingduchess: Would you be down to remove this dependency? I feel like it's not such a complex string to generate, and it's not being updated dynamically.

__test__/pages/bib/bibPage.test.tsx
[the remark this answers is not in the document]
EdwinGuzman: Yes, but isn't that because of the value that is passed?

src/types/itemTypes.ts
EdwinGuzman: Can leave out for now

EdwinGuzman (approved):
Looks okay to me but there are questions from others that should be addressed before merging.

__test__/pages/bib/bibPage.test.tsx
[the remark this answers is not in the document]
dgcohen: You're right. We're passing the label as the data test id directly, we should address this in a separate task. Creating a ticket for it so it doesn't get lost in the shuffle.

charmingduchess (approved):

PR #231, opened 2024-06-03
SCC-4116 - DRB No Results

dgcohen:
## Ticket:

- JIRA ticket [SCC-4116](https://jira.nypl.org/browse/SCC-4116)

- Renders DRBContainer in Search Results regardless of totalWorks value.
- Adds the DRB no results copy and image based on DFE and renders when totalWorks is 0.
- Adds fileMock file with an empty export to fix failing tests caused by the static image import.
- Adds some copy/styling tweaks to DRBContainer in general as requested by Apoorva in VQA. 

## How has this been tested?

- I have updated the tests for DRBContainer to correctly test for the no results state.

### Checklist:

- [x] Visually QA'd by Apoorva
- [x] Accessibility QA'd by Clare


charmingduchess (approved):

CHANGELOG
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Prerelease
 
+### Updated
```
EdwinGuzman, line 10: This sounds more like an "Added" or "fixed" feature?

src/components/DRB/DRBContainer.tsx
```diff
@@ -39,38 +42,68 @@ const DRBContainer = ({
...
+              </>
+            </ExternalLink>
+          </>
+        ) : (
+          <>
+            <ExternalLink
+              href={appConfig.urls.drbAbout}
+              aria-label="Explore Digital Research Books Beta"
+            >
+              <Box mb="xs" bg="ui.white" p="s" borderRadius="5px">
```
EdwinGuzman, line 91: You shouldn't render block elements like `div`s inside anchor elements. So update this a span but you make need to make it `display: "inline-block"`.

src/components/DRB/DRBContainer.tsx
```diff
@@ -39,38 +42,68 @@ const DRBContainer = ({
...
+                <Image src={drbPromoImage} alt="" />
```
EdwinGuzman, line 92: Importing the image into the component and then passing it as a value feels wrong. If you put the image file in the `public` directory, then you should be able to put the URL directly here.

EdwinGuzman (changes requested):

pages/search/index.tsx
```diff
@@ -138,13 +138,13 @@ export default function Search({
             ) : null}
             {isLoading ? (
               <SkeletonLoader showImage={false} />
-            ) : drbResponse?.totalWorks > 0 ? (
+            ) : (
               <DRBContainer
                 drbResults={drbResults}
-                totalWorks={drbResponse.totalWorks}
+                totalWorks={drbResponse?.totalWorks || 0}
```
EdwinGuzman, line 144: Based on this and the tests, I think `DRBContainer` should itself be able to render the no results state as its default view. Right now, it depends on the parent passing in the value or `0` and the tests have it explicitly as `0`.

I think `DRBContainer` can set its `totalWorks` prop with an initial default value of `0`.

src/components/DRB/DRBContainer.tsx
dgcohen: The Next JS docs say that the Image component can take a statically loaded asset. It can also take an absolute path, but it'll change based on the environment and using a relative path requires the loader prop:

https://nextjs.org/docs/pages/api-reference/components/image

EdwinGuzman: We discussed this offline: we'll keep this as recommended by Next for local images, including the usage of Next's `Image` component.

EdwinGuzman (approved):

PR #232, opened 2024-06-03
Scc 4141/css standardization

charmingduchess:
## Ticket:

- JIRA ticket [SCC-4141](https://jira.nypl.org/browse/SCC-4141)

- removes redundant styling from modal props
- moves modal heading icons to be children of the text components

## How has this been tested?

Visual verification

-

### Checklist:


7emansell (commented):
Left margin on the check modal for canceling a request is off (same for the change password modal, but that might as well be since there's no stated design for that modal). Otherwise lgtm

src/components/MyAccount/CheckoutsTab/RenewButton.tsx
```diff
@@ -28,22 +28,22 @@ const RenewButton = ({
...
+        <Text>
           This item has been renewed. It is now due back on {checkout?.dueDate}.
         </Text>
       </Box>
     ),
     closeButtonLabel: "OK",
     headingText: (
       <Heading className={styles.modalHeading}>
-        <>
+        <Text>
```
EdwinGuzman, line 39: I don't think it's valid HTML to have a `p` element inside of an `h*` element. Can you move or add any necessary styles to the `Heading` parent?

EdwinGuzman (changes requested):
I can't test this right now because of Sierra Test issues, but the update generally looks okay. The only update that should be made is removing the `Text` component inside of the `Heading` component. It realistically won't cause any issues but it's invalid HTML.

If you need the specific text styles, try moving the styles to the `Heading` component and perhaps use the `size` prop for different font sizes.

charmingduchess:
@EdwinGuzman updated to remove the text components and consolidate the styles to one place in the css module.

src/components/MyAccount/ModalHeading.tsx
```diff
@@ -0,0 +1,21 @@
+import { Heading } from "@nypl/design-system-react-components"
+import styles from "../../../styles/components/MyAccount.module.scss"
+
+// this method is a workaround because the size of the heading can't be updated
+// from the css module. it's a method because the headingText prop of the modal
+// props object requires ONLY a Heading component and won't accept a custom
+// component to be passed in.
```
EdwinGuzman, line 7: This is interesting. For the first part, I think it's because the style props or using the `sx` or `__css` have higher priority than the classname rules. So it sounds like the use of CSS modules isn't completely supported.

The `headingText` prop should accept a string or HTML element. We do prefer a DS `Heading` component but you should be able to pass any other element such as `h2`, `h3`, etc. or even `<span>`s. But you do lose the default styling if it's a span. Can you share what you tried that didn't work?

styles/components/MyAccount.module.scss
```diff
@@ -42,25 +42,26 @@
...
-    letter-spacing: normal;
-  }
-  svg {
-    margin-top: var(--nypl-space-s);
-  }
+  padding-top: var(--nypl-space-s);
+  font-weight: var(--nypl-fontWeights-medium);
+  letter-spacing: normal;
+}
+.string-only {
```
EdwinGuzman, line 52: Nit pick: perhaps `string-only-heading` is a better defining name?

EdwinGuzman (approved):
This looks good and I have some questions and a nitpick but not blockers.

One thing that is different than other modals is the cancel request one. The skeleton loader and heading are not aligned but not sure if it needs to be.
 
Screenshot 2024-06-06 at 4 29 33 PM

src/components/MyAccount/ModalHeading.tsx
charmingduchess: replaced that method invocation with `<h5>`'s. thanks for the suggestion.

EdwinGuzman (approved):
I really liked the idea of the modal heading function you created and I _think_ I know why it wasn't working. It is most likely a logic check in the DS, but either way, this looks good.

PR #233, opened 2024-06-03
SCC-4032 - Extract getEmailParams from Feedback API route and add test

dgcohen:
## Ticket:

- JIRA ticket [SCC-4032](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4032)

Trying to clear this ticket that was added after the Feedback form functionality was added in a hurry for the 2AD launch. We currently don't have tests or mocks for our API route handlers, but I was able to abstract the param building functionality so we could at least test that.

We can revisit adding tests for API routes later, but that would be its own larger project.


charmingduchess (approved):

EdwinGuzman (approved):
Great

PR #234, opened 2024-06-04
SCC-4136 - Bib Details Pagination

dgcohen:
## Ticket:

- JIRA ticket [SCC-4136](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4136)

- Adds an api route /api/bib/[id]/items used for client side item fetching
- Renders the Pagination component on the Bib Page
- Adds a fetcher callback and state setting for loading, error and successfully fetched items (would like to move this to a custom hook eventually but will keep it here for now)
- Refactors bib utils to remove BibParams and rely solely on query params for fetching bibs, since BibParams weren't being used and there was excessive conversion back and forth.
- Adds a utility function to render the Item Table heading and moves getPaginationOffsetStrings from searchUtils to appUtils.

## How has this been tested?

- Tested manually and added unit tests.

### Checklist:


pages/bib/[id].tsx
```diff
@@ -49,8 +109,50 @@ export default function BibPage({
...
+              level="h3"
+              size="heading4"
+              mb={{ base: "s", md: "m" }}
+            >
+              Items in the library and off-site
+            </Heading>
+            {itemsLoading ? (
+              <SkeletonLoader showImage={false} />
+            ) : itemFetchError ? (
+              <Notification
```
Comment (EdwinGuzman) [inline pages/bib/[id].tsx:125]: Ah, use the `Banner` component for notifications that are local to the page. https://nypl.github.io/nypl-design-system/reservoir/v3/?path=/docs/components-feedback-banner--docs -- use the "negative" variant.

EdwinGuzman (changes requested):
Looks good and is functional, but I noticed two issues and did not review that code yet. 

- This seems to be true for almost any bib, the result display heading always seems to display `-19-0` https://research-catalog-git-scc-4136-bib-details-pagination-nypl.vercel.app/research/research-catalog/bib/b10833141:

Screenshot 2024-06-07 at 1 34 08 PM

- The `Pagination` component is not initializing correctly. On first load, the "Previous" button displays and the "1", which I expect to be selected, is not selected.

Screenshot 2024-06-07 at 1 34 11 PM

Once I click on a number or "Next", it works fine. This also works well when hitting the page directly with a requested page in the url.

dgcohen:
> Looks good and is functional, but I noticed two issues and did not review that code yet.
> 
> * This seems to be true for almost any bib, the result display heading always seems to display `-19-0` https://research-catalog-git-scc-4136-bib-details-pagination-nypl.vercel.app/research/research-catalog/bib/b10833141:
> 
> Screenshot 2024-06-07 at 1 34 08 PM
> * The `Pagination` component is not initializing correctly. On first load, the "Previous" button displays and the "1", which I expect to be selected, is not selected.
> 
> Screenshot 2024-06-07 at 1 34 11 PM
> Once I click on a number or "Next", it works fine. This also works well when hitting the page directly with a requested page in the url.

Looks like the two issues you pointed out were related and caused by setting the default value of itemPage to null in getServerSideProps. Couldn't pass undefined since it needed to be serializable and passing null doesn't use the default prop. Set a default of 1 in the json serialization and seems to be working now.

EdwinGuzman (commented):
Great! thanks for all the updates. I will do another closer look and debugging next week. Some more issues while testing with the New York as an example:
- there was a fetch error but the Banner did not come up. I was on page 37 and clicked on "Next" and this was the resulting API call endpoint `/api/bib/b10833141/items?b10833141?items_size=20&items_from=7400&item_page=371&id=b10833141&merge_checkin_card_items=true`
The string seems to be malformed and it says item_page 371! The endpoint returned a 500 error so the banner should come up:
Screenshot 2024-06-07 at 4 22 10 PM

- When you refresh the page, then the banner comes up and it looks good. I was going to say, "delete the pagination because it's not useful", but when I click on a page, the table loads correctly. I wonder if there's a better way to tell the user, "hey try again by refreshing the page or select a new page" (but selecting a new page may not be too intuitive and the "try a new query" is not too helpful if the user doesn't know how to do so).

Screenshot 2024-06-07 at 4 24 07 PM

- The data loads correctly but the display is still off sometimes. The first time I tried it, I kept getting "1-20 of 816 items" when not on page 1. And then I got other values such as 
Screenshot 2024-06-07 at 4 26 02 PM

I think I know what's happening. If you go to a page directly such as https://research-catalog-git-scc-4136-bib-details-pagination-nypl.vercel.app/research/research-catalog/bib/b10833141?item_page=7, the text is `121-140 of 816 items"`. This seems right, but when you click on a new page, the value does not update. It might be related to the shallow router `push` call or the state values not updating.

src/utils/bibUtils.ts
charmingduchess, line 1: 👍🏻

charmingduchess (changes requested):
Looks good!
A couple bugs I'm seeing running locally:
- After the bib page loads, hitting Next reloads the page with the same first page of results. 
- when I input a page number into the url, the numbers expand like this and I don't see the page I'm on in black. I think this can be replicated by copy pasting any url with a `item_page` provided
Screenshot 2024-06-10 at 3 25 24 PM
I'm also not seeing the header with the "x-x results of ..." But it looks like you removed that intentionally.

pages/api/bib/[id]/items.ts
```diff
@@ -0,0 +1,33 @@
...
+      res.status(500).json({
+        error: "Error fetching Bib items for this query",
+      })
+    } else {
+      res.status(200).json({
+        items,
+      })
+    }
+  }
+  if (req.method === "POST") {
```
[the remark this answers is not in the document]
Comment (dgcohen) [reply pages/api/bib/[id]/items.ts:26]: no reason, just moved that statement to an else block

dgcohen:
> Looks good! A couple bugs I'm seeing running locally:
> 
> * After the bib page loads, hitting Next reloads the page with the same first page of results.
> * when I input a page number into the url, the numbers expand like this and I don't see the page I'm on in black. I think this can be replicated by copy pasting any url with a `item_page` provided
>   Screenshot 2024-06-10 at 3 25 24 PM
>   I'm also not seeing the header with the "x-x results of ..." But it looks like you removed that intentionally.

This was happening because I wasn't converting the page number in the query param to a number in getServerSideProps. Fixed.

pages/api/bib/[id]/items.ts
```diff
@@ -0,0 +1,32 @@
...
+      res.status(500).json({
+        error: "Error fetching Bib items for this query",
+      })
+    } else {
+      res.status(200).json({
+        items,
+      })
+    }
+  } else {
+    res.status(500).json({
```
Comment (EdwinGuzman) [inline pages/api/bib/[id]/items.ts:26]: This seems more like a 400 since it's user error.

pages/api/bib/[id]/items.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+async function handler(req: NextApiRequest, res: NextApiResponse) {
+  const id = req.query.id as string
+  const { discoveryBibResult, status } = await fetchBib(id, req.query)
+  const items = discoveryBibResult?.items
+
+  if (req.method === "GET") {
+    if (status !== 200 || !items?.length) {
+      res.status(500).json({
```
Comment (EdwinGuzman) [inline pages/api/bib/[id]/items.ts:17]: Perhaps 400? 500 is more for an internal server error but this is an API fetch error.

pages/bib/[id].tsx
```diff
@@ -49,8 +111,50 @@ export default function BibPage({
...
+                <>
+                  <Heading
+                    data-testid="item-table-displaying-text"
+                    level="h4"
+                    size="heading5"
+                    mb={{ base: "s", md: "m" }}
+                  >
+                    {buildItemTableDisplayingString(
+                      itemPage,
+                      bib.numPhysicalItems
```
Comment (EdwinGuzman) [inline pages/bib/[id].tsx:142]: The text `Displaying 1-20 of 816 items` never updates on the client side. So if you keep pressing "next" in the pagination, the "1-20" never updates. If you go to a page with `item_page=X` directly, it'll work. 

This is because it's reading `itemPage` from `getServerSideProps` which is ideal for getting the data from the server on first load. But on clientside interaction, the value gets updated appropriately from `query.item_page` (from the `useRouter()` hook). So I think you need to find a way to get `itemPage` on first load only and then update the value based on user interaction the Pagination from `handlePageChange` function.

EdwinGuzman (changes requested):
Looks good but just one request for updating the display text!

pages/bib/[id].tsx
Comment (dgcohen) [reply pages/bib/[id].tsx]: Good catch. This was happening because the page number wasn't being updated in state. Just added a useState that fixed the issue. Could also access the param via the router query but it would require another parseInt.

pages/api/bib/[id]/items.ts
Comment (dgcohen) [reply pages/api/bib/[id]/items.ts]: Good call, fixed.

EdwinGuzman (approved):

charmingduchess (approved):
One tiny note! Looks great.

src/server/api/bib.ts
```diff
@@ -30,12 +30,18 @@ export async function fetchBib(
...
   const [bibResponse, annotatedMarcResponse] = await Promise.allSettled([
     await client.get(
-      `${DISCOVERY_API_SEARCH_ROUTE}/${getBibQuery(id, bibParams)}`
+      `${DISCOVERY_API_SEARCH_ROUTE}/${standardizedId}${getBibQueryString({
+        ...bibQuery,
+        id: standardizedId,
+      })}`
     ),
     // Don't fetch annotated-marc for partner records:
     isNyplBibID(id) &&
```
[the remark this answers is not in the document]
dgcohen: Good catch! fixed

PR #241, opened 2024-06-13
SCC-4140 - Bib Details View All

dgcohen:
## Ticket:

- JIRA ticket [SCC-4140](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4140)

- Adds View All items functionality to the bib page
- Introduces the fetchItems function, which fetches the bib's items in a loop with a configurable batch size separate from pagination and returns a failed status if any of the requests fail.
- Updates the items api route handler to invoke fetchItems instead of fetchBib
- Refactors getBibQueryString to ensure consistent results with no repeated or unnecessary query params
- Introduces the /all page file with its own getServerSideProps, so all items are client-side fetched when the page is accessed with /all subroute.
- Adds new loading state copy (for View All only) and control of focus state during and after item loading.
- Renames itemMessage to numItemsMessage and SearchResultsItem to DiscoveryItemResult.

## How has this been tested?

- Unit tests added/updated

- New focus control should be reviewed by Clare for accessibility

### Checklist:

- [x] Clare has reviewed focus state changes for accessibility.

[SCC-4140]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4140?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/server/api/tests/items.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      "b17418167",
+      {
+        items_from: 0,
+      },
+      false
+    )) as ItemsResponse
+    expect(itemsResponse.items.length).toEqual(0)
+    expect(itemsResponse.status).toEqual(400)
+  })
+  it("should fetch items in batches when viewAllItems is true", async () => {
```
EdwinGuzman, line 98: This doesn't really seem like it's testing that the fetch was called multiple times for the batch fetch. Can you try adding an assertion for expect fetch `toBeCalled` x times?

src/types/itemTypes.ts
```diff
@@ -38,6 +41,11 @@ export interface ItemMetadata {
   bibId?: string
 }
 
+export interface ItemsResponse {
+  items: DiscoveryItemResult[]
+  status: 200 | 400
```
EdwinGuzman, line 46: I'm a bit wary of this because maybe it should align with HTTP codes an allow _any_ and all status codes?

pages/bib/[id]/index.tsx
```diff
@@ -165,26 +224,96 @@ export default function BibPage({
...
+                    >
+                      <ProgressIndicator
+                        id="bib-all-items-loading"
+                        labelText={bib.itemsViewAllLoadingMessage}
+                        size="small"
+                        indicatorType="circular"
+                        mr="xs"
+                        isIndeterminate
+                      />
+                      <Label
```
Comment (EdwinGuzman) [inline pages/bib/[id]/index.tsx:271]: This component needs an `id` prop.

pages/bib/[id]/index.tsx
```diff
@@ -76,27 +89,62 @@ export default function BibPage({
     isArchiveCollection: bib.isArchiveCollection,
   })
 
-  const refreshItemTable = async (newQuery: BibQueryParams) => {
+  // Load all items via client-side fetch if page is first loaded with viewAllItems prop passed in
+  // Namely, when the page is accessed with the /all route
+  useEffect(() => {
+    if (viewAllItems) void refreshItemTable(query, true)
```
Comment (EdwinGuzman) [inline pages/bib/[id]/index.tsx:95]: What's the purpose of `void` here?

pages/bib/[id]/index.tsx
```diff
@@ -76,27 +89,62 @@ export default function BibPage({
...
+    // Disable eslint exhaustive-deps rule because we only want this to run once on page load
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
Comment (EdwinGuzman) [inline pages/bib/[id]/index.tsx:98]: Why can't you do this on the server side again? If I land on `/all`, why can't all items be fetched on the server and returned as props to the client? This would be similar to how `fetchBib` is called on `/all` but would it work to call `fetchBib` and then `fetchItems` with the all flag set to true? Is this partially because of the loading time before the page loads?

EdwinGuzman (commented):
I need to break this PR review into multiple passes so this is my first pass with some questions and comments.

src/server/api/items.ts
```diff
@@ -0,0 +1,70 @@
...
+    // Return a failed response if the first fetch fails
+    return {
+      items: [],
+      status: 400,
+    }
+  }
+
+  // If View All is enabled, fetch the items in large batches
+  for (
+    let batchNum = 1;
```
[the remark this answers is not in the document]
dgcohen: Yeah exactly, this allows us to use a different batch number to cut down on the total number of requests.

pages/bib/[id]/index.tsx
```diff
@@ -165,26 +224,96 @@ export default function BibPage({
...
+                      <Label
+                        htmlFor="bib-all-items-loading"
+                        ref={viewAllLoadingTextRef}
+                        fontSize={{
+                          base: "mobile.body.body1",
+                          md: "desktop.body.body1",
+                        }}
+                        fontWeight="medium"
+                        mb={0}
+                        // Label component does not expect tabIndex prop, so we are ignoring the typescript error that pops up.
```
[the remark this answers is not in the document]
Comment (dgcohen) [reply pages/bib/[id]/index.tsx:287]: I made it a label since the ProgressIndicator label is not rendered when the size is "small". Otherwise I could probably override the styling, maybe we can fix this on the DS label as a follow up.

src/models/Bib.ts
```diff
@@ -66,6 +67,24 @@ export default class Bib {
...
+  }
+
+  get numItemsMessage() {
+    return `${this.numItems} ${this.resourceType}${
+      this.numItems !== 1 ? "s" : ""
+    }`
+  }
+
+  get itemsViewAllLoadingMessage() {
+    return `Loading all ${this.numPhysicalItems} items...this may take a few moments.`
```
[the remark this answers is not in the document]
dgcohen: that makes sense, will double check with Apoorva since that's the copy I got from her.

src/types/itemTypes.ts
dgcohen: I think we would need an external package to get all available status codes without specifying themselves. I'm not totally sure it's necessary though since we manually specify what specific status code gets sent back in fetchItems.

pages/bib/[id]/index.tsx
Comment (dgcohen) [reply pages/bib/[id]/index.tsx:95]: That operator prevents an unhandled Promise warning.

pages/bib/[id]/index.tsx
Comment (dgcohen) [reply pages/bib/[id]/index.tsx:98]: Doing this for the reason you mentioned, to avoid an extremely long page load time on bibs with a lot items. This will be necessary when JS is disabled, but since almost all users will have it enabled I don't think it makes sense to prevent the rest of the page from rendering.

src/models/Bib.ts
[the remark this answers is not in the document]
dgcohen: Just changed the copy to "Loading all 2000 items. This may take a few moments..." on Apoorva's suggestion.

src/server/api/tests/items.test.ts
dgcohen: good call, done.

pages/bib/[id]/index.tsx
```diff
@@ -165,26 +224,96 @@ export default function BibPage({
...
               )}
-              <Pagination
-                id="bib-items-pagination"
-                initialPage={itemTablePage}
-                currentPage={itemTablePage}
-                pageCount={Math.ceil(bib.numPhysicalItems / ITEM_BATCH_SIZE)}
-                onPageChange={handlePageChange}
-                my="xl"
-              />
+              <Box display="flex" my="xl" justifyContent="space-between">
```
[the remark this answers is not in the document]
Comment (dgcohen) [reply pages/bib/[id]/index.tsx]: Just extracted ItemTableControls to its own component, if it looks good to you and @charmingduchess I will add tests, if not I can undo.

pages/api/bib/[id]/items.ts
```diff
@@ -9,8 +10,9 @@ import { fetchBib } from "../../../../src/server/api/bib"
  */
 async function handler(req: NextApiRequest, res: NextApiResponse) {
   const id = req.query.id as string
-  const { discoveryBibResult, status } = await fetchBib(id, req.query)
-  const items = discoveryBibResult?.items
+  const viewAllItems = req.query?.view_all_items === "true"
+
+  const { status, items } = await fetchItems(id, req.query, viewAllItems)
```
Comment (EdwinGuzman) [inline pages/api/bib/[id]/items.ts:15]: Nitpick: It feels a bit unnecessary to get the `id` from `req.query.id` and pass it to `fetchItems` as an argument when the entire `req.query` object is passed as the second argument.

Comment (EdwinGuzman) [reply pages/api/bib/[id]/items.ts:15]: Same for `view_all_items` being checked. Would it make sense to pass the entire object and let `fetchItems` process the object? that function could be its own small function in `/src/server/api/items` that lives alongside `fetchItems`.

src/server/api/items.ts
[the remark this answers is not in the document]
EdwinGuzman: Yes, but this already fetches the first 20 and then throws it away and fetches the first 150. Minor issue.

pages/bib/[id]/index.tsx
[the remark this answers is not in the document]
Comment (EdwinGuzman) [reply pages/bib/[id]/index.tsx]: How does it look like when it's regular size? I would prefer to update this on the DS side.

pages/bib/[id]/index.tsx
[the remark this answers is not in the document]
Comment (EdwinGuzman) [reply pages/bib/[id]/index.tsx]: Yes, looks good.

src/types/itemTypes.ts
EdwinGuzman: Then we should have a global list of status codes used in the app for consistency.

pages/bib/[id]/index.tsx
Comment (EdwinGuzman) [reply pages/bib/[id]/index.tsx:93]: Yes, but we should think of features as "assume users don't have JS (or ship as little JS as possible to the client) and what features can we progressively add". It's a shift from thinking of it as "users won't disable JS", which is generally true, and also aligns with "server-first" approaches.

src/components/ItemTable/ItemTableControls.tsx
```diff
@@ -0,0 +1,117 @@
...
+          >
+            <ProgressIndicator
+              id="bib-all-items-loading"
+              labelText={bib.itemsViewAllLoadingMessage}
+              size="small"
+              indicatorType="circular"
+              mr="xs"
+              isIndeterminate
+            />
+            <Label
```
EdwinGuzman, line 67: Create a ticket for this https://newyorkpubliclibrary.atlassian.net/browse/DSD-1798

But now I'm rethinking this and wondering if perhaps it's okay to leave the `ProgressIndicator` as is and replace `Label` with a span and remove its `htmlFor` attribute. Internally, the `ProgressIndicator` will add the labelText as an aria-label so this visible `Label` is redundant for screen readers.

EdwinGuzman (commented):
Minor comments. This looks good to me but @charmingduchess can you do a better pass in the utility functions for the query builder?

pages/bib/[id]/index.tsx
```diff
@@ -55,17 +57,23 @@ export default function BibPage({
...
 }: BibPropsType) {
-  const { pathname, push, query } = useRouter()
+  const { push, query } = useRouter()
   const metadataTitle = `Item Details | ${SITE_NAME}`
   const bib = new Bib(discoveryBibResult)
   const displayLegacyCatalogLink = isNyplBibID(bib.id)
 
   const [itemsLoading, setItemsLoading] = useState(false)
   const [itemFetchError, setItemFetchError] = useState(bib.showItemTableError)
+  const [viewAllEnabled, setViewAllEnabled] = useState(viewAllItems)
```
Comment (charmingduchess) [inline pages/bib/[id]/index.tsx:69]: `viewAllEnabled` sounds like a feature flag to me. Can you change this variable name?

pages/bib/[id]/index.tsx
Comment (charmingduchess) [inline pages/bib/[id]/index.tsx:1]: I'd be interested in seeing how this would look if `viewAllEnabled` was based on the presence of `/all` in the url. I like the pattern of the url determining the state. The logic isn't terribly complicated here, but I think it adds clarity to see that there is one flow of state, from the url down to the component. I don't know that it's necessarily the best way to do things, but that's how we handled the tabs in MyAccount, and it's also how the filters work. 
That said, I'm not seeing where the `viewAllItems` prop is being passed in though, so maybe there something happening somewhere that I'm missing.

Comment (dgcohen) [reply pages/bib/[id]/index.tsx:1]: Thanks, I definitely agree with what you're saying but with Next version 12, the only way to target /all specifically without a catch-all route is to use a unique file-based page component (pages/bib/[id]/all.tsx). Notice how that file simply exports the page component from the index file but passes the viewAllEnabled in its own getServerSideProps function.

This might be possible to do without a separage page file in Next 13 but I'm not entirely sure of the implantation details yet.

pages/api/bib/[id]/items.ts
Comment (dgcohen) [reply pages/api/bib/[id]/items.ts]: Very good suggestions. Just updated fetchItems to just take a bibQuery param with the option to pass in a batch size query param for testing.

src/types/itemTypes.ts
dgcohen: Confirmed over slack that this can be done as a follow up so as not to further increase the scope of this already large PR. Here is the ticket for that: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4172

pages/bib/[id]/index.tsx
Comment (dgcohen) [reply pages/bib/[id]/index.tsx:93]: Chatted about this with @EdwinGuzman over slack, we will have a conversation with Vera when she is back on how want to approach JS-disabled detection on the server side.

pages/bib/[id]/index.tsx
```diff
@@ -12,20 +12,24 @@ import {
...
   SITE_NAME,
   BASE_URL,
+  FOCUS_TIMEOUT,
 } from "../../../src/config/constants"
+import { appConfig } from "../../../src/config/config"
 import { fetchBib } from "../../../src/server/api/bib"
 import {
-  buildItemTableDisplayingString,
   getBibQueryString,
+  buildItemTableDisplayingString,
```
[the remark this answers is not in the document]
Comment (dgcohen) [reply pages/bib/[id]/index.tsx:23]: This change seems easy but causes a circular dependency that breaks all the tests. Going to leave it as is but opening a ticket to address the circular dependency issue:

https://newyorkpubliclibrary.atlassian.net/browse/SCC-4173

EdwinGuzman (commented):
The latest round of updates looks good. I don't see the banner with the "enable js" message when I turn off javascript in my browser. I don't think that's working and if it's not, then it's okay to remove that code and implement it later.

EdwinGuzman (approved):
Great, thanks for removing the no-JS banner for now.

charmingduchess (approved):
Working nicely! Was there ever discussion of having the `view fewer` button above the items table?

dgcohen:
> button

Haven't brought that up but I like the idea, we should propose that to Apoorva.

PR #246, opened 2024-06-20
Scc 4162/vqa account settings

charmingduchess:
## Ticket:

- JIRA ticket [SCC-4162](https://newyorkpubliclibrary.atlassian.net/jira/software/c/projects/SCC/boards/126?assignee=61e836e95fcc37006878c15d&selectedIssue=SCC-4162)

-    adds dashes to the phone number display

-    move update pin/password button

-    align pin password modal body with heading

-    Ensure the edit/save/cancel buttons move to the bottom of the tab on mobile view

-

### Checklist:

### DIFF
--- a/styles/components/MyAccount.module.scss
+++ b/styles/components/MyAccount.module.scss
@@ -35,13 +35,10 @@
     display: flex;
     flex-direction: column;
     @media screen and (min-width: 600px) {
-      flex-direction: row;
-    }
-    @media screen and (max-width: 800px) and (min-width: 601px) {
       flex-direction: column-reverse;
     }
-    @media (max-width: 600px) {
-      flex-direction: column;
+    @media screen and (min-width: 800px) {
+      flex-direction: row;
     }
   }
 }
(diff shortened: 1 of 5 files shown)


styles/components/MyAccount.module.scss
```diff
@@ -31,11 +31,14 @@
     margin-top: var(--nypl-space-l);
     display: flex;
     flex-direction: row;
-    @media (max-width: 600) {
+    @media screen and (max-width: 600px) {
```
EdwinGuzman, line 34: Can you switch these over to use `min-width` instead? That'll follow the "mobile-first" approach. So in this case it'll be `flex-direction: column` be default and once the `min-width: 600px` value kicks in, it'll switch to `flex-direction: row`.

You can also use `var(--nypl-breakpoint-medium)` as the CSS var.

EdwinGuzman (commented):
Looks good just one minor change but not a blocker.

Can you merge `main` in first before I approve? I want to see if this PR is affected by the updated styles in `main`.

EdwinGuzman (approved):
There's a minor responsive issue with the buttongroup at exactly 600px where it doesn't go into full width but I think that's a DS problem.

charmingduchess:
> There's a minor responsive issue with the buttongroup at exactly 600px where it doesn't go into full width but I think that's a DS problem.

I had one of the min width set to 601 pixels. removed that and also rearranged the order of the styling to simplify.

PR #249, opened 2024-06-24
Scc 4146/account redirect

charmingduchess:
## Ticket:

- JIRA ticket [SCC-4146](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4146)

- Adds redirect loop handling to the account page. In the event that a user lands on my account and is not authenticated, we redirect to the CAS login page. if the user is redirected back to RC but is still not authenticated, we will redirect them back to log in up to 3 times. After that, display an error page indicating that a login server error has occurred.

## How has this been tested?

I've tried it locally, by removing the `!isAuthenticated` check [here](https://github.com/NYPL/research-catalog/blob/7b449a81490ca1ec4db14e109ed014314701921c/pages/account/%5B%5B...index%5D%5D.tsx#L88). when I do this, the redirect url intermittently is a backend next url:
`http://local.nypl.org:8080/research/research-catalog/_next/data/development/account.json` and the raw redirect object from `getServerSideProps`. It also sets an additional nyplAccounRedirects cookie from the same `_next/data` path. Not sure if this is due to the contrived nature of the test or what!

UPDATE: setting the domain and path explicitly in the setHeaders invocation fixed the issue of the double cookies, but not of the backend url ending up in the browser url bar.

In terms of jest testing, I have the following test cases passing, open to adding more considering how difficult this is to test locally.
- "redirects if doRedirect is true"
-  "does not redirect if doRedirect method returns false"
-  "does not redirect if patron is authenticated"
- "updates the nyplAccountRedirectsCookie upon redirecting"

-

### Checklist:

[SCC-4146]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


charmingduchess:
@EdwinGuzman I added the path and domain, which seems to eliminate the extra cookie coming from the server. The expired cookies from yesterday (under a different path) were still around, but I'm not sure that's a big issue. I will investigate further! But I'm feeling good about where this is now.

EdwinGuzman:
@charmingduchess great, can you set it ready for review and add Diego and perhaps Paul?

EdwinGuzman (approved):

nonword (approved):
The cookie-setting/checking stuff looks right. Not sure about the intermittent next redirect issue though..

dgcohen (approved):

PR #250, opened 2024-06-26
Scc 4168/chris qa1

charmingduchess:
## Ticket:

- JIRA ticket [SCC-4168](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4168)

- sort checkouts and holds into chronological order
- append authors to titles
- updates hold and checkout fixtures, as well as refactoring some tests to be more flexible to differently ordered fixtures
- removes a bunch of unused imports and vars (sorry for the noise)

-

### Checklist:


src/components/MyAccount/CheckoutsTab/CheckoutsTab.test.tsx
```diff
@@ -33,11 +33,14 @@ describe("CheckoutsTab", () => {
     const component = render(
       <CheckoutsTab patron={processedPatron} checkouts={processedCheckouts} />
     )
+    const renewableCheckout = processedCheckouts[0]
+    const row = component.getByText(renewableCheckout.title).closest("tr")
```
EdwinGuzman, line 37: I just noticed this does `component.getBy...` where `component` is coming from the render function. Usually, it's  `screen.getBy...` and screen is imported from the package. Did the latter not work and you had to use `component` from the render function? They're essentially the same so more out of curiosity.

src/models/MyAccount.ts
```diff
@@ -154,12 +154,25 @@ export default class MyAccount {
     return bibs.reduce((bibDataMap: BibDataMapType, bibFields) => {
       const { isResearch, isNyplOwned } =
         this.getResearchAndOwnership(bibFields)
-      const title = bibFields.title
+      const title = `${bibFields.title}${
+        bibFields.author && isNyplOwned ? ` / ${bibFields.author}` : ""
```
EdwinGuzman, line 158: I'm guessing that the date + "- author" is a data problem?

Screenshot 2024-06-26 at 2 48 45 PM

EdwinGuzman (approved):
Just one question

src/components/MyAccount/CheckoutsTab/CheckoutsTab.test.tsx
charmingduchess: No this was inadvertent. I'll update.

charmingduchess: Ok, actually... Component is returned by render which comes from our wrapped methods. I'm not exactly sure why that's the pattern here, but it's working, and I'm not so inclined to fix it. Unless you think its important to have it standardized.

src/models/MyAccount.ts
charmingduchess: Yes. I think some author fields have the years the author is alive for, and also the word author apparently.

src/components/MyAccount/CheckoutsTab/CheckoutsTab.test.tsx
EdwinGuzman: ah, right. it's ok to keep

PR #251, opened 2024-06-26
SCC-4139 - Bib Details Item Filters

dgcohen:
## Ticket:

- JIRA ticket [SCC-4139](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4139)

- Renders the FiltersContainer component and moves the control of the router to the parent component (BibPage index).
- Slightly tweaks the styling and components used in FiltersContainer, but mind that these changes are temporary and will be removed once the UI is updated as part of this ticket: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4176
- Uses numItemsMatched instead of numPhysicalItems to render item total when filters are applied and adjusts display string generator utils.
- Updates logic in bib index page to always update the full bib instead of just the ItemTable (necessary now that we need access to numItemsMatched).
- Renames "aggs' to "aggregation" throughout the app, personal preference since it makes for easier searching.

## How has this been tested?

- Updated or added unit tests where relevant
- Removed some integration tests for FiltersContainer since most of the internal components will soon be replaced.
- Note that I decided not to add integration tests related to filtering for BibPage yet since the UI for the controls will be changing soon.

- Similarly to the integration tests, we should hold off on VQA and accessibility testing til the filter control components are updated.

### Checklist:

- [X] I updated the CHANGELOG with the appropriate information and JIRA ticket number (if applicable).
- [X] All new and existing tests passed.

[SCC-4139]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/ItemFilters/FiltersContainer.tsx
```diff
@@ -1,73 +1,103 @@
...
+    setWhichFilterIsOpen("")
+  }
+
+  const clearAllFilters = () => {
+    handleFiltersChange({})
     setWhichFilterIsOpen("")
-    router.push("/search/advanced" + url)
+  }
+
+  const handleRemoveFilterClick = ({ id }: TagSetFilterDataProps) => {
```
EdwinGuzman, line 84: I mentioned this before but clicking on a tag to remove the filter does not seem to work.

src/components/ItemFilters/FiltersContainer.tsx
```diff
@@ -78,62 +108,83 @@ const ItemFilterContainer = ({ itemAggs }: ItemFilterContainerProps) => {
...
+                appliedFilters={appliedFilters}
+                submitFilters={submitFilters}
+              />
+            ))}
+          </Box>
+        </Box>
+        <Box className={filterGroupClassName} minWidth={440}>
+          <Label
+            id="year-filter-label"
+            htmlFor="year-filter"
```
EdwinGuzman, line 145: Not a fan of this but I know why you're doing this. Not your fault but something I need to chat with Marty about regarding showing labels in the `SearchBar` component.

src/components/ItemFilters/FiltersContainer.tsx
```diff
@@ -78,62 +108,83 @@ const ItemFilterContainer = ({ itemAggs }: ItemFilterContainerProps) => {
...
-          </CardContent>
-        </Card>
+          </Label>
+          <SearchBar
+            id="year-filter"
+            labelText="Apply"
+            textInputProps={{
+              placeholder: "YYYY",
+              isClearable: true,
+              labelText: "Item Search",
```
EdwinGuzman, line 157: This should be "Search by year"

src/utils/itemFilterUtils.ts
```diff
@@ -15,71 +19,92 @@ export const combineRecapLocations = (locations: string[]) => {
...
+      )
+      const valueLabel =
+        field === "year"
+          ? filterValue
+          : fieldAggregations?.labelForValue(filterValue)
+      if (valueLabel) {
+        filters.push({
+          label: `${capitalize(field)} > ${valueLabel}`,
+          id: kebabCase(`${field}-${valueLabel}`),
+          iconName: "close",
```
EdwinGuzman, line 89: From the docs, you don't need this property and the console also warns you, so make sure to check the messages:
Screenshot 2024-07-10 at 6 31 16 PM

EdwinGuzman (changes requested):
This is a big PR so this is the first found of review. I pointed out some changes but the biggest that I haven't looked into is that clicking on a tagset is not working. The "clear filters" button is working though.

On this page https://research-catalog-git-scc-4139-bib-details-item-filters-nypl.vercel.app/research/research-catalog/bib/b10833141?item_location=loc%3Amal82&item_format=Text&item_status=status%3Ana
with those three filters selected, I got an error from the API endpoint:

https://research-catalog-git-scc-4139-bib-details-item-filters-nypl.vercel.app/research/research-catalog/api/bib/b10833141/items?items_size=20&items_from=0&item_page=1&item_location=loc:mal82&item_format=Text&item_status=status:na&merge_checkin_card_items=true

worth looking into the issue.

pages/api/bib/[id]/items.ts
```diff
@@ -18,6 +18,7 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
       })
     } else {
       res.status(200).json({
+        discoveryBibResult,
```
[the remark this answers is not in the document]
Comment (dgcohen) [reply pages/api/bib/[id]/items.ts:21]: That makes sense. Would it be preferable to pass showItemTable instead of the whole bib? Or would it be better to remove the items endpoint entirely and move the looped items fetch to the bib endpoint and refresh the whole bib?

The first option would be easy to implement on top of what I have, but the second would reduce the amount of code we'll have to maintain. What do you and @charmingduchess think?

src/components/ItemFilters/FiltersContainer.tsx
dgcohen: I see what happened here. I tried to make the ids uniqe and kebab-cased but forgot that they needed to be the value to allow for proper clearing of the filter. Fixed.

src/utils/utilsTests/itemFilterUtils.test.tsx
```diff
@@ -71,53 +73,75 @@ describe("Item Filter Utils", () => {
...
     it("with all filters", () => {
-      expect(buildAppliedFiltersString(query, aggs)).toBe(
-        "Filtered by location: 'Offsite', format: 'Text', status: 'Available'"
-      )
-    })
-    it("all filters filters", () => {
-      expect(buildAppliedFiltersString(query, aggs)).toBe(
-        "Filtered by location: 'Offsite', format: 'Text', status: 'Available'"
-      )
+      expect(buildAppliedFiltersTagSetData(query, aggregations)).toStrictEqual([
```
charmingduchess, line 116: This and the next test need to be updated with ids that match the labels

charmingduchess (commented):
Seems good overall. I'm not seeing the tagset displayed anywhere on the bib page. I also see this button styled in a strange way:
Screenshot 2024-07-11 at 3 11 18 PM

I'm definitely in favor of removing the items route and using the existing bib route. Is that possible to implement without reloading the entire bib page while the items load?

src/components/ItemFilters/FiltersContainer.tsx
```diff
@@ -1,73 +1,103 @@
...
   )
+
   const ref = useRef<HTMLDivElement>(null)
+
   useCloseDropDown(() => setWhichFilterIsOpen(""), ref)
 
   const [whichFilterIsOpen, setWhichFilterIsOpen] = useState("")
 
-  const itemsMatched = buildItemsMatchedStringString(router.query)
+  const filtersAreApplied = areFiltersApplied(appliedFilters)
```
EdwinGuzman, line 64: Why not pass this as a prop instead of having to recheck the value here?

src/components/ItemTable/ItemTableControls.tsx
```diff
@@ -58,7 +60,7 @@ const ItemTableControls = ({
           >
             <ProgressIndicator
               id="bib-all-items-loading"
-              labelText={bib.itemsViewAllLoadingMessage}
+              labelText={bib.getItemsViewAllLoadingMessage(filtersAreApplied)}
```
EdwinGuzman, line 63: Make this a variable so it's only called once.

pages/bib/[id]/index.tsx
```diff
@@ -122,37 +131,58 @@ export default function BibPage({
...
+      const response = await fetch(
+        `${BASE_URL}/api/bib/${bib.id}${bibQueryString}`,
+        {
+          method: "get",
+          signal,
+        }
+      )
+      if (response?.ok) {
+        const { discoveryBibResult } = await response.json()
+        setBib(new Bib(discoveryBibResult))
```
Comment (EdwinGuzman) [inline pages/bib/[id]/index.tsx:150]: Is it not possible to do something like `discoveryBibResult.updateItems(...)` instead of having to create a new bib model and update the state?

You'd have to do it with other properties but it seems better, if possible, in order to reduce one more state variable.

Comment (dgcohen) [reply pages/bib/[id]/index.tsx:150]: Yeah that's possible, the only difference is we'd need to keep track of the variables that need to be refreshed and pass them in explicitly. I'm cool with that approach but pointing it out.

Comment (EdwinGuzman) [reply pages/bib/[id]/index.tsx:150]: That makes sense to me since you're being upfront about what needs to be updated in the model. It's still the same model with the same base information so that's why it doesn't make sense to me to reecreate the model every time. Any thoughts @charmingduchess ?

Comment (EdwinGuzman) [reply pages/bib/[id]/index.tsx:150]: The overhead isn't big, there's really no big risk in re-creating the model, but this is from a conceptual approach.

src/utils/utilsTests/itemFilterUtils.test.tsx
dgcohen: That's what i thought too when getting ready to open the PR but after I made the change it broke the item clearing functionality.

The way the TagSet component currently works, clicking on a tag passes the id as an attribute, so that needs to match the filter value (instead of the label) so we know which filter needs to be removed in removeValueFromFilters.

EdwinGuzman (approved):
Approving so far as is with the caveat that:

- eventually filtering should work as a standard HTML form for the no-js/serverside form submission,
- there should be a ticket for updating the state so bib is not re-created on every fetch call,
- functions will most likely update when `MultiSelect` and `FilterBarInline` are introduced,
- waiting on approval from Vera.

pages/bib/[id]/index.tsx
Comment (dgcohen) [reply pages/bib/[id]/index.tsx:150]: Discussed this offline with Edwin and decided to create a follow-up ticket to optimize state in the component. I added details there:

https://newyorkpubliclibrary.atlassian.net/browse/SCC-4193

pages/bib/[id]/index.tsx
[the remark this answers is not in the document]
Comment (dgcohen) [reply pages/bib/[id]/index.tsx:1]: Good catch! Added some logic prevent the item table from rendering and returning "No items found" in the display string when no items are found for a valid query.

Checking in with Apoorva about the No results wording.

pages/bib/[id]/index.tsx
```diff
@@ -122,37 +131,58 @@ export default function BibPage({
...
-    controllerRef.current = new AbortController()
-    const signal = controllerRef.current.signal
-
-    const response = await fetch(
-      `${BASE_URL}/api/bib/${bib.id}/items${bibQueryString}`,
-      {
-        method: "get",
-        signal,
+        setItemsLoading(false)
+        itemTableScrollRef.current?.scrollIntoView({
```
Comment (charmingduchess) [inline pages/bib/[id]/index.tsx:153]: is this scrolling behavior a current requirement or is it leftover from DFE? I believe we had the scrolling from when we were batch loading all the items from view all items.I think the experience is its potentially an accessibility issue and jarring to have the screen move and obscure the applied filters

charmingduchess (commented):
I just have one comment about the UX. Otherwise looks good.

pages/bib/[id]/index.tsx
Comment (dgcohen) [reply pages/bib/[id]/index.tsx:153]: Good call out about obscuring the applied filters, but this was an accessibility suggested by Edwin and confirmed by Clare.

Comment (charmingduchess) [reply pages/bib/[id]/index.tsx:153]: what is the issue that the scroll into view solves?

charmingduchess (approved):

PR #256, opened 2024-07-08
SCC-4154 timed logout modal

charmingduchess:
## Ticket:

- JIRA ticket [SCC-4154](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4154)

- Adds a timeout dialog warning a user they may be automatically logged out. 

A couple notes:

- I currently have the timeout warning showing up with 5 minutes left just for ease of testing, but this will eventually be a 2 minute warning. 
- I was also considering moving to the account page all the logic dealing with figuring out how much time is left, and by extension the rendering of the modal. However, I decided to keep that logic in the modal component to avoid coupling these two components, with an eye towards the rest of the website having a similar logout timeout. 

## How has this been tested?

I'm really struggling with how to test this. Not really sure how to have a minute long timeout within a test. To that end, I was considering moving the logic to calculate the time left and whether to render the modal outside of the modal component. 
I have not yet written any tests that pass. But running this locally, I am able to stay logged in, see the token get reset. The log me out flow is also working as far as I can see. 

- I imagine this will need some focus handling but I'm not sure what that should be.

### Checklist:

[SCC-4154]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/MyAccount/TimedLogoutModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+  timeoutWindow = 5,
+}: {
+  expirationTime: string
+  stayLoggedIn: () => void
+  timeoutWindow?: number
+}) => {
+  const [open, setOpen] = useState(false)
+  const router = useRouter()
+  const redirectUri = useLogoutRedirect()
+  const logOutAndRedirect = () => {
```
EdwinGuzman, line 33: Ah, I thought this function lived outside of the file. So my previous suggestion of mocking this function doesn't apply.

EdwinGuzman (approved):

src/utils/cookieUtils.ts
```diff
@@ -0,0 +1,43 @@
+const includeDomain = process.env.NODE_ENV !== "test"
+
+export const getTimeLeft = (expTime) => {
+  console.log(expTime)
+  return (new Date(expTime).getTime() - new Date().getTime()) / 1000
+}
+/**
+ * setCookieWithMaxAge(sKey, expiration, value)
+ * set a cookie
+ * @param {string} sKey - The name of the cookie
```
dgcohen, line 10: Code looks good but can you replace the JSDoc in the comment with typescript annotations?

src/utils/cookieUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @param {string} expiration - when to expire the cookie. Defaults to immediately expire
+ * @param {string} value - The value of the cookie, defaults to an empty string
+ */
+
+const setCookieWithExpiration = (sKey, expiration, value = "") => {
```
dgcohen, line 15: Where does the sKey name come from? Can this be renamed as just "key"?

src/utils/cookieUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+  const now = new Date()
+  now.setTime(now.getTime() + minutes * 60 * 1000 + seconds * 1000)
+  return now.toUTCString()
+}
+
+/**
+ * deleteCookie(sKey)
+ * Delete the cookie by having it expired.
+ *
+ * @param {string} sKey - The name of the cookie to be looked up.
```
dgcohen, line 31: Same comment as above for replacing JS doc with type annotations.

src/utils/cookieUtils.ts
```diff
@@ -0,0 +1,43 @@
+const includeDomain = process.env.NODE_ENV !== "test"
+
+export const getTimeLeft = (expTime) => {
+  console.log(expTime)
```
dgcohen, line 4: Remove this console.log

src/utils/cookieUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @param {string} value - The value of the cookie, defaults to an empty string
+ */
+
+const setCookieWithExpiration = (sKey, expiration, value = "") => {
+  document.cookie =
+    `${sKey}=${value}; expires=${expiration || Date.now()}; ` +
+    (includeDomain ? "path=/; domain=.nypl.org;" : "")
+}
+
+export const incrementTime = (minutes, seconds = 0) => {
```
dgcohen, line 21: Small nit but can you rename this to something like getIncrementedTime or getIncrementedCurrentTime or something like that? I think it would make it more legible since to me incrementTime suggests that there might be a side effect to an existing value.

src/components/MyAccount/TimedLogoutModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+            }${timeUntilExpiration.seconds}`}
+          </span>
+        </CardHeading>
+        <CardContent>
+          <CardActions className={styles.modalButtons}>
+            <Spacer />
+            <Button
+              buttonType="secondary"
+              onClick={logOutAndRedirect}
+              id="logout -button"
```
dgcohen, line 103: Remove whitespace here

src/components/MyAccount/TimedLogoutModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+  // Theoretically, accountPageExp should disappear after 5mins, causing
+  // logOutAndRedirect() to be fired above, but let's make sure a failure
+  // there never allows the timer to pass zero:
+  if (timeUntilExpiration.minutes <= 0 && timeUntilExpiration.seconds <= 0) {
+    logOutAndRedirect()
+  }
+  if (!open) return null
+
+  return (
+    <div
```
dgcohen, line 72: Use a <Box> here just to keep consistency with other components.

src/components/MyAccount/TimedLogoutModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+        backgroundColor="ui.bg.default"
+        className={styles.logoutModalBody}
+      >
+        <CardHeading
+          sx={{ display: "flex", justifyContent: "space-between" }}
+          subtitle="Do you want to stay logged in?"
+          size="heading6"
+        >
+          Your session is about to expire
+          <span>
```
dgcohen, line 91: Use a Box as="span" here just for consistency with other components.

dgcohen (changes requested):
Code looks good and works well! Just a few minor comments.

src/components/MyAccount/TimedLogoutModal.tsx
charmingduchess: Yea it's a little tricky because it needs the redirect uri. I could pull it out and make the uri a param. I may try for that.

src/utils/cookieUtils.ts
charmingduchess: it was just like that in DFE

src/components/MyAccount/TimedLogoutModal.tsx
charmingduchess: @EdwinGuzman do you want me to try and pull this out and try to get tests running here?

EdwinGuzman: Hmm i think it might be useful to test the side effects from calling this function, but I think you can mock the external router and deleteCookie function instead and verify that they're called. Can you try that first and if that doesn't work then you can try pulling this function out and running a check that this function was called rather than the two inside of it.

EdwinGuzman: Yes, I'd be okay with that.

dgcohen (approved):

PR #260, opened 2024-07-15
Scc 4189/add setters

charmingduchess:
## Ticket:

- JIRA ticket [SCC-4189](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4189)

- Adds a data refresh after any data submission (except for pin update). This includes: renewal, cancellation, updating pickup location, and any visible patron data. The fetch is happening in different places depending on the confirmation/modal flow for that action. I took the liberty of adding a skeleton loader to the hold cancel flow because otherwise the data refresh was quite jarring.

This required a pretty serious overhaul of  how My Account is working. All of the data that is displayed now is relying on a PatronDataContext, which is also responsible for fetching and updating the most recent my account data. Currently, the pickupLocations are also being refreshed unneccessarily. I wanted to get this done as a POC and then go back and refactor as needed. That seems to be the only update I would have to incorporate (I will probably work on it while I await the review of this PR).

I was moving pretty quickly through this, so I imagine there are some inconsistencies here and there in terms of what variables are being called and when props are being combined into one or not. Again, I was trying to get this off the ground quickly, so if anything stands out, please don't hesitate to be nitpicky.

## How has this been tested?

I added a prop to the context to facilitate testing. I created an optional `testSpy` prop on the Provider so I can pass in a spy. I was really struggling with the jest interface. Hopefully this is kosher enough to keep! 

- I have a feeling that updating data in the background as a side effect opens up some potential accessibility issues.  

### Checklist:

[SCC-4189]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


EdwinGuzman (commented):
I haven't reviewed the code yet but some somethings I noticed with patron `2 3333 09498 3077`:
- There were three checkouts that were renewable and they all had the "Due back by" date as "August 5, 2024". When I clicked "Renew", the message said it was successful and the new due date is "August 5, 2024". When it updated automatically (nice), the updated date is "August 6, 2024". This was true for the other items as well.
- The other checkouts are not renewable so I can't test renewing an a checkout that is already due. What is the time increment for a renewal?
- Canceling a request works well!
- Updating a patron's phone, email, or notification preference (home library does not appear as an option to update) are not updating after saving and it goes back to the previous value. When I refresh the page, the expected value appears. So this is working properly when updating the account but just not in the local version.

EdwinGuzman:
@charmingduchess yes, the latest update fixed the issue and now it's updating appropriately. I encountered this issue when going from any tab to the "Requests" tab, it looks like the locations aren't available before the tab panel renders:
Screenshot 2024-07-16 at 11 27 03 AM

I also don't always see the list of locations to update when I try to edit the account settings. Sometimes it's a blank space:
Screenshot 2024-07-16 at 11 36 43 AM

src/context/PatronDataContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+} from "../types/myAccountTypes"
+import { BASE_URL } from "../config/constants"
+/**
+ * Wrapper context component that controls state for the PatronData component
+ */
+export const PatronDataContext = createContext<PatronDataContextType | null>(
+  null
+)
+
+export const PatronDataProvider = ({
```
EdwinGuzman, line 14: This is great and make sense this is needed for this section of RC.

This is not a necessary change but just want to show a pattern I like to use which is to create a custom hook in this same file:

```
function usePatronDataContext() {
  const context = useContext(PatronDataContext);
  if (!context) {
    console.warn("usePatronDataContext must be used within PatronDataProvider");
    // throw error ?
  } 
  return context;
}
```

This context will only be used in `/account` so there's no worry about not having the provider, but it's a general pattern you might find in other react apps. This also reduces importing both `useContext` and `PatronDataContext` into every component where you need patron data.

```
const { updatedAccountData } = usePatronDataContext();
```

I'm not sure if this will make the tests any easier or more complicated. Try it and if it's not worth it, then scrap it.

EdwinGuzman (commented):
This looks good. Try the suggested custom hook and if it doesn't work out, then proceed  without it.

Since this is a big overhaul as mentioned, I think you're right about the accessibility concern. Let's try to move this over to QA so Clare can review and then you can refactor as needed, or you _could_ deploy this branch to QA now before it's merged in (but will need a few updates in the QA deploy script).

pages/api/account/[id].ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  const patronId = req.query.id as string
+  let responseMessage = ""
+  let responseStatus = 400
+  if (req.method !== "GET")
```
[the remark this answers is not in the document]
Comment (charmingduchess) [reply pages/api/account/[id].ts:25]: it's a get request! mistake is in the message. this is the patron data fetch, not an update.

src/components/MyAccount/RequestsTab/RequestsTab.test.tsx
```diff
@@ -106,27 +99,18 @@ describe("RequestsTab", () => {
       }
     )
     await userEvent.click(component.getAllByText("OK")[0])
-    bodyRows = component.getAllByRole("rowgroup")[1]
-    expect(within(bodyRows).getAllByRole("row").length).toBe(4)
+    // TODO: figure out how to verify
```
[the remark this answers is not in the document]
charmingduchess: i added a spy prop on the context, forgot to update this test to use it. just fixed.

pages/account/[[...index]].tsx
```diff
@@ -134,9 +109,11 @@ export async function getServerSideProps({ req, res }) {
   const tabsPath = match ? match[1] : null
   const id = patronTokenResponse.decodedPatron.sub
   try {
-    const client = await sierraClient()
-    const { checkouts, holds, patron, fines, pickupLocations } =
-      await MyAccountFactory(id, client)
+    const patronData = await fetchPatronData(id)
+    const { checkouts, holds, patron, fines, pickupLocations } = JSON.parse(
```
Comment (EdwinGuzman) [inline pages/account/[[...index]].tsx:113]: Hmm I don't like that you have to do JSON.parse to get the data you need. Ideally `fetchPatronData` would return the data directly and it would call `JSON.parse` internally. In this case, I'm not sure if you should set the expectation for the caller to use JSON.parse so perhaps this can be reverted. I was originally thinking that 
```
const client = await sierraClient()
    const { checkouts, holds, patron, fines, pickupLocations } =
      await MyAccountFactory(id, client)
```

would be a small function itself, but because it's two lines, maybe it's not worth it.

src/context/PatronDataContext.tsx
charmingduchess: I'm going to skip this for now.

pages/account/[[...index]].tsx
Comment (charmingduchess) [reply pages/account/[[...index]].tsx]: updated the method!

EdwinGuzman (approved):

PR #263, opened 2024-07-19
Scc 4196/update location modal bug

charmingduchess:
## Ticket:

- JIRA ticket [SCC-4196](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4196)

- Fixes the issue of the location not updating on the second opening of the modal. I solved it by using hold.pickupLocation.code as a key. When the key updates, the component completely rerenders from scratch (thanks [@thedocs](https://react.dev/learn/preserving-and-resetting-state#resetting-a-form-with-a-key)).
- because the component is completely rerendering, I had to replace the focus on the update button by having some state controlled outside of the updateLocation. this means I can probably replace the skeleton loader! which i will work on as i await review on this initial work.

## How has this been tested?

Moved focus tests to Request Tab tests. Added additional test to ensure value updates.

-

### Checklist:

### DIFF
--- a/src/components/MyAccount/RequestsTab/UpdateLocation.tsx
+++ b/src/components/MyAccount/RequestsTab/UpdateLocation.tsx
@@ -22,7 +22,7 @@ import { PatronDataContext } from "../../../context/PatronDataContext"
 interface UpdateLocationPropsType {
   hold: Hold
   pickupLocationOptions: SierraCodeName[]
-  key: number
+  key: string
   patronId: number
 }
 
--- a/src/components/MyAccount/RequestsTab/UpdateLocation.tsx
+++ b/src/components/MyAccount/RequestsTab/UpdateLocation.tsx
@@ -15,7 +15,7 @@ import {
 } from "@nypl/design-system-react-components"
 import type { Hold, SierraCodeName } from "../../../types/myAccountTypes"
 import styles from "../../../../styles/components/MyAccount.module.scss"
-import { useContext, useRef, useState, useEffect } from "react"
+import { useContext, useRef, useState, useEffect, type Dispatch } from "react"
 import { BASE_URL } from "../../../config/constants"
 import { PatronDataContext } from "../../../context/PatronDataContext"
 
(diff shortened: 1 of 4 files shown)


src/components/MyAccount/RequestsTab/UpdateLocation.tsx
```diff
@@ -15,28 +15,32 @@ import {
...
 import { PatronDataContext } from "../../../context/PatronDataContext"
 
 interface UpdateLocationPropsType {
   hold: Hold
   pickupLocationOptions: SierraCodeName[]
-  key: number
+  key: string
   patronId: number
+  setLastUpdatedHoldId: Dispatch<string>
+  focus: boolean
```
EdwinGuzman, line 28: How do you feel about updating this to `isFocused`? Makes it slightly more readable.

src/components/MyAccount/RequestsTab/UpdateLocation.tsx
```diff
@@ -166,7 +170,6 @@ const UpdateLocation = ({
       </Box>
     ),
     onClose: () => {
-      setModalProps(confirmLocationChangeModalProps as DefaultModalProps)
```
EdwinGuzman, line 169: You no longer need to do this because it resets on the next rerender, right?

EdwinGuzman (approved):
Looks good. Clever way to manage the focus when there are various holds and potential update location buttons for each hold, although it's quite coupled as a parent-child feature. In other words, it's a good update for this tab but won't work in other places (but I know that's the point since updating the location for a hold only happens here). The pattern itself is good for iterating over various items that each may need to manage focusing.

src/components/MyAccount/RequestsTab/UpdateLocation.tsx
charmingduchess: Ended up refactoring this to match the pattern I'm using with the loading states.

src/components/MyAccount/RequestsTab/UpdateLocation.tsx
charmingduchess: correct!

charmingduchess:
@EdwinGuzman Yea... it is very coupled, that's also how the tabs are generally. It's not ideal, but this pattern is proving very useful.

PR #277, opened 2024-08-01
SCC-4187 - Server render ItemTable

dgcohen:
## Ticket:

- JIRA ticket [SCC-4187](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4187)

- Removes the useEffect that was initializing the client-side fetch on first render.
- Passes the view_all_item query param in the server side fetch in the /all page component.

## How has this been tested?

- Tested locally, not sure if we need unit tests since server-rendering should be the default.

### Checklist:

[SCC-4187]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -63,8 +63,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Updated
 
+- Updated Bib Page to server-render the item table by default when accessed via the /all route (SCC-4187)
 - Created update patron settings functionality via the Settings tab in My Account. Currently only supports phone, email, and notification preference updates. (SCC-3928)
-
 - Extracted SearchResultsSort into its own component to render it in a different location on mobile.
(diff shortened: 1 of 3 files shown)


EdwinGuzman (approved):
Fantastic. This is a good start and what I mean about server-first approach. You removed code and, again, this should be the default way we build apps and then add additional client functionality.

I tested this by turning js off from the home page, searching, and landing on the bib page. The `/all` "new yorker" page with ~800 bib doesn't actually take too long to load.

What doesn't work is the `MultiSelect` dropdowns and pagination but those can be covered later (most likely in the DS).

charmingduchess (approved):

PR #280, opened 2024-08-06
SCC-4203 - Format Item filter data to work with DS components

dgcohen:
## Ticket:

- JIRA ticket [SCC-4203](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4203)

- Connects the new DS components with real filter data and updates change handlers to correctly update the filters with the new components.
- Refactored ItemTableData model (removes a couple of unused fields) and added a formattedFilterData function that returns the data in the format expected by the DS MultiSelect components.
- Fixed an issue where the ItemTable error message was showing when an in-progress re-fetch was aborted (i.e. when filters are updated during an in-flight request).
- Had to set the year in state due to e.target.value being undefined in jest tests.

## How has this been tested?

- Added integration tests for Item Filtering in Bib page tests.
- Added unit tests for ItemTableData

- The form currently doesn't work with JS disabled for several reasons (MultiSelect components are not expanded by default, Search bar is it's own independent form, etc), but I will open some follow-up ticket sto fix this.

### Checklist:

[SCC-4203]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/itemFilterUtils.ts
```diff
@@ -65,19 +66,34 @@ export const buildAppliedFiltersTagSetData = (
...
+  const newValues = appliedFilters[field].filter(
+    (filterValue: string) => filterValue !== id
+  )
+  return [newValues, field]
+}
+
+export const getSelectedCheckboxesFromAppliedFilters = (
+  appliedFilters: AppliedItemFilters
+): SelectedCheckboxes => {
+  return {
```
charmingduchess, line 92: I don't think every multiselect needs to know the selcted values for all the other multiselects. The docs are a little unclear, but they say
> The value is an object with a key of the MultiSelect component's id value set to an object with an "items" key. The "items" value is an array of the selected items. 

I think this method might be extraneous, and you can generate the `selectedItems` values dynamically per multiselect on line 156 in `ItemFilters.tsx`

src/models/ItemFilterData.ts
```diff
@@ -21,6 +21,16 @@ export class ItemFilterData {
     return this.options
   }
 
+  get formattedFilterData(): FilterCheckboxGroup {
+    return {
+      id: this.field,
+      name: capitalize(this.field),
+      items: this.displayOptions().map((option) => {
```
charmingduchess, line 28: Maybe you can update `displayOptions` to return objects in this `{id, name}` format. The method  only used here and in one other place.

charmingduchess (changes requested):
I think there are some extra methods and variables that can be eliminated, but over all looks solid!

src/components/ItemFilters/ItemFilters.tsx
```diff
@@ -37,6 +41,9 @@ const ItemFilters = ({
   appliedFilters = { location: [], format: [], status: [], year: [] },
   filtersAreApplied = false,
 }: ItemFilterContainerProps) => {
+  // We have to set the year value in state to be able to test form control in jest.
```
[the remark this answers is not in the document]
dgcohen: The problem was occurring because we were using event.target.value to grab the value entered into the year form, so the integration test for submitting the form and updating the router ("updates the router when year filter is submitted") was failing.

src/models/ItemFilterData.ts
dgcohen: Good call! Done

charmingduchess (approved):

src/components/ItemFilters/ItemFilters.tsx
```diff
@@ -114,19 +106,22 @@ const ItemFilters = ({
           renderMultiSelect={renderMultiSelect}
         />
         <Box minWidth="440">
-          <Label id="year-filter-label" htmlFor="year-filter">
+          <Label id="year-filter-label" htmlFor="searchbar-form-year-filter">
```
EdwinGuzman, line 109: If you're going to update `htmlFor` here, then the id for `SearchBar` should also be updated.

src/components/ItemFilters/ItemFilters.tsx
```diff
@@ -137,28 +132,27 @@ const ItemFilters = ({
...
+          id={`${checkboxGroup.id}`}
           data-testid={`${checkboxGroup.id}-multi-select`}
           items={checkboxGroup.items}
           key={checkboxGroup.id}
-          onChange={(e: React.ChangeEvent<HTMLInputElement>): void => {
-            console.log(e.target.value)
+          onChange={async (e: React.ChangeEvent<HTMLInputElement>) => {
+            await handleMultiSelectChange(e.target.id, checkboxGroup.id)
           }}
-          onClear={() => {
```
EdwinGuzman, line 151: You do need the `onClear` function. Right now the count button with the "X" does not work. You can click on it but it doesn't remove the selected filters.

EdwinGuzman: You probably need to use  `handleRemoveFilter` here.

src/models/ItemFilterData.ts
```diff
@@ -1,30 +1,40 @@
+import { capitalize } from "lodash"
```
EdwinGuzman, line 1: Eventually, try to remove `lodash` and create the functions locally. Lodash is a big package and I think unnecessary nowadays.

src/components/ItemFilters/ItemFilters.tsx
```diff
@@ -37,6 +38,9 @@ const ItemFilters = ({
   appliedFilters = { location: [], format: [], status: [], year: [] },
   filtersAreApplied = false,
 }: ItemFilterContainerProps) => {
+  // We have to set the year value in state to be able to test form control in jest.
+  // TODO: Remove this if we can find a better way to test form submissions in jest.
+  const [year, setYear] = useState(appliedFilters.year[0] || "")
```
EdwinGuzman, line 43: This seems fine, I think we wanted to keep track of the text value as state for controlled components.

EdwinGuzman (changes requested):
Some changes. I'm not sure if this is part of the update or data-related but there are two values in the Format `MultiSelect` that are not formats and cause issues rendering the count button:
Screenshot 2024-08-13 at 9 50 48 AM

Here's an example set of selected filters https://research-catalog-git-scc-4203-format-item-filter-da-0aed58-nypl.vercel.app/research/research-catalog/bib/b10833141?item_location=loc%3Amal82%2Cloc%3Arc2ma%2Cloc%3Arcma2&item_format=AUG.+23%2C+2021-CURRENT%2CText&item_status=status%3Ana

src/components/ItemFilters/ItemFilters.tsx
dgcohen: I changed this because that what gets autogenerated by the DS, the might be a more robust way to fix that.

dgcohen:
> Some changes. I'm not sure if this is part of the update or data-related but there are two values in the Format `MultiSelect` that are not formats and cause issues rendering the count button: Screenshot 2024-08-13 at 9 50 48 AM
> 
> Here's an example set of selected filters https://research-catalog-git-scc-4203-format-item-filter-da-0aed58-nypl.vercel.app/research/research-catalog/bib/b10833141?item_location=loc%3Amal82%2Cloc%3Arc2ma%2Cloc%3Arcma2&item_format=AUG.+23%2C+2021-CURRENT%2CText&item_status=status%3Ana

I believe this is a data issue (that also exists on DFE) that's caused by the format values having a comma, which is what we use as the multi-value delineator. There may be an app level solution to this, but we should probably ticket that separately.

src/components/ItemFilters/ItemFilters.tsx
dgcohen: great catch! had to add a new callback (handleClearFilterGroup) since handleRemoveFilter finds and removes a single filter value.

EdwinGuzman (approved):

PR #287, opened 2024-08-08
account settings bug and error handling

charmingduchess:
https://nypl.slack.com/files/U04TFQGQFTQ/F07GS083E7J/image.png
addressing this error which appeared in prod. it was happening due to a notification preference that was null.
also adds error handling to the entire return statement of my account page.

### DIFF
--- a/pages/account/[[...index]].tsx
+++ b/pages/account/[[...index]].tsx
@@ -37,46 +37,54 @@ export default function MyAccount({
     document.cookie = newExpirationTime
     setExpirationTime(inFive)
   }
-
+  const serverError = (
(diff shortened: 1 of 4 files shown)


EdwinGuzman (approved):
This is good as a hotfix but we should revisit the try/catch for the returned render HTML. The error is a syntax/data issue so I think this should still try to render the layout and other details of the page. The error should be caught lower in the DOM tree.

charmingduchess:
Yes I agree! But I also think nothing should be allowed to completely break the app. So to avoid unforeseen errors from doing so, this is a stopgap.

PR #292, opened 2024-08-09
Add launch profile for debugging in VSCode

dgcohen:

- Launch profile for debugging Next apps in VSCode, very convenient way to inspect variables with breakpoints instead of using console.log.


EdwinGuzman (approved):
Sometime this week, can you show us how you're using this?

PR #305, opened 2024-08-22
Scc 4233/add username to account header

charmingduchess:
## Ticket:

- JIRA ticket [SCC-4233](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4233)

- Fetches username from sierra
- displays username

## How has this been tested?

Updated unit tests and verified that usernames display. I don't have a patron account on hand with no username, so if anyone has one, that would be good to check!

-

### Checklist:

[SCC-4233]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/__test__/fixtures/processedMyAccountData.ts
+++ b/__test__/fixtures/processedMyAccountData.ts
@@ -317,7 +317,7 @@ export const filteredPickupLocations = [
 
 export const processedPatron = {
   notificationPreference: "z",
-  username: "spaghettimaster",
+  username: "pastadisciple",
   name: "Strega Nonna",
   barcode: "23333121538324",
   formattedBarcode: "2 3333 12153 8324",
--- a/__test__/fixtures/processedMyAccountData.ts
+++ b/__test__/fixtures/processedMyAccountData.ts
@@ -317,6 +317,7 @@ export const filteredPickupLocations = [
 
 export const processedPatron = {
   notificationPreference: "z",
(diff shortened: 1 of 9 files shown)


__test__/fixtures/processedMyAccountData.ts
```diff
@@ -317,6 +317,7 @@ export const filteredPickupLocations = [
 
 export const processedPatron = {
   notificationPreference: "z",
+  username: "spaghettimaster",
```
dgcohen, line 320: nit: change to "stregadisciple"

dgcohen (approved):
Looks good and working locally. Just one minor comment.

dgcohen:
Also out of curiosity, did we ever consider if patron has enough data and functionality to warrant its own class?

charmingduchess:
@dgcohen the data is pretty static so i don't think it needs to be a class.

dgcohen:
> @dgcohen the data is pretty static so i don't think it needs to be a class.

Yeah that makes sense, though you can argue the same for some of the other classes we have. It's definitely out of scope of this PR but it might be a good idea for us to start thinking about criteria for what goes in a model vs a utils file.

PR #316, opened 2024-09-06
SCC-4264 - VQA Second Pass

dgcohen:
## Ticket:

- JIRA ticket [SCC-4264](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4264)

-

-

### Checklist:

### DIFF
--- a/pages/test-pickup-times/index.tsx
+++ b/pages/test-pickup-times/index.tsx
@@ -120,7 +120,7 @@ const TestPickupTimes = (params) => {
               <Select
                 id="scenario"
                 name="scenario"
-                labelText="Preloaded Scenarios"
+                labelText="Preloaded scenarios"
                 value={scenario}
                 onChange={(e) =>
                   setScenario((e.target as HTMLInputElement).value)
@@ -138,7 +138,7 @@ const TestPickupTimes = (params) => {
 
               <TextInput
                 id="holdingLocation"
-                labelText="Item Location"
+                labelText="Item location"
                 name="holdingLocation"
                 value={holdingLocation}
                 onChange={(e) => setHoldingLocation(e.target.value)}
@@ -163,7 +163,7 @@ const TestPickupTimes = (params) => {
               </Select>
               <TextInput
                 id="deliveryLocation"
-                labelText="Delivery Location (opt)"
+                labelText="Delivery location (opt)"
                 name="deliveryLocation"
                 value={deliveryLocation}
                 onChange={(e) => setDeliveryLocation(e.target.value)}
--- a/src/components/ItemFilters/ItemFilters.tsx
+++ b/src/components/ItemFilters/ItemFilters.tsx
@@ -173,14 +173,18 @@ const ItemFilters = ({
         />
       </Box>
       {filtersAreApplied ? (
-        <Box display="flex" mr="s" mb="m">
+        <Box
+          display="flex"
+          flexDirection={{ base: "column", md: "row" }}
+          mb="m"
+        >
           <Text
-            fontSize="body2"
+            display="block"
+            fontSize="desktop.body.body2"
             fontWeight="bold"
-            mr="s"
-            mb={0}
+            mr={{ base: "0", md: "s" }}
+            mb={{ base: "xxs", md: 0 }}
             lineHeight={2}
-            display="table-cell"
           >
             Active filters
           </Text>
--- a/styles/components/BibDetails.module.scss
+++ b/styles/components/BibDetails.module.scss
@@ -6,6 +6,13 @@
     border: none;
     margin-bottom: var(--nypl-space-m);
   }
+
+  dt {
+    @media (max-width: 600px) {
+      margin-bottom: var(--nypl-space-xxs);
+    }
+  }
+
   dl dt,
   dl dd {
     padding-top: 0;
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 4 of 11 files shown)


charmingduchess:
I'm seeing title casing on the item table headings in desktop mode
Screenshot 2024-09-18 at 4 37 18 PM

charmingduchess (approved):

__test__/pages/bib/bibPage.test.tsx
```diff
@@ -168,11 +168,41 @@ describe("Bib Page Item Table", () => {
     await userEvent.click(yearFilter.querySelector("button[type='submit']"))
 
     expect(mockRouter.asPath).toBe("/bib/pb5579193?item_date=2005")
+  })
+
+  it("clears the year filter when the year tag is clicked", async () => {
+    const yearTag = screen.queryByText("Year > 2005").closest("button")
+    await userEvent.click(yearTag)
+
+    expect(mockRouter.asPath).toBe("/bib/pb5579193")
```
EdwinGuzman, line 177: Hmm I just noticed that these tests depend on the previous test having run and pass. If this test ever gets commented out, skipped, or deleted, the next test will fail. Not something to fix now but tests should be as isolated as possible, unless if there is a need to make them sequential.

__test__/pages/bib/bibPage.test.tsx
```diff
@@ -168,11 +168,41 @@ describe("Bib Page Item Table", () => {
...
+    await userEvent.type(yearField, "ABCD")
+    await userEvent.click(yearFilter.querySelector("button[type='submit']"))
+    expect(mockRouter.asPath).toBe("/bib/pb5579193")
+    expect(
+      screen.queryByText("Error: Please enter a valid year.")
+    ).toBeInTheDocument()
 
-    // Clear the year field
-    await userEvent.click(yearFilter.querySelector("button[type='button']"))
+    // not of length 3
```
EdwinGuzman, line 199: Do you mean 4?

pages/test-pickup-times/index.tsx
```diff
@@ -120,7 +120,7 @@ const TestPickupTimes = (params) => {
               <Select
                 id="scenario"
                 name="scenario"
-                labelText="Preloaded Scenarios"
+                labelText="Preloaded scenarios"
```
EdwinGuzman, line 123: I know you're making titles sentence case but this is out of scope.

src/components/ItemFilters/ItemFilters.test.tsx
```diff
@@ -8,75 +8,101 @@ import Bib from "../../models/Bib"
...
-    const outsideOfTheFilter = screen.getByLabelText("Search by year")
+    it("renders the year search form when showDateFilter is true", async () => {
+      expect(screen.getByLabelText("Search by year")).toBeInTheDocument()
+    })
 
-    const locationFilterButton = screen
-      .getByTestId("location-multi-select")
-      .querySelector("button")
-    await userEvent.click(locationFilterButton)
+    it("closes open filters when user clicks outside of the filter", async () => {
```
EdwinGuzman, line 47: This is fine to keep here but also note that this is an implementation detail of the component, no? And this test already exists in the DS (in a similar way) so there's no need to replicate the test here, but okay to keep if you want to.

src/components/ItemFilters/ItemFilters.tsx
```diff
@@ -106,7 +111,14 @@ const ItemFilters = ({
 
   const handleYearSubmit = async (e: SyntheticEvent) => {
     e.preventDefault()
-    await submitFilters(year.length ? [year] : [], "year")
+    const isNumeric: boolean = /^[0-9]*$/.test(year)
+
+    if (year.length !== 4 || !isNumeric) {
```
EdwinGuzman, line 116: Double-check this. @charmingduchess I vaguely remember that there's a need for searching 3 digit years, right?

styles/components/BibDetails.module.scss
```diff
@@ -6,6 +6,13 @@
     border: none;
     margin-bottom: var(--nypl-space-m);
   }
+
+  dt {
+    @media (max-width: 600px) {
+      margin-bottom: var(--nypl-space-xxs);
```
EdwinGuzman, line 12: I think this should be `xs` for "8px"

EdwinGuzman (changes requested):
This was not addressed:
> 2. Would it be possible for ‘active filters’ to remain in one line instead of wrapping like this?

See the QA doc for screenshot. It seems this PR is mostly for updating tests and conditionally rendering the table but most of the VQA feedback will be done after the next DS release, right? (Specifically for the `Table` component). I might be missing more context. But otherwise, the updated tests look good.

__test__/pages/bib/bibPage.test.tsx
dgcohen: I agree. I noticed that too but didn't fix since it's out of scope. Will open a ticket to fix this.

src/components/ItemFilters/ItemFilters.test.tsx
dgcohen: Correct, this is a holdover from when this functionality was in research-catalog. This diff indicates that I moved it, but can remove it as part of this PR since it doesn't need to test the component's implementation.

src/components/ItemFilters/ItemFilters.tsx
dgcohen: We are currently validating for 4 characters in DFE, but it makes sense to me that 3 digit years (or fewer) would be valid.

src/components/ItemFilters/ItemFilters.test.tsx
EdwinGuzman: Gotcha. That wasn't clear. So then these tests were added back when this was not a default feature of the DS and was app-specific. I think they can be removed since the feature and tests are in the DS component.

EdwinGuzman (approved):

PR #321, opened 2024-09-18
add search tips

charmingduchess:
## Ticket:

- JIRA ticket [SCC-4623](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4263)

- add search tips to search bar per search scope option

- this seems like something that we would definitely want someone using a screen reader to be able to hear. how do we ensure that happens?

### Checklist:

### DIFF
--- a/src/utils/searchUtils.ts
+++ b/src/utils/searchUtils.ts
@@ -14,6 +14,55 @@ import { RESULTS_PER_PAGE } from "../config/constants"
 import { collapseMultiValueQueryParams } from "./refineSearchUtils"
 import { getPaginationOffsetStrings } from "./appUtils"
 
+export const searchFormDropDownOptionsWithSearchTip = [
+  {
+    text: "Keyword",
+    value: "all",
+    searchTip: "Enter a title, or use quotes to search for an exact phrase.",
+  },
+  { text: "Title", value: "title", searchTip: "title tip" },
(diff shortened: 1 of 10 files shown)


src/utils/searchUtils.ts
```diff
@@ -14,6 +14,77 @@ import { RESULTS_PER_PAGE } from "../config/constants"
 import { collapseMultiValueQueryParams } from "./refineSearchUtils"
 import { getPaginationOffsetStrings } from "./appUtils"
 
+export const searchFormOptions = [
```
dgcohen, line 17: Should this go in constants.ts? I think we should keep mostly functions in util files for consistency.

src/utils/searchUtils.ts
```diff
@@ -14,6 +14,77 @@ import { RESULTS_PER_PAGE } from "../config/constants"
...
+]
+
+export const searchFormSelectOptions = searchFormOptions.map(
+  ({ text, value }) => ({
+    text,
+    value,
+  })
+)
+
+const propForObjectInArray = (property, arrayOfObjects) => {
```
dgcohen, line 75: Could we make searchFormOptions a hashmap instead of an array to cut down on the functions that iterate over all the options?

dgcohen (approved):

PR #328, opened 2024-09-23
SCC-4287 - Incorporate View All Endpoint

dgcohen:
## Ticket:

- JIRA ticket [SCC-4287](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4287)

- Integrate view_all query param on client side
- Only call batched fetch when item filters are active
- Added TODO notes for changes to be made when filter params are supported by the all_items endpoint in discovery.

## How has this been tested?

- Updated getBibQueryString unit tests and removed those specific to batched fetch.

- NA

### Checklist:

[SCC-4287]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/server/api/bib.ts
+++ b/src/server/api/bib.ts
@@ -154,17 +154,17 @@ async function fetchAllBibItemsWithQuery(
   const totalBatchNum = Math.ceil(numItems / batchSize)
 
   for (let batchNum = 1; batchNum <= totalBatchNum; batchNum++) {
-    const pageQueryString = getBibQueryString(
-      {
-        ...bibQuery,
-        item_page: batchNum,
-      },
-      false
-    )
-    const bibPage = await client.get(
-      `${DISCOVERY_API_SEARCH_ROUTE}/${bibQuery.id}${pageQueryString}`
-    )
     try {
+      const pageQueryString = getBibQueryString(
+        {
+          ...bibQuery,
+          item_page: batchNum,
+        },
+        false
+      )
+      const bibPage = await client.get(
+        `${DISCOVERY_API_SEARCH_ROUTE}/${bibQuery.id}${pageQueryString}`
+      )
       if (bibPage?.items?.length) {
         items.push(...bibPage.items)
       } else {
--- a/src/server/api/bib.ts
+++ b/src/server/api/bib.ts
@@ -101,11 +101,12 @@ export async function fetchBib(
       "item_status",
       "item_date",
     ]
-    // Only call the batched fetch when some of the filters are active
-    if (
+    const allItemsAndFiltersActive =
       bibQuery?.all_items &&
-      itemFilterParams.some((param) => param in Object.keys(bibQuery))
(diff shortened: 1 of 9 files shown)


src/utils/bibUtils.ts
```diff
@@ -104,12 +99,13 @@ export function getBibQueryString(
         .join("")
     : ""
 
-  const paginationQuery = `items_size=${batchSize}&items_from=${itemsFrom}&item_page=${itemPage}`
+  const paginationQuery = bibQuery?.all_items
```
EdwinGuzman, line 102: I think a better name would be `paginationOrAllQuery`

EdwinGuzman (approved):
Minor update but otherwise looks good

src/utils/bibUtils.ts
dgcohen: Sounds good i like that naming better

EdwinGuzman:
Why did you have to restore the batched fetch when filter query params are present?

Also, I just noticed the fetching function throws an error but the callee doesn't catch it. That'll break the page so catch it and log the error to the console as well.

charmingduchess (changes requested):
Looks good! just some small updates

src/server/api/bib.ts
```diff
@@ -91,9 +93,13 @@ export async function fetchBib(
         status: 404,
       }
     }
+
     // Populate bib's items with all the items if View All is enabled
-    if (bibQuery?.view_all_items) {
-      discoveryBibResult.items = await fetchAllBibItems(
+    // TODO: Remove this when view_all endpoint in discovery supports query params
```
EdwinGuzman, line 98: Do you still need this comment here?

src/server/api/bib.ts
```diff
@@ -137,33 +143,37 @@ async function fetchBibSubjectHeadings(bibId: string) {
...
+      false
     )
     const bibPage = await client.get(
       `${DISCOVERY_API_SEARCH_ROUTE}/${bibQuery.id}${pageQueryString}`
     )
-    if (bibPage?.items?.length) {
-      items.push(...bibPage.items)
-    } else {
-      throw new Error("There was en error fetching items in one of the batches")
+    try {
```
EdwinGuzman, line 167: The `client.get` call should go in the try statement. Because if that call fails and throws an error, it's still not caught. What's being tested here is if `bibPage.items` has values but that wouldn't really crash.

EdwinGuzman (commented):
One minor update. If you haven't already, create a ticket to remove code/tests when the view_all endpoint supports query params.

charmingduchess (approved):

EdwinGuzman (approved):

PR #329, opened 2024-09-26
SCC-4291 - Populate Bib's subjectHeadings with SHEP data if available

dgcohen:
## Ticket:

- JIRA ticket [SCC-4291](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4291)

- Adds subjectHeadings attribute to Bib model and populates it if present in the discovery result
- Renames old buildSubjectHeadings in bibDetails model to buildSubjectLiterals and introduces new buildSubjectHeadings function to format subjectHeading links correctly.

## How has this been tested?

- Added unit tests

- NA

### Checklist:

[SCC-4291]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


EdwinGuzman:
Looks good but do you have a few bibs and example for testing?

dgcohen:
> few bibs and example for testing?

You can test on any bib with subject headings locally with VPN enabled (urls should fall back to /search with the subjectLiteral query param when VPN is disabled)

Here is the bib i've been using to test: http://local.nypl.org:8080/research/research-catalog/bib/b10833141

Just verify if the Subject heading URLs match those on prod.

EdwinGuzman:
Hmm I'm not exactly seeing this work. I'll test this again next week but this type of "how to review/QA" will also be helpful once it's passed on to QA.

EdwinGuzman (approved):

src/models/BibDetails.ts
```diff
@@ -362,6 +364,36 @@ export default class BibDetails {
   }
 
   buildSubjectHeadings(): SubjectHeadingDetail {
+    if (!this.bib.subjectHeadings) return
```
charmingduchess, line 367: `buildSubjectHeadings` and `buildSubjectLiterals` have a lot of duplicated logic. It looks like the main logical difference is the check for uuid on line 369, and otherwise the split is which key it's mapping over and the url string. Can the duplicated logic be extracted? 
Also, I think it would add some clarity to include `"SHEP"` in the subject headings function name, as well as include `"links"` or `"urls"` in both function names.

charmingduchess (changes requested):
Does this code require an update to the bib details page to use the new methods created here?

dgcohen:
> Does this code require an update to the bib details page to use the new methods created here?

No it should just work if the subject headings are fetched and assigned to the bib.

src/models/BibDetails.ts
dgcohen: These are good suggestions but I would like to do these refinements after the release since they don't affect functionality.

charmingduchess (approved):

PR #330, opened 2024-09-27
Scc 4260/rename fields

charmingduchess:
## Ticket:

- Noref

- Update label for identifiers
- fix hydration error


dgcohen (approved):

src/config/constants.ts
```diff
@@ -78,7 +78,6 @@ export const SEARCH_FORM_OPTIONS = {
   },
   title: {
     text: "Title",
-    value: "title",
```
dgcohen, line 81: Good catch

PR #332, opened 2024-09-27
Bump DS to 3.4.0-rc and nvmrc to node 20

dgcohen:
Some version updates. Didn't mention the DS bump in the changelog since we do that regularly, but let me know if you'd like me to add a note.


EdwinGuzman (approved):
Yes, please do flag the ds version bump in the changelog. It's still a release candidate and not a stable version so it's ok to promote to QA but not prod. But I think this is supposed to go out 10/2 and the next DS release is 10/3. Can you ask Marty if that's the right timeline?

EdwinGuzman (commented):
Make sure you also flag this with Chris/Alkim for the QA process since this is a change and it might impact their automated tests (pointing it out since this caused local unit tests to fail).

EdwinGuzman (approved):

EdwinGuzman:
Add for the `Table` component - if it's on mobile add the `isScrollable` prop, otherwise, don't set it.

PR #336, opened 2024-09-30
Hotfix/view fewer link not rendering

dgcohen:

- Updates numItems on Bib model to account for filtersAreApplied
- This was causing the showViewAllItemsLink to always be false since it wasn't accounting for the filters being applied or not.

## How has this been tested?

- We have some tests for this but we didn't account for all the cases. Will add these as a follow up.

### Checklist:

### DIFF
--- a/src/models/SearchResultsBib.ts
+++ b/src/models/SearchResultsBib.ts
@@ -26,7 +26,7 @@ export default class SearchResultsBib extends Bib {
       : null
   }
 
-  get showViewAllItemsLink() {
+  showViewAllItemsLink() {
     return this.numPhysicalItems > ITEMS_PER_SEARCH_RESULT
   }
 
--- a/src/models/SearchResultsBib.ts
+++ b/src/models/SearchResultsBib.ts
@@ -43,6 +43,12 @@ export default class SearchResultsBib extends Bib {
       : null
   }
 
+  numItems() {
+    return this.hasPhysicalItems
+      ? this.numPhysicalItems
+      : this.numElectronicResources
+  }
+
   getYearFromResult(result: DiscoveryBibResult) {
     const { dateStartYear, dateEndYear } = result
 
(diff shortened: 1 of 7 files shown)


charmingduchess (approved):
Looks good! Makes sense to move that logic to the bib. Feels like some solid OOP. I think this work shows that you were well set up with the bib model.

src/models/SearchResultsBib.ts
```diff
@@ -43,6 +43,12 @@ export default class SearchResultsBib extends Bib {
       : null
   }
 
+  numItems() {
+    return this.hasPhysicalItems
```
EdwinGuzman, line 47: Why not `this.hasPhysicalItems ?? this.numElectornicResources`?

EdwinGuzman (approved):
The update makes sense and works and I'm approving this to get this deployed, though tests would be helpful.

The update is now coupling a boolean flag that is passed in to return the correct number of items. It feels like the kind of logic that the user of the object should do rather than the object itself.

src/models/SearchResultsBib.ts
dgcohen: just moved this logic without changing it, but can clean it up if you'd like.

do you mean this.numPhysicalItems || this.numElectronicResources ?

dgcohen:
> The update makes sense and works and I'm approving this to get this deployed, though tests would be helpful.
> 
> The update is now coupling a boolean flag that is passed in to return the correct number of items. It feels like the kind of logic that the user of the object should do rather than the object itself.

I agree about tests, was planning to add some more robust ones as a follow-up.

As for where this logic should live, it could make sense to move it to the component but I'd like to do that as a follow up as well.

src/models/SearchResultsBib.ts
EdwinGuzman: Hmm no. I just noticed that it's `has` and `num` + `PhysicalItems` and that made it hard to read. So ignore the comment.

PR #340, opened 2024-10-01
Bump DS version and fix minor styling issues for Bib Page deployment

dgcohen:
This PR includes the following changes needed for the Bib Page release

-  Bump DS version
- Add column minWidths
- Fix subtle text alignment difference between ItemTable columns


CHANGELOG
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Updated
 
+- Bump DS version to 3.4.0
 - Bump nvmrc version to Node 20 and DS package to release candidate 3.4.0-rc.
```
EdwinGuzman, line 26: I think you it's ok to remove "and DS package to release candidate 3.4.0-rc." since this should not really go into `production` and was used for testing. So you can leave the 3.4.0 message above.

EdwinGuzman (approved):
The update looks good. There are some issues but not blockers. They should be addressed soon though. Some of the focus rings get cut off on the left side of the page. It doesn't seem related to the CSS tweaks but rather the DS. I'll talk to Marty but try to address them in the app if possible:

**Search page**
Screenshot 2024-10-02 at 6 45 27 PM

After you make a search, the focus ring displays as a horizontal line and not a box:
Screenshot 2024-10-02 at 6 45 02 PM

**Bib page**
Screenshot 2024-10-02 at 6 44 22 PM

PR #341, opened 2024-10-02
SCC-4193 - Optimize state in bib page

dgcohen:
## Ticket:

- JIRA ticket [SCC-4193](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4193)

- Replaces setting of whole Bib in state with numItems and itemTableData, which are the only pieces of data that need to be refreshed.
- While the Bib model still needs to get re-initialized based on the way new item table data is fetched, it's not necessary to keep the whole bib in state.

## How has this been tested?

NA

NA

### Checklist:

[SCC-4193]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


charmingduchess (commented):
Looks good. I'm wondering if there is a performance issue here as the code stands?

src/components/ItemTable/ItemTableControls.tsx
```diff
@@ -38,8 +38,9 @@ const ItemTableControls = ({
   numItemsTotal = 0,
   filtersAreApplied = false,
 }: ItemTableControlsProps) => {
-  const viewAllLoadingMessage =
-    bib.getItemsViewAllLoadingMessage(filtersAreApplied)
+  const viewAllLoadingMessage = `Loading all ${
+    filtersAreApplied ? "matching" : numItemsTotal
+  } items. This may take a few moments...`
```
EdwinGuzman, line 43: `getItemsViewAllLoadingMessage` is only used here, why not change it instead of moving the functionality here? Or at least delete the declaration from the Bib model.

EdwinGuzman (changes requested):
Looks good just one question

src/components/ItemTable/ItemTableControls.tsx
dgcohen: Good catch, i removed it in the Bib model since it's only used here, but can add it back if you'd like greater separation of logic from presentation.

CHANGELOG
```diff
@@ -35,8 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Updated
 
-- Bump DS version to 3.4.0
-- Bump nvmrc version to Node 20.
+- Bump nvmrc version to Node 20 and DS package to release candidate 3.4.0-rc.
 - Update error message in Item Filter Year to include new error message prefix from the DS.
```
EdwinGuzman, line 45: Revert this so that it does say `3.4.0` and the node 20 update as separate lines and not have the `-rc` indicated here.

src/components/ItemTable/ItemTableControls.tsx
EdwinGuzman: Either approach is ok but I brought it up just as a question.

EdwinGuzman (approved):
Just minor update to the changelog

EdwinGuzman (approved):

dgcohen:
> Looks good. I'm wondering if there is a performance issue here as the code stands?

Nothing noticeable but Edwin suggested this change because we don't need to refresh the whole bib in state. @EdwinGuzman would it be worth performing benchmark tests to see if this is necessary?

EdwinGuzman:
@dgcohen yes, that'd be great. How about we keep this on hold for a bit and you run some perf tests (lighthouse or another tool) with the current site against this PR? The difference may be negligible but code-wise, I think this is a helpful update.

PR #345, opened 2024-10-03
Fix table column widths on mobile

dgcohen:


src/components/ItemTable/ItemTable.tsx
```diff
@@ -28,18 +28,26 @@ const ItemTable = ({ itemTableData }: ItemTableProps) => {
...
+                { width: "33.3%", minWidth: 150, maxWidth: 250 },
+                { width: "33.3%", minWidth: 150, maxWidth: 250 },
+              ]
+            : [
+                { minWidth: 350, maxwidth: 350 },
+                { minwidth: 150, maxWidth: 200 },
+                { minwidth: 150, maxWidth: 150 },
+                { minwidth: 150, maxWidth: 150 },
+                { minwidth: 200, maxWidth: 250 },
+                { minwidth: 150, maxWidth: 200 },
```
bigfishdesign13, line 44: `minWidth` and `maxWidth` should use camel case. The all lowercase props are not being applied properly.

PR #358, opened 2024-10-10
 SCC-4253: Edit email in My Account 2.0 

7emansell:
## Ticket:

Resolves JIRA ticket [SCC-4253](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4253).

- Creates individually editable form component for updating user's email addresses in My Account, `EmailForm`
- Writes tests for `EmailForm`

NOTE!! With the expectation this will only be merged in once My Account 2.0 work is complete, the email form is not integrated with the rest of the existing fields, it's just standalone for now in `MyAccount/NewSettings`. I'll remove `MyAccount/Settings` once everything has been migrated over. 

## How has this been tested?

Locally

### Checklist:

[SCC-4253]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


EdwinGuzman:
Regarding your note, make a feature branch (maybe `my-account-2.0`?) and point this PR into that branch. It'll have to be updated with `main` every so often but you can merge and close your PRs and ticket while everything is complete and ready for QA, VQA, etc.

src/components/MyAccount/NewSettings/EmailForm.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+      if (response.status === 200) {
+        await getMostUpdatedSierraAccountData()
+        setIsSuccess(true)
+        setIsEditing(false)
+      } else {
+        setTempEmails([...emails])
+      }
+    } catch (error) {
+      console.error("Error submitting emails:", error)
```
EdwinGuzman, line 96: Similar to the success banner, if there's an error the user should know something went wrong. The current banner only displays the success message so that will need an update.

This is not in the req/ac for this ticket so it could be separate work.

src/components/MyAccount/NewSettings/EmailForm.tsx
```diff
@@ -0,0 +1,207 @@
...
+    } finally {
+      setIsLoading(false)
+    }
+  }
+
+  return (
+    <Flex flexDir="row" alignItems="flex-start" width="100%">
+      <Flex gap="xs" alignItems={"center"} sx={{ paddingTop: "xs" }}>
```
EdwinGuzman, line 104: nit, you don't need the braces around "center".

EdwinGuzman: Also, does `paddingTop="xs"` not work?

src/components/MyAccount/NewSettings/EmailForm.tsx
```diff
@@ -0,0 +1,207 @@
...
+            onClick={() => setIsEditing(true)}
+            sx={{ paddingLeft: "xs", paddingRight: "xs", marginLeft: "xxl" }}
+          >
+            <Icon name="editorMode" align="left" size="medium" />
+            Edit
+          </Button>
+        </Flex>
+      )}
+      {isEditing && (
+        <Flex justifySelf="flex-end" sx={{ marginLeft: "auto" }}>
```
EdwinGuzman, line 184: Can you try using the `ButtonGroup` component? If you already did, did it not work? It has a default gap between the buttons so you shouldn't need the margins in the first button.

src/components/MyAccount/NewSettings/EmailForm.tsx
```diff
@@ -0,0 +1,207 @@
...
+            + Add an email address
+          </Button>
+        </Flex>
+      ) : (
+        <Flex flexDir="row" alignItems="flex-start">
+          <Flex flexDir="column" alignItems="flex-start">
+            {emails.map((email, index) => (
+              <Text key={index} sx={{ paddingTop: "xs", marginBottom: "xs" }}>
+                {email}{" "}
+                {index === 0 && <span style={{ color: "gray" }}>(P)</span>}
```
EdwinGuzman, line 168: This `(P)` is not in the designs (or I can't see it). Use `ui.gray.semi-dark` (perhaps?) or another design token gray value.

src/components/MyAccount/NewSettings/EmailForm.tsx
```diff
@@ -0,0 +1,207 @@
...
+                isInvalid={error && !validateEmail(email)}
+                invalidText="Please enter a valid email address."
+                onChange={(e) => handleInputChange(e, index)}
+                isRequired
+                isClearable
+                isClearableCallback={() => handleClearableCallback(index)}
+                sx={{ width: "300px" }}
+              />
+              {index !== 0 && (
+                <Button
```
EdwinGuzman, line 136: Minorish but the focus ring floats over the input field:
Screenshot 2024-10-10 at 3 18 57 PM

src/components/MyAccount/NewSettings/EmailForm.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+  it("allows adding a new email field", () => {
+    render(component)
+
+    fireEvent.click(screen.getByRole("button", { name: /edit/i }))
+
+    fireEvent.click(
+      screen.getByRole("button", { name: /\+ add an email address/i })
+    )
+
+    expect(screen.getAllByLabelText("Update email").length).toBe(3)
```
charmingduchess, line 83: this could be made less brittle if you expect it to be `patron.emails.length + 1`

charmingduchess (commented):
Looks great! A couple minor in line suggestions, and two larger ones:
- Can this PR target a feature branch? I don't think we want this in main until the whole feature is ready to be QA'd. 
- Second, I noticed you have the account settings loading layer over the entire tab, like it was in the original implementation. Now that all the inputs are going to be separate forms, can you update the code so the `ProgressIndicator` obscures just the updating field?  I built out the `PatronDataContext` with its own `patronDataLoading` state to facilitate that. 
You can see an example implementation of this idea [here](https://github.com/NYPL/research-catalog/blob/main/src/components/MyAccount/CheckoutsTab/RenewButton.tsx#L128).

src/components/MyAccount/NewSettings/EmailForm.tsx
7emansell: It is but I think it only appears when there's more than one email. Color updated

src/components/MyAccount/TimedLogoutModal.tsx
```diff
@@ -41,12 +41,12 @@ const TimedLogoutModal = ({
     buildTimeLeft(expirationTime)
   )
 
-  // if (
-  //   typeof document !== "undefined" &&
-  //   !document.cookie.includes("accountPageExp")
-  // ) {
-  //   logOutAndRedirect()
-  // }
+  if (
```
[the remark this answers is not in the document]
7emansell: Was trying to turn off automatic log out when I was working and over-un-commented

src/components/MyAccount/NewSettings/EmailForm.tsx
7emansell: Added a very basic failure banner, "Your changes were not saved"... should I ask Apoorva for copy?

src/components/MyAccount/NewSettings/EmailForm.tsx
```diff
@@ -0,0 +1,207 @@
...
+    setError(hasInvalidEmail)
+  }
+
+  const handleAddEmail = () => {
+    const updatedEmails = [...tempEmails, ""]
+    setTempEmails(updatedEmails)
+
+    // Immediately revalidate all emails.
+    const hasInvalidEmail = updatedEmails.some((email) => !validateEmail(email))
+    setError(hasInvalidEmail)
```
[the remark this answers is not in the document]
7emansell: Discussed offline with @charmingduchess and I think this is expected– it is invalid when the field is empty, and invalid until a valid email is entered. I think it's more trouble than it's worth to start litigating at what character the user actually has an error and at what character we believe they're still just typing

src/components/MyAccount/NewSettings/EmailForm.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+    fireEvent.change(input, { target: { value: "modified@example.com" } })
+
+    fireEvent.click(screen.getByRole("button", { name: /cancel/i }))
+
+    expect(screen.getByText("streganonna@gmail.com")).toBeInTheDocument()
+    expect(
+      screen.queryByDisplayValue("modified@example.com")
+    ).not.toBeInTheDocument()
+  })
+})
```
EdwinGuzman, line 138: Very nice set of tests.

src/components/MyAccount/NewSettings/EmailForm.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+      />
+    </PatronDataProvider>
+  )
+
+  it("renders correctly with initial email", () => {
+    render(component)
+
+    expect(screen.getByText("streganonna@gmail.com")).toBeInTheDocument()
+
+    expect(screen.getByRole("button", { name: /edit/i })).toBeInTheDocument()
```
dgcohen, line 41: Is there a reason you're using a regex here? Not concerned, just curious.

src/components/MyAccount/NewSettings/EmailForm.tsx
EdwinGuzman: Add the basic failure banner with the suggested copy and we'll run it by her during VQA.

src/components/MyAccount/NewSettings/NewAccountSettingsTab.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const {
+    updatedAccountData: { patron },
+  } = useContext(PatronDataContext)
+  const [isSuccess, setIsSuccess] = useState(false)
+  const [isFailure, setIsFailure] = useState(false)
+
+  return (
+    <>
+      {isSuccess && (
+        <Banner
```
EdwinGuzman, line 16: The two `Banner`s are very similar expect for the text. Can you try to make the text and type conditional based on `isSuccess` and `isFailure` so that the component isn't repeated?

src/components/MyAccount/NewSettings/SaveCancelButtons.tsx
```diff
@@ -0,0 +1,33 @@
+import { ButtonGroup, Button } from "@nypl/design-system-react-components"
+
+const SaveCancelButtons = ({ error, onCancel, onSave }) => {
```
EdwinGuzman, line 3: How about `isDisabled` instead of `error`? Give these types as well.

src/components/MyAccount/NewSettings/SettingsFormLabel.tsx
```diff
@@ -0,0 +1,21 @@
+import { Flex, Icon, Text } from "@nypl/design-system-react-components"
```
EdwinGuzman, line 1: Is the syntax colors gone from this file? Not a real issue but just odd.

dgcohen (approved):
Looks good and working for me locally. One small question, are the email field and add button supposed to be full width on mobile?

Screenshot 2024-10-16 at 10 17 44 AM

src/components/MyAccount/NewSettings/EmailForm.tsx
```diff
@@ -0,0 +1,230 @@
...
+        >
+          <SettingsFormLabel icon="communicationEmail" text="Email" />
+          {isEditing ? (
+            <Flex
+              marginLeft={{ base: "l", lg: "unset" }}
+              marginTop={{ base: "xs", lg: "unset" }}
+              flexDir="column"
+              width="-webkit-fill-available"
+            >
+              {tempEmails.map((email, index) => (
```
EdwinGuzman, line 127: Can you also wrap this in a `Form` or HTML `<form>` element?
If you use the DS one, it might make it easier to have the default layout with spacing values (and means you might be able to remove the Flex components with the explicit margin values).

dgcohen:
Also, I'm seeing a Z-index issue with the clear button, not sure if this is a DS issue:

Screenshot 2024-10-16 at 10 20 41 AM

EdwinGuzman (commented):
Good updates and added more comments. I like the two new reusable components, perhaps they can go into a directory that explicitly says they're reusable or shared? It's not necessary though.

The name for `SettingsFormLabel` makes it seem like it's a label for an input and can be misleading since form `<label>`s are associated with HTML input elements. This is a different type of label and maybe it's heading or just `SettingsLabel`?

src/components/MyAccount/NewSettings/EmailForm.tsx
7emansell: Where can I find the suggested copy? @EdwinGuzman

7emansell:
@EdwinGuzman Where is the suggested copy for the failure banner? (sorry if I've asked this somewhere else on this PR, i thought i did but now can't find it)

src/components/MyAccount/NewSettings/EmailForm.test.tsx
7emansell: No particular reason I guess, would it be better to be all regex or all strings?

EdwinGuzman:
> can I make a bug fix on the DS

For these z-index issue, yes.

src/components/MyAccount/NewSettings/EmailForm.test.tsx
EdwinGuzman: I _think_ strings are suppose to be a faster lookup, but it might be negligible. The bottleneck, if any, might be the `getByRole` function, but either regex or string is ok.

src/components/MyAccount/NewSettings/SaveCancelButtons.tsx
```diff
@@ -0,0 +1,43 @@
...
+        buttonType="secondary"
+        onClick={onCancel}
+      >
+        Cancel
+      </Button>
+      <Button
+        id="save-button"
+        isDisabled={isDisabled}
+        buttonType="primary"
+        minWidth={"123px"}
```
EdwinGuzman, line 34: This seems oddly specific. Was the width too small before?

src/components/MyAccount/NewSettings/NewAccountSettingsTab.tsx
EdwinGuzman, line 1: This might need some tests. Perhaps not `SettingsLabel` because it's small and straightforward enough, but it wouldn't hurt to have a simple test that verifies that the props were called when the buttons are clicked.

In this case, it'd be helpful to test the success and error messages but it depends heavily on the set functions being called within the components. So this is starting to become integration tests.

You can leave these comments for later but add tests eventually.

src/components/MyAccount/NewSettings/EmailForm.tsx
```diff
@@ -0,0 +1,243 @@
...
+                <Form id="email-form" key={index} gap={"grid.xxs"}>
+                  <Flex mb="s" width="fill">
+                    <TextInput
+                      sx={{
+                        width: {
+                          base: "87%",
```
EdwinGuzman, line 140: Why this value? Did it just look right?

EdwinGuzman (approved):
This looks good and the comments are not blockers. You should write more tests but since this is going into the feature branch and won't be released, you'll have time to write more involved tests later.

The mobile input's (with the trash can) width is fluid on mobile. I don't think it's a big problem but it feels like something that will be noticed in vqa:
Screenshot 2024-10-17 at 3 23 51 PM
Screenshot 2024-10-17 at 3 23 57 PM

src/components/MyAccount/NewSettings/SaveCancelButtons.tsx
7emansell: Yes i think it shouldn't stack

src/components/MyAccount/NewSettings/EmailForm.tsx
EdwinGuzman: I think it's noticeable enough that it'll be picked up. A static value sounds good but don't think it'll scale either. Let's keep this as is for now.

src/components/MyAccount/NewSettings/SaveCancelButtons.tsx
EdwinGuzman: I think you should be able to pass `buttonWidth="full"` in `ButtonGroup` for it to expand and not stack. Try that.

src/components/MyAccount/NewSettings/EmailForm.tsx
charmingduchess: you can reference the original [My Account Figma ](https://www.figma.com/design/k2E6Fc3bFV2GebWwCkmJQD/My-Account?node-id=2834-49530&node-type=canvas&t=XQqkUwOe2U2TltEc-0) file for guidelines on widths and button location. I think it's fair to assume that those should stay the same for button placement and widths.

charmingduchess (approved):

PR #369, opened 2024-10-18
SCC-3759 - Delivery Locations fetching

dgcohen:
## Ticket:

- JIRA ticket [SCC-3759](https://newyorkpubliclibrary.atlassian.net/browse/SCC-3759)

- Adds locations server file with delivery locations fetching function
- Adds locations utils derived from DFE (and heavily refactored) that formats location structure from api response to the new DeliveryLocation type that is more easily parsed by the Hold page component.
- I was debating on whether to create a model for Locations but i now think it's overkill since the structure of DeliveryLocation can be easily formatted as json and won't change when it reaches the client side. Let me know if you disagree.
- I got rid of the closed locations environment variables since these are now hard to change and require a change to the terraform repo. This also allows us to use an array instead of a semicolon delineated string.
- We are currently using the first part of the short name in the closed locations list in DFE, maybe we should talk about using the slug instead so that we can make a typed list of allowed location values.

## How has this been tested?

- Added unit tests for location utils.

NA

### Checklist:

[SCC-3759]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-3759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/config/constants.ts
```diff
@@ -119,3 +120,18 @@ export const SEARCH_FORM_OPTIONS = {
     placeholder: `${example} Ornithology or Greek Architecture`,
   },
 }
+
+export const NYPL_LOCATIONS = {
```
charmingduchess, line 124: Right now in[ NYPL core,](https://github.com/NYPL/nypl-core/blob/master/vocabularies/csv/locations.csv) both schomburg and schwarzman labels already match this pattern (ie they start `{shortName}-`). For LPA locations, they are all prefixed with "Performing Arts Research Collections", which could be updated to "Library for the Performing Arts". I'm wondering if it's worth re-thinking this whole business with the mapping and the slugs. One alternate solution might be:
We could update the LPA locations with this short name in the pref label, and use the a substring of the preflabel to map to the address. 
We could even return the address in the discovery API response, though that would require a little more time to update the API responses. Updating NYPL core with a new LPA label would be pretty trivial, and would not delay this work at all. I think this would cut down a lot of the complexity of this area of the code.

@nonword do you have thoughts on this?

nonword: The primary purpose of `location.prefLabel` in NYPL-Core is display of location labels in RC (via the discovery-api), so the question becomes: Do we want to represent LPA locations as "Performing Arts Research Collections" in some RC contexts (Search Results, Item Table) but call it "Library for the Performing Arts" on the Hold Request page? It does seem to me these should be unified, but I'm not sure which should be updated. It looks like DFE has used the "Library for the Performing Arts" delivery label for the past 8 years, so perhaps that's the old one and we can retire it. That would allow these "shortName" values to be derived from `location.prefLabel` in the delivery locations response. The address may have to continue to be hardcoded for now since the best source for that is the location-service, but the addresses there slightly disagree with the addresses shown here and we probably don't need that level of data-driven brittleness for something that never changes

charmingduchess: Sean says to go ahead with it as is, and we'll return to this convo after conferring comms. 
@dgcohen I'm wondering if the code can be updated to only update LPA prefLabels? It could act as inline documentation of the special case of LPA prefLabels requiring transformation when SASB and schomburg do not.

src/types/locationTypes.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export interface DiscoveryLocationItem {
+  eddRequestable?: boolean
+  deliveryLocation?: DiscoveryLocationElement[]
+}
+
+/* eslint-disable @typescript-eslint/naming-convention */
+
+export interface DiscoveryLocationElement {
+  "@id"?: string
```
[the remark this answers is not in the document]
dgcohen: I believe that's how it's formatted in the discovery response

src/config/constants.ts
dgcohen: Sounds good, i added a ternary that makes non-LPA locations default to the prefLabel. Let me know if that's not the desired behavior.

EdwinGuzman (commented):
This looks ok to me but I will defer to Vera and Paul for their review on this.

PR #392, opened 2024-11-06
SCC-4337: Password, home library, and notification for My Account 2.0

7emansell:
## Ticket:

Resolves JIRA ticket [SCC-4337](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4337)

- Adds home library, notification preference, and password forms
   - Extracts/renames the form types as `SettingsSelectForm` (home library, notification preference) and `SettingsInputForm` (phone, email)
- Adds tests for each
- Updates tab state to prevent multiple forms from being edited simultaneously

## How has this been tested?

Locally

### Checklist:


src/components/MyAccount/NewSettings/PasswordForm.tsx
```diff
@@ -0,0 +1,262 @@
...
+    }
+  }
+
+  return (
+    <>
+      {isLoading ? (
+        <SkeletonLoader contentSize={2} showImage={false} headingSize={0} />
+      ) : isEditing ? (
+        <>
+          <Flex flexDir={{ base: "column", lg: "row" }}>
```
7emansell, line 111: Noting that L111-190 is extremely repetitive but also resistant to being pulled out as a component... the `onChange`s are sensitive I guess

src/components/MyAccount/NewSettings/SettingsSelectForm.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+  const sortedPickupLocations = [
+    patronData.homeLibrary,
+    ...pickupLocations.filter(
+      (loc) => loc.code.trim() !== patronData.homeLibrary.code.trim()
+    ),
+  ]
+
+  const options =
+    type === "notification" ? notificationPreferenceMap : sortedPickupLocations
```
charmingduchess, line 57: can this be added to the form utils body?
also note that we don't want to allow patrons to change their preference to none, so this option should be removed from the dropdown if it's not their currently selected option.

src/components/MyAccount/NewSettings/SettingsSelectForm.tsx
```diff
@@ -0,0 +1,199 @@
...
+    body: (code) =>
+      type === "notification"
+        ? {
+            fixedFields: { "268": { label: "Notice Preference", value: code } },
+          }
+        : { homeLibraryCode: `${code}` },
+  }
+
+  const [selection, setSelection] = useState(
+    type === "notification"
```
charmingduchess, line 75: these can also be on the formutils as `inititalState`

charmingduchess (changes requested):
Looking good!

On a general note, I'm interested in how the different `set` method calls are coupled. In some cases, you've combined them into a method (like `cancelEditing`). I like that pattern. I notice that `setIsLoading(false)` and `setIsOtherEditing(false)` are basically always called together after submission. I think those two could benefit from being combined into a method like  `teardownSubmission`, passed down from the account tab. Potentially other groups could as well. 
It could be defined per component, something like 
```
<InputForm
  cancelEditing={() => {setIsEditing(false)
                                      setIsLoading(false)
}}
```
or maybe predefined for methods that would be the same across components.
My thinking is, if we have the parent form state, the children don't need to worry about whether editing is true or loading is false. Instead, they could rely on an API like `cancelEditing`, `teardownSubmission`, etc, that way, if there are changes to what actual `set` methods are being called within those, it can change in one place.

This is my impression from going over the code. I don't think it absolutely has to happen, but I'm also thinking doing this work might help to crystallize what setStates can be inferred from others, and potentially remove unnecessary state management.

src/components/MyAccount/NewSettings/SettingsSelectForm.tsx
charmingduchess: also note that we don't want to allow patrons to change their preference to none, so this option should be removed from the dropdown if it's not their currently selected option.

7emansell: Had to add none as a possibility to the Patron type to do this, not sure why that wasn't already the case, but lmk if that's intentional.

charmingduchess: Hmm I imagine that it has something to do with not letting patrons sign up without an email or phone number. Probably ok but let's make sure to flag it for QA.

src/components/MyAccount/NewSettings/NewAccountSettingsTab.tsx
```diff
@@ -1,44 +1,95 @@
...
+import PasswordForm from "./PasswordForm"
 
 const NewAccountSettingsTab = () => {
   const {
-    updatedAccountData: { patron },
+    updatedAccountData: { patron, pickupLocations },
   } = useContext(PatronDataContext)
-  const [isSuccess, setIsSuccess] = useState(false)
-  const [isFailure, setIsFailure] = useState(false)
+  const [status, setStatus] = useState<[string, string?]>(["none"])
```
charmingduchess, line 12: Alrighty... I did not consider that having the other error message was going to add another dimension to the success/failure state. Now that I see this added dimension, I see there are in fact two states needed to be kept track of (failure/success, and whether the failure is a current password failure). Could you try and rethink another approach to this banner situation? Don't worry about my specific directive before to collapse status into one state, unless it can end up being truly one state.

src/components/MyAccount/NewSettings/SettingsSelectForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+    initialState:
+      type === "notification"
+        ? notificationPreferenceMap.find(
+            (pref) => pref.code === patronData.notificationPreference
+          )?.name
+        : patronData.homeLibrary.name,
+    options:
+      type === "notification"
+        ? notificationPreferenceMap
+        : sortedPickupLocations,
```
EdwinGuzman, line 69: The `sortedPickupLocations` array is not expensive to create, but one concern is having each separate "library" or "notification" select form know about (by creating) both type of data. You can probably pass the data you need from the parent instead of doing both operations here.

It's not a big concern because they're not expensive but something to keep in mind.

EdwinGuzman (commented):
This looks good overall. The form layout jumps a bit when you click on any of the edit buttons. It's probably because the height is not the same in the view type vs the edit type. The select dropdown is bigger and adds some height.


https://github.com/user-attachments/assets/604cdbe4-1b0a-4d6f-a028-a089be5f068e

EdwinGuzman:
I also think the select should span the entire width on mobile but get Apoorva's take on it:
Screenshot 2024-11-13 at 7 02 20 PM

src/components/MyAccount/NewSettings/HomeLibraryForm.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+      },
+      status: 200,
+    } as Response)
+  })
+
+  it("renders correctly with initial location", () => {
+    render(component)
+
+    expect(
+      screen.getByText(processedPatron.homeLibrary.name)
```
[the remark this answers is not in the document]
7emansell: I'm confused, I think this IS testing that a specific value appears (it'll fail if the library that appears is not the user's currently selected home library)

src/components/MyAccount/NewSettings/NewAccountSettingsTab.tsx
```diff
@@ -1,44 +1,104 @@
...
-                ? "Your changes were not saved."
-                : "Your changes were saved."}
-            </div>
-          }
-          type={isFailure ? "warning" : "positive"}
-          sx={{ marginTop: "m" }}
-        />
+      {status !== "" && (
+        <div ref={bannerRef} tabIndex={-1}>
+          <Banner
```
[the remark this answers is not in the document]
7emansell: When is the wrapper div not rendered around the banner?

src/components/MyAccount/NewSettings/NotificationForm.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+        pickupLocations={pickupLocations}
+        type="notification"
+      />
+    </PatronDataProvider>
+  )
+
+  it("renders correctly with initial preference", () => {
+    render(component)
+
+    expect(screen.getByText(processedPatronPref)).toBeInTheDocument()
```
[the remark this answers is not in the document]
7emansell: `processedPatronPref` is what the patron's data is set as. This is testing whether the component displays that as the initial value for the form, before the patron begins editing. Am I missing some other initial state that I should be checking?

src/components/MyAccount/NewSettings/NewAccountSettingsTab.tsx
```diff
@@ -1,44 +1,104 @@
...
+                  ) : (
+                    <Text marginBottom={0} color={"ui.black !important"}>
+                      Your changes were not saved.
+                    </Text>
+                  )
+                ) : (
+                  <Text marginBottom={0} color={"ui.black !important"}>
+                    Your changes were saved.
+                  </Text>
+                )}
```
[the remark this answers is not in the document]
7emansell: I agree that the code is not easy to read but I'm not sure how assigning it to a variable would help here

src/components/MyAccount/NewSettings/NewAccountSettingsTab.tsx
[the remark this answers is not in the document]
7emansell: Okay I'll flag it in the AC. It definitely seems wrong that the div should always render even when there's no banner visible inside to be screen-read, and the focus does (or should) always go to the banner when it appears

7emansell:
> This looks good overall. The form layout jumps a bit when you click on any of the edit buttons. It's probably because the height is not the same in the view type vs the edit type. The select dropdown is bigger and adds some height.

I think this is now corrected on desktop but not on mobile... I need some general guidance from Apoorva about responsive behavior on the mobile/tablet breakpoints so will make sure to ask during VQA

EdwinGuzman (approved):

PR #393, opened 2024-11-08
SCC-3760 - EDD page UI

dgcohen:
## Ticket:

- JIRA ticket [SCC-3760](https://newyorkpubliclibrary.atlassian.net/browse/SCC-3760)

- Adds the EDD request page and associated components (EDD Form)
- Adds validation logic to focus on first invalid field on submission
- Extracts hold request banner into its own component and adds logic for different statuses in EDD form (still haven't implemented statuses in Hold page since there are only 2 states)
- Adds basic tests (more detailed tests forthcoming)

## How has this been tested?

- Added basic unit tests

NA

### Checklist:

[SCC-3760]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-3760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


charmingduchess (changes requested):
Looks good!
I think it would be a good idea to rename `holdUtils` and `holdTypes` to `holdPageUtils` and `holdPageTypes`

src/reducers/eddFormReducer.ts
```diff
@@ -0,0 +1,17 @@
+import type { EDDRequestParams, EDDFormAction } from "../types/holdTypes"
+
+export const eddFormReducer = (
```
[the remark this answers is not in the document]
charmingduchess: @dgcohen is this possible? this seems redundant to the other reducer.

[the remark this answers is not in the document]
charmingduchess: @dgcohen what do you think about this?

src/components/HoldPages/EDDRequestForm.tsx
```diff
@@ -0,0 +1,282 @@
...
+  const [invalidFields, setInvalidFields] = useState(initialEDDInvalidFields)
+
+  // Create refs for fields that require validation to focus on the first invalid field on submit
+  const [validatedInputRefs] = useState(
+    invalidFields.reduce((acc, field) => {
+      return { ...acc, [field.key]: createRef() }
+    }, {})
+  )
+
+  // Separate validation from input change to validate on tabbing (blur) and to prevent validation on every keystroke on fields other than email
```
EdwinGuzman, line 62: 👍

src/components/HoldPages/EDDRequestForm.tsx
```diff
@@ -0,0 +1,282 @@
...
+  return (
+    <Form
+      id="edd-request-form"
+      data-testid="edd-request-form"
+      // We are using a post request on hold requests when JS is disabled
+      method="post"
+      action={`${BASE_URL}/api/hold/request/${holdId}/edd`}
+      onSubmit={validateAndSubmit}
+      mb="l"
+      maxWidth={{ lg: "75%" }}
```
EdwinGuzman, line 115: Is this from the figma design?

src/components/HoldPages/HoldRequestBanner.tsx
```diff
@@ -45,37 +47,42 @@ const HoldRequestBanner = ({
...
-            for assistance, or{" "}
-            <RCLink href="/search">start a new search.</RCLink>
+            {errorMessage}
+            {pageStatus !== "invalid" ? (
+              <>
+                {" Please try again, "}
+                <Button
+                  id="hold-contact"
+                  onClick={() =>
+                    onContact({
```
EdwinGuzman, line 57: 👍

src/utils/holdPageUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+  prevInvalidFields: EDDFormValidatedField[]
+): EDDFormValidatedField[] => {
+  return prevInvalidFields.map((field) => {
+    if (field.key === fieldName) {
+      switch (field.key) {
+        // Validate email field
+        case "email":
+          return {
+            key: "email",
+            isInvalid: !fieldValue.length || !EMAIL_REGEX.test(fieldValue),
```
EdwinGuzman, line 62: I forgot already, but is this similar or the same validation as the My Account email test?

EdwinGuzman (commented):
Looks great but have a few questions.

One big question about how the form works, when it fails I get the error message and that's fine, but all my data is gone. I would expect it to stay on the page for me to edit and make any updates but the form clears. We can discuss this as a group but I think the existing form does this, and in general, I think we should maintain the user data in forms when they fail.

charmingduchess:
> One big question about how the form works, when it fails I get the error message and that's fine, but all my data is gone. I would expect it to stay on the page for me to edit and make any updates but the form clears. We can discuss this as a group but I think the existing form does this, and in general, I think we should maintain the user data in forms when they fail.
I feel like I remember this persisting user data being implemented when I first started working here, and we probably should include it again.

src/components/HoldPages/EDDRequestForm.tsx
```diff
@@ -0,0 +1,282 @@
...
+    setInvalidFields((prevInvalidFields) =>
+      getUpdatedInvalidFields(target.name, target.value, prevInvalidFields)
+    )
+  }
+
+  const handleInputChange = (e: SyntheticEvent) => {
+    e.preventDefault()
+    const target = e.target as HTMLInputElement
+
+    dispatch({
```
charmingduchess, line 76: I think this reducer is overkill for the single kind of input we're handling here.
I'd love to see it removed and instead just using
`const [eddFormState, setEddFormstate] = useState(initialFormState)`
and then calling `setEddFormState` instead of a dispatch to the reducer.

src/utils/holdPageUtils.ts
charmingduchess: I don't know that Emma's work is referencing this constant yet, but this constant is based on Emma's validation regex.

__test__/pages/hold/eddRequestPage.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+        res: mockRes,
+        req: mockReq,
+      })
+      expect(mockRes.setHeader.mock.calls[0]).toStrictEqual([
+        "Set-Cookie",
+        "nyplAccountRedirects=1; Max-Age=10; path=/; domain=.nypl.org;",
+      ])
+    })
+  })
+  describe("aeon redirect handling", () => {
```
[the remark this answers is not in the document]
dgcohen: http://local.nypl.org:8080/research/research-catalog/hold/request/b21086080-i34439033/edd

src/components/HoldPages/EDDRequestForm.tsx
dgcohen: Good call! replace reducer with useState

dgcohen:
> Looks great but have a few questions.
> 
> One big question about how the form works, when it fails I get the error message and that's fine, but all my data is gone. I would expect it to stay on the page for me to edit and make any updates but the form clears. We can discuss this as a group but I think the existing form does this, and in general, I think we should maintain the user data in forms when they fail.

Fixed this by moving the state up to the page component level. I also removed the reducer and replaced with useState, but that was more to simplify things and didn't contribute to the fix.

src/config/constants.ts
```diff
@@ -139,3 +142,55 @@ export const NYPL_LOCATIONS = {
 
 export const HOLD_PAGE_HEADING = "Request for on-site use"
 export const EDD_PAGE_HEADING = "Request scan"
+
+export const EDD_FORM_FIELD_COPY = {
```
EdwinGuzman, line 146: Good idea.

EdwinGuzman (approved):
Let's keeping move forward!

PR #397, opened 2024-11-20
SCC-4236: Username for My Account 2.0 

7emansell:
## Ticket:

Resolves JIRA ticket [SCC-4236](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4236).

- Adds `UsernameForm` to profile header
- Adds `api/account/username` route to validate new patron usernames
- Removes "discovery" from `src/config` (it's a duplicate of "platform")
- Updates `nyplApiClient` to take an api version (defaults to `v0.1`)
- Pulls out the settings status banner into a reusable `StatusBanner`

### Checklist:

### DIFF
--- a/pages/api/account/helpers.ts
+++ b/pages/api/account/helpers.ts
@@ -1,5 +1,6 @@
 import sierraClient from "../../../src/server/sierraClient"
 import type { HTTPResponse } from "../../../src/types/appTypes"
+import nyplApiClient from "../../../src/server/nyplApiClient"
 
 /**
  * PUT request to Sierra to update patron PIN, first validating with previous PIN.
@@ -23,6 +24,39 @@ export async function updatePin(
     return {
       status: error.response.status,
       message: error.response.data.message || error.response.data.description,
+    }
+  }
+}
+
+/**
+ * PUT request to Sierra to update patron username, first validating that it's available.
+ * Returns status and message about request.
+ */
+export async function updateUsername(
+  patronId: string,
+  newUsername: string
+): Promise<HTTPResponse> {
+  try {
+    const platformClient = await nyplApiClient({
+      version: "v0.3",
+    })
+    const response = await platformClient.post("/validations/username", {
+      username: newUsername,
+    })
+    if (response?.type === "available-username") {
+      const client = await sierraClient()
+      await client.put(`patrons/${patronId}`, {
+        varFields: [{ fieldTag: "u", content: newUsername }],
+      })
+      return { status: 200, message: `Username updated to ${newUsername}` }
+    } else if (response?.type === "unavailable-username") {
+      // Not an error for the username to be taken, FE should read the message
+      return { status: 200, message: "Username taken" }
+    } else return { status: 500, message: "Username update failed" }
+  } catch (error) {
+    return {
+      status: error.status,
+      message: error.message || error.response.data.description,
     }
   }
 }
--- a/pages/api/account/helpers.ts
+++ b/pages/api/account/helpers.ts
@@ -37,26 +37,39 @@ export async function updateUsername(
   newUsername: string
 ): Promise<HTTPResponse> {
   try {
-    const platformClient = await nyplApiClient({
-      version: "v0.3",
-    })
-    const response = await platformClient.post("/validations/username", {
-      username: newUsername,
-    })
-    if (response?.type === "available-username") {
+    // If the new username is an empty string, skips validation and directly updates in Sierra.
+    const client = await sierraClient()
+    if (newUsername === "") {
       const client = await sierraClient()
       await client.put(`patrons/${patronId}`, {
         varFields: [{ fieldTag: "u", content: newUsername }],
       })
-      return { status: 200, message: `Username updated to ${newUsername}` }
-    } else if (response?.type === "unavailable-username") {
-      // Not an error for the username to be taken, FE should read the message
-      return { status: 200, message: "Username taken" }
-    } else return { status: 500, message: "Username update failed" }
+      return { status: 200, message: "Username removed successfully" }
+    } else {
+      const platformClient = await nyplApiClient({ version: "v0.3" })
+      const response = await platformClient.post("/validations/username", {
+        username: newUsername,
+      })
+
+      if (response?.type === "available-username") {
+        await client.put(`patrons/${patronId}`, {
+          varFields: [{ fieldTag: "u", content: newUsername }],
+        })
+        return { status: 200, message: `Username updated to ${newUsername}` }
+      } else if (response?.type === "unavailable-username") {
+        // Username taken but not an error, returns a message.
+        return { status: 200, message: "Username taken" }
+      } else {
+        return { status: 500, message: "Username update failed" }
+      }
+    }
   } catch (error) {
     return {
-      status: error.status,
-      message: error.message || error.response.data.description,
+      status: error?.status || 500,
+      message:
+        error?.message ||
+        error.response?.data?.description ||
+        "An error occurred",
     }
   }
 }
--- a/src/components/MyAccount/NewSettings/AddButton.tsx
+++ b/src/components/MyAccount/NewSettings/AddButton.tsx
@@ -1,7 +1,7 @@
 import { Button } from "@nypl/design-system-react-components"
 
 type AddButtonProps = {
-  inputType: string
+  inputType?: string
   label: string
   onClick: () => void
 }
@@ -9,7 +9,7 @@ type AddButtonProps = {
 const AddButton = ({ inputType, label, onClick }: AddButtonProps) => {
   return (
     <Button
-      id={`add-${inputType}-button`}
+      id={inputType ? `add-${inputType}-button` : "add-button"}
       buttonType="text"
       onClick={onClick}
       size="large"
--- a/src/components/MyAccount/NewSettings/NewAccountSettingsTab.tsx
+++ b/src/components/MyAccount/NewSettings/NewAccountSettingsTab.tsx
@@ -1,9 +1,10 @@
-import { Flex, Banner, Link, Text } from "@nypl/design-system-react-components"
+import { Flex } from "@nypl/design-system-react-components"
 import { useContext, useEffect, useRef, useState } from "react"
 import { PatronDataContext } from "../../../context/PatronDataContext"
 import SettingsInputForm from "./SettingsInputForm"
 import SettingsSelectForm from "./SettingsSelectForm"
 import PasswordForm from "./PasswordForm"
+import { StatusBanner } from "./StatusBanner"
 
 type StatusType = "" | "failure" | "success"
 
@@ -33,47 +34,11 @@ const NewAccountSettingsTab = () => {
     }
   }, [status])
 
-  const bannerContent = (
-    <div style={{ alignItems: "center" }}>
-      {status === "failure" ? (
-        statusMessage !== "" ? (
-          <Text marginBottom={0} color={"ui.black !important"}>
-            {statusMessage} Please try again or{" "}
-            <Link
-              sx={{
-                color: "ui.link.primary !important",
-                textDecorationColor: "ui.link.primary !important",
-                textDecoration: "underline",
-              }}
-              href="https://www.nypl.org/get-help/contact-us"
-            >
-              contact us
-            </Link>{" "}
-            for assistance.
-          </Text>
-        ) : (
-          <Text marginBottom={0} color={"ui.black !important"}>
-            Your changes were not saved.
-          </Text>
-        )
-      ) : (
-        <Text marginBottom={0} color={"ui.black !important"}>
-          Your changes were saved.
-        </Text>
-      )}
-    </div>
-  )
-
   return (
     <>
       {status !== "" && (
         <div ref={bannerRef} tabIndex={-1}>
-          <Banner
-            sx={{ marginTop: "m" }}
-            isDismissible
-            content={bannerContent}
-            type={status === "failure" ? "negative" : "positive"}
-          />
+          <StatusBanner status={status} statusMessage={statusMessage} />
         </div>
       )}
       <Flex flexDir="column" sx={{ marginTop: "xl", gap: "s" }}>
--- a/src/components/MyAccount/NewSettings/StatusBanner.tsx
+++ b/src/components/MyAccount/NewSettings/StatusBanner.tsx
@@ -1,0 +1,50 @@
+import { Banner, Text, Link } from "@nypl/design-system-react-components"
+
+export type StatusType = "" | "failure" | "usernameFailure" | "success"
+
+type StatusBannerProps = {
+  status: StatusType
+  statusMessage: string
+}
+
+export const StatusBanner = ({ status, statusMessage }: StatusBannerProps) => {
+  const bannerContent = (
+    <div style={{ alignItems: "center" }}>
+      {status === "failure" ? (
+        statusMessage !== "" ? (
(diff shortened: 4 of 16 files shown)


src/server/nyplApiClient/index.ts
```diff
@@ -22,13 +22,16 @@ export class NyplApiClientError extends Error {
   }
 }
 
-const nyplApiClient = async (options = { apiName: "platform" }) => {
-  const { apiName } = options
+const nyplApiClient = async ({
+  apiName = "platform",
+  version = "v0.1",
+} = {}) => {
```
7emansell, line 28: Should this be more specific ? this is my first time using this client so I'm not super familiar with its use across the repo

src/components/MyAccount/NewSettings/StatusBanner.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+type StatusBannerProps = {
+  status: StatusType
+  statusMessage: string
+}
+
+export const StatusBanner = ({ status, statusMessage }: StatusBannerProps) => {
+  const bannerContent = (
+    <div style={{ alignItems: "center" }}>
+      {status === "failure" ? (
```
charmingduchess, line 13: Can you pull out this conditional logic so it's not in nested ternaries, and illustrates the possibilities more clearly? Something along the lines of...
```
let statusComponent
if (status === 'success'){
  statusComponent = ...
} if (status === 'failure && statusMessage){
statusComponent = ...
else {
statusComponent = }
etc
```
potentially also with some descriptive variable names that point to when a `statusMessage` might be there or not

src/components/MyAccount/NewSettings/UsernameForm.tsx
```diff
@@ -0,0 +1,211 @@
...
+      width="100%"
+    >
+      <Flex gap="xs" marginRight="140px" paddingTop="xs" alignItems="center">
+        <Icon name="actionIdentity" size="large" />
+        <Text size="body1" sx={{ fontWeight: "500", marginBottom: 0 }}>
+          Username
+        </Text>
+      </Flex>
+
+      {isLoading ? (
```
charmingduchess, line 117: Similar request here. Can you replace this nested ternary with a conditional outside of the return statement so it's a easier to scan which components get rendered when?

src/components/MyAccount/ProfileHeader.tsx
```diff
@@ -10,18 +11,31 @@ import styles from "../../../styles/components/MyAccount.module.scss"
...
+
+  const usernameState = {
+    setUsernameStatus,
+    setUsernameStatusMessage,
+  }
 
   const profileData = (
     [
       { icon: "actionIdentityFilled", term: "Name", description: patron.name },
-      {
```
charmingduchess, line 20: Did you try passing in 
`<UsernameForm patron={patron} usernameState={usernameState} />` as the description instead of escaping this pattern?

src/server/nyplApiClient/index.ts
charmingduchess: What do you mean by specific? Also, can you explain what the `= {}` in the function signature does? 
FWIW this looks fine by me!

charmingduchess: @dgcohen maybe you can confirm if this looks kosher? the username validation endpoint requires v0.3, so i suggested emma make the version be a param. at the same time, i thought this was a good opportunity to trim away the duplicated discovery/platform urls in the config.

src/components/MyAccount/NewSettings/UsernameForm.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+      <UsernameForm
+        patron={processedPatron}
+        usernameState={mockUsernameState}
+      />
+    </PatronDataProvider>
+  )
+
+  const noUsernameComponent = (
+    <PatronDataProvider
+      testSpy={accountFetchSpy}
```
charmingduchess, line 44: Unless you want to validate something on this spy, you don't need to provide this value.

src/components/MyAccount/NewSettings/NewAccountSettingsTab.tsx
charmingduchess, line 1: I like this extraction 🌝

src/components/MyAccount/NewSettings/AddButton.tsx
```diff
@@ -1,15 +1,15 @@
...
+  inputType?: string
   label: string
   onClick: () => void
 }
 
 const AddButton = ({ inputType, label, onClick }: AddButtonProps) => {
   return (
     <Button
-      id={`add-${inputType}-button`}
+      id={inputType ? `add-${inputType}-button` : "add-button"}
```
charmingduchess, line 12: I think it's a good idea to have the id be descriptive. What do you think about reverting this change and adding an input type to the username AddButton?

charmingduchess (changes requested):

src/components/MyAccount/ProfileHeader.tsx
7emansell: So true

src/components/MyAccount/NewSettings/AddButton.tsx
7emansell: I'm hesitant to add another prop to UsernameForm that would only be used in this one place

src/components/MyAccount/NewSettings/StatusBanner.tsx
7emansell: Lmk if this is better– I do think, since the parent component is responsible for passing the `statusMessage`, and there's several potential parents and reasons for passing a specific message, any explanation or indication of what that message will be should also live on the parent component if that makes sense

src/components/MyAccount/NewSettings/UsernameForm.tsx
7emansell: Again lmk if this is better/ more readable

src/components/MyAccount/ProfileHeader.tsx
7emansell: Alignment is off now but working on that

src/server/nyplApiClient/index.ts
7emansell: I don't know actually how it could be more specific, just generally looking for confirmation that this makes sense from people who wrote/used this before

src/components/MyAccount/NewSettings/AddButton.tsx
7emansell: Maybe i should just pass id directly to AddButton instead of setting like this?

charmingduchess: Just leave it as is. This is a totally fine pattern.

src/components/MyAccount/NewSettings/UsernameForm.tsx
charmingduchess: Same note here - the extraction is good, but the nested ternary itself is hard to read. Can you replace it?

src/server/api/bib.ts
```diff
@@ -7,7 +7,6 @@ import {
 } from "../../utils/bibUtils"
 import nyplApiClient from "../nyplApiClient"
 import {
-  DISCOVERY_API_NAME,
```
EdwinGuzman, line 10: Can you remove this variable from `constants.ts`? It's no longer being used.

src/server/nyplApiClient/index.ts
EdwinGuzman: I think it's clear enough when you call the function what version is being used since you pass it as an argument. All tests also continue to pass so it shouldn't break anything.

One possible option to make this clearer (maybe?) is to make this a higher-order function. So it can be something like 
```
const baseNyplApiClient = ({
  version = "v0.1",
} = {}) => {
  return async ({
    apiName = "platform",
  } = {}) => {
      if (CACHE.clients[apiName]) {
        return CACHE.clients[apiName]
      }
    
      const baseUrl =
        appConfig.apiEndpoints[apiName][appEnvironment] + "/" + version
    //...
  })
})

export const nyplApiClientv01 = baseNyplApiClient()
export const nyplApiClientv03 = baseNyplApiClient({ version: "v0.3" })
```

And then call either function like they're being called now such as `const platformClient = await nyplApiClientv03()`. 

Caveat: the syntax might be off and I haven't tested this. If you like this approach, do it in another PR because of its use across the repo. But what do you think, @charmingduchess ?

src/components/MyAccount/ProfileHeader.tsx
```diff
@@ -53,19 +74,32 @@ const ProfileHeader = ({ patron }: { patron: Patron }) => {
...
-        border: "none",
-        h2: { border: "none", paddingTop: 0 },
-        marginBottom: "xxl",
-      }}
-    >
-      {profileData}
-    </List>
+    <>
+      {usernameStatus !== "" && (
+        <div
```
EdwinGuzman, line 79: I think we talked about this in a different PR, check with Clare if the banner message is read. In the [Dynamic Banners](https://nypl.github.io/nypl-design-system/reservoir/v3/?path=/docs/components-feedback-banner--docs#dynamic-banners) we say that this wrapper div should always be rendered and have an `aria-live` attribute. If Clare signs off on this, let me know so we can update the docs.

src/components/MyAccount/NewSettings/UsernameForm.tsx
```diff
@@ -0,0 +1,216 @@
...
+            display: "flex",
+            flexDirection: "column-reverse",
+            label: {
+              fontWeight: "400",
+              color: error ? "ui.error.primary" : "ui.black",
+            },
+          }}
+          value={tempInput || ""}
+          id="username-input"
+          labelText="Must be 5-15 characters and use only letters (a-z) and numbers (0-9)"
```
EdwinGuzman, line 119: The space between the input and label seems too small, right?

src/components/MyAccount/NewSettings/UsernameForm.tsx
```diff
@@ -0,0 +1,216 @@
...
+    <>
+      <Flex flexDir="row" alignItems="flex-start">
+        <TextInput
+          sx={{
+            width: { base: "87%", md: "300px" },
+            display: "flex",
+            flexDirection: "column-reverse",
+            label: {
+              fontWeight: "400",
+              color: error ? "ui.error.primary" : "ui.black",
```
EdwinGuzman, line 114: This should happen automatically when `isInvalid` is true. Or is this because the additional `!validateUsername` check below affects the text color?

EdwinGuzman (commented):
> NOTE: the deletion process is still up for debate. As it is right now, the user can submit an empty string which works the same as using the trash icon to officially delete

What's the debate? That there are two approaches for the same outcome?

This looks really good. There's the alignment issue in the description property when the username edits and I also left other comments.

src/components/MyAccount/NewSettings/StatusBanner.tsx
```diff
@@ -0,0 +1,60 @@
...
+)
+
+const specificFailureContent = (statusMessage: string) => {
+  return (
+    <Text marginBottom={0} color={"ui.black !important"}>
+      {statusMessage} Please try again or{" "}
+      <Link
+        sx={{
+          color: "ui.link.primary !important",
+          textDecorationColor: "ui.link.primary !important",
```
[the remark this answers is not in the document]
7emansell: They are blue in the designs and this does work but i'll change it

src/components/MyAccount/NewSettings/UsernameForm.tsx
7emansell: This is because the label and the invalid text are the same copy, and render in the same place (below the field). Since that's not how this component expects to work, I'm manually making the label text become the invalid text when needed

src/components/MyAccount/NewSettings/UsernameForm.tsx
7emansell: Yes, j added more styling

7emansell: It's like a mix of the label and helper text on this component I'm not sure how design arrived at this

src/components/MyAccount/NewSettings/StatusBanner.tsx
```diff
@@ -7,35 +7,45 @@ type StatusBannerProps = {
...
-          <Text marginBottom={0} color={"ui.black !important"}>
-            Your changes were not saved.
-          </Text>
-        )
-      ) : (
-        <Text marginBottom={0} color={"ui.black !important"}>
-          Your changes were saved.
-        </Text>
-      )}
+      {status === "failure"
```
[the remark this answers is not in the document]
7emansell: Okay ternary gone but I'd like to register my dissent... I think this makes it less readable both here and in the username form

src/server/nyplApiClient/index.ts
charmingduchess: I like the look of that!

src/components/MyAccount/NewSettings/UsernameForm.tsx
charmingduchess: Can you use `helperText` here instead of `labelText`? It looks like the helper text is automatically with the correct styling for this.

src/components/MyAccount/NewSettings/UsernameForm.tsx
charmingduchess: If there is a default that makes sense functionally (like the helper text) coming from reservoir that deviates from the designs in a small way, can you ask Apoorva if it's alright to use the default instead of overriding?

src/components/MyAccount/NewSettings/StatusBanner.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export type StatusType = "" | "failure" | "usernameFailure" | "success"
+
+type StatusBannerProps = {
+  status: StatusType
+  statusMessage: string
+}
+
+const successContent = (
+  <Text marginBottom={0} color="ui.black !important">
```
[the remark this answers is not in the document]
7emansell: Yes

pages/api/account/helpers.ts
```diff
@@ -27,6 +28,52 @@ export async function updatePin(
...
+export async function updateUsername(
+  patronId: string,
+  newUsername: string
+): Promise<HTTPResponse> {
+  try {
+    // If the new username is an empty string, skips validation and directly updates in Sierra.
+    const client = await sierraClient()
+    if (newUsername === "") {
+      const client = await sierraClient()
+      await client.put(`patrons/${patronId}`, {
```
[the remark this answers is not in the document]
7emansell: Yes just 200 if successful, no body

src/components/MyAccount/NewSettings/UsernameForm.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+    fireEvent.click(screen.getByRole("button", { name: /edit/i }))
+
+    const input = screen.getByLabelText(
+      "Must be 5-15 characters and use only letters (a-z) and numbers (0-9)"
+    )
+    fireEvent.change(input, { target: { value: "newUsername" } })
+
+    fireEvent.click(screen.getByRole("button", { name: /save changes/i }))
+
+    await waitFor(() => expect(fetch).toHaveBeenCalledTimes(2))
```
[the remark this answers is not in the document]
7emansell: No, it's hitting platform API for validation then sierra API to actually set the data

EdwinGuzman (approved):

PR #405, opened 2024-11-25
Scc 4334/replace owning institution

charmingduchess:
## Ticket:

- JIRA ticket [SCC-4334](https://newyorkpubliclibrary.atlassian.net/jira/software/c/projects/SCC/boards/126?assignee=61e836e95fcc37006878c15d&selectedIssue=SCC-4334)

- Add owning institution to the bib details for partner records only
- modify `BibDetails.prototype.buildStandardDetail` to accommodate owner and extent instance variables, instead of only being able to build details straight from the bib. This pattern could be extended to the subject heading and other details, but maybe for later work.

## How has this been tested?

- Verifying a partner record shows up with owning institution. Can test with `pb2608686`
- Verify NYPL record does not show up with owning institution

-

### Checklist:

[SCC-4334]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/models/BibDetails.ts
```diff
@@ -43,6 +45,17 @@ export default class BibDetails {
     this.bottomDetails = this.buildBottomDetails()
   }
 
+  buildOwner(): string[] {
```
dgcohen, line 48: Just curious, why does this have to be an array if it always returns one element?

src/models/BibDetails.ts
```diff
@@ -124,17 +140,10 @@ export default class BibDetails {
           detail = this.buildInternalLinkedDetail(fieldMapping)
         else if (fieldMapping.field === "subjectLiteral")
           detail = this.subjectHeadings
-        else if (fieldMapping.field === "extent") detail = this.extent
-        else if (fieldMapping.field === "owner")
-          detail = this.bib.owner && {
```
dgcohen, line 129: Where do the owner and extent attributes get called now that this has been removed?

src/models/BibDetails.ts
charmingduchess: I think I was trying to keep the details as standard as possible. The subject heading urls are nested arrays, and the intention was to have all the values for the bib details be string arrays.

charmingduchess:
@dgcohen I added 2 more commits to try and use the same pattern more, as well as renaming a method.

src/models/BibDetails.ts
charmingduchess: Let me know if you want me to explain! It's a little tricky but I'm hoping this streamlines things a little.

EdwinGuzman (commented):
I can't say I understand the business req/logic but the code looks right. Why did you have to change the return type from an object to an array of data? It seems like a general update.

__test__/pages/search/searchResults.test.tsx
```diff
@@ -32,11 +32,10 @@ describe("Search Results page", () => {
       })
       await userEvent.click(field)
       await userEvent.click(screen.getByText("Apply filters"))
-      // This was the only way to get this test to pass. waitFor was not in fact waiting, even with same timeout.
-      setTimeout(() => {
```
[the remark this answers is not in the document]
charmingduchess: not sure. it was not passing, so I changed it to waitFor, and the test started passing.

charmingduchess:
@EdwinGuzman The methods that I updated had been returning a label and a value. I wanted to DRY out the logic that was adding the label, especially because the labels were hardcoded in some cases. I updated the property methods (ie extent, subjectLiteral, and other values that are transformed) to only return the value, and let `buildStandardDetail` add the label.

EdwinGuzman (approved):

dgcohen (approved):

PR #409, opened 2024-12-03
SCC-4384: My Account 2.0 VQA/a11y round 1

7emansell:
## Ticket:

- JIRA ticket [SCC-4384](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4384)

- Addresses comments in [My Account 2.0](https://docs.google.com/document/d/1Rwi0eSEVBEfAMK1ypQ0NkGNd3FtI0ngRAaD-ONPIEvo/edit?tab=t.0) VQA and accessibility review
    - Mainly spacing (VQA) and focus behavior (a11y)

### Question/request for help: 
On mobile when you click edit for Home library/ Notification preference/Password, the fields jump up slightly like they're losing padding/margin or height is just not the same between the two states. I would love a fresh pair of eyes on this since I'm having trouble tracking down why this is happening

- Adds focus management to all form fields

### Checklist:

[SCC-4364]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SCC-4384]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/MyAccount/Settings/SettingsInputForm.tsx
+++ b/src/components/MyAccount/Settings/SettingsInputForm.tsx
@@ -189,7 +189,7 @@ const SettingsInputForm = ({
         <SettingsLabel icon={formUtils.icon} text={formUtils.inputLabel} />
         {isLoading ? (
           <SkeletonLoader
-            sx={{ "> div": { marginTop: "-xs" } }}
+            sx={{ "> div": { marginTop: "-s" } }}
             contentSize={2}
             showImage={false}
             headingSize={0}
@@ -240,7 +240,7 @@ const SettingsInputForm = ({
                 </Flex>
               </Form>
             ))}
-            <Flex flexDir="row">
+            <Flex flexDir="row" width="fill">
               <AddButton
                 inputType={inputType}
                 label={formUtils.addButtonLabel}
--- a/src/components/MyAccount/Settings/SettingsInputForm.tsx
+++ b/src/components/MyAccount/Settings/SettingsInputForm.tsx
@@ -44,6 +44,7 @@ const SettingsInputForm = ({
   const [tempInputs, setTempInputs] = useState([...inputs])
 
   const inputRefs = useRef<Array<TextInputRefType | null>>([])
+  const editingRef = useRef<HTMLButtonElement | null>()
 
   const focusLastInput = () => {
     const lastIndex = tempInputs.length - 1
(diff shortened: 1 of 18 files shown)


src/components/MyAccount/Settings/SettingsInputForm.tsx
```diff
@@ -42,9 +43,26 @@ const SettingsInputForm = ({
...
+  const editingRef = useRef<HTMLButtonElement | null>()
+
+  const focusLastInput = () => {
+    const lastIndex = tempInputs.length - 1
+    if (inputRefs.current[lastIndex]) {
+      inputRefs.current[lastIndex]?.focus()
+    }
+  }
+
+  useEffect(() => {
```
charmingduchess, line 56: With three text inputs open for the phone settings, this focus is triggering on every keystroke input. ie Typing a single character into input 2 jumps the focus to input 3. I think you want an empty array to only run on the first render. I think this will make typescript unhappy but that's ok.

src/components/MyAccount/Settings/SettingsInputForm.tsx
```diff
@@ -183,10 +208,15 @@ const SettingsInputForm = ({
                     sx={{
                       width: { base: "87%", md: "300px" },
                     }}
+                    ref={(el) => (inputRefs.current[index] = el)}
```
charmingduchess, line 211: Not sure if you came across [this](https://react.dev/learn/manipulating-the-dom-with-refs), but your solution is react.dev approved! Looking at their solution, though, I think it might be important to remember to delete refs from the array once the input is removed from the dom.

charmingduchess: I tried adding 5 inputs, then deleting one from the middle, and the count on the array of input refs actually went up to 6. It doesn't seem to cause any problems, but I think to avoid bugs down the line, the inputRefs array should match the inputs that are actually present on the dom.

charmingduchess: Actually just encountered a bug! after opting to stay logged in from the timed logout modal, the focus is lost and the last input is null, which leads to a loss of focus. I think  this can be avoided by removing refs as the inputs are removed.

charmingduchess (changes requested):

src/components/MyAccount/Settings/SettingsInputForm.tsx
7emansell: updated it to `tempInputs.length` which I think is right because then it updates only when a new input is added/removed

PR #412, opened 2024-12-05
SCC-4384: My Account 2.0 VQA/a11y round 1

7emansell:
## Ticket:

- JIRA ticket [SCC-4384](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4384)

- Addresses comments in [My Account 2.0](https://docs.google.com/document/d/1Rwi0eSEVBEfAMK1ypQ0NkGNd3FtI0ngRAaD-ONPIEvo/edit?tab=t.0) VQA and accessibility review
    - Mainly spacing (VQA) and focus behavior (a11y)

### Question/request for help: 
On mobile when you click edit for Home library/ Notification preference/Password, the fields jump up slightly like they're losing padding/margin or height is just not the same between the two states. I would love a fresh pair of eyes on this since I'm having trouble tracking down why this is happening

- Adds focus management to all form fields

### Checklist:

[SCC-4384]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/MyAccount/Settings/PasswordForm.tsx
+++ b/src/components/MyAccount/Settings/PasswordForm.tsx
@@ -1,5 +1,6 @@
-import { useContext, useState } from "react"
+import { forwardRef, useContext, useRef, useState } from "react"
 import { PatronDataContext } from "../../../context/PatronDataContext"
+import type { TextInputRefType } from "@nypl/design-system-react-components"
 import {
   Banner,
   Flex,
@@ -30,37 +31,37 @@ export const passwordFormMessages = {
   INVALID: "Invalid new pin/password.",
 }
 
-const PasswordFormField = ({
-  label,
-  handler,
-  name,
-  isInvalid,
-}: PasswordFormFieldProps) => {
-  return (
-    <Flex
-      flexDir={{ base: "column", lg: "row" }}
-      alignItems="flex-start"
-      gap={{ base: "xs", lg: "unset" }}
-    >
-      <SettingsLabel icon="actionLockClosed" text={label} />
-      <TextInput
-        marginLeft={{ sm: "m", lg: 0 }}
-        width="320px"
-        id={name}
-        name={name}
-        type="password"
-        isRequired
-        showLabel={false}
-        isClearable
-        showRequiredLabel={false}
-        labelText={label}
-        onChange={handler}
-        invalidText="Pin/passwords do not match."
-        isInvalid={isInvalid}
-      />
-    </Flex>
-  )
-}
+const PasswordFormField = forwardRef<TextInputRefType, PasswordFormFieldProps>(
+  ({ label, handler, name, isInvalid }: PasswordFormFieldProps, ref) => {
+    return (
+      <Flex
+        flexDir={{ base: "column", lg: "row" }}
+        alignItems="flex-start"
+        gap={{ base: "xs", lg: "unset" }}
+      >
+        <SettingsLabel icon="actionLockClosed" text={label} />
+        <TextInput
+          ref={ref}
+          marginLeft={{ sm: "m", lg: 0 }}
+          width="320px"
+          id={name}
+          name={name}
+          type="password"
+          isRequired
+          showLabel={false}
+          isClearable
+          showRequiredLabel={false}
+          labelText={label}
+          onChange={handler}
+          invalidText="Pin/passwords do not match."
+          isInvalid={isInvalid}
+        />
+      </Flex>
+    )
+  }
+)
+
+PasswordFormField.displayName = "PasswordFormField"
 
 const PasswordForm = ({ patronData, settingsState }: PasswordFormProps) => {
   const { getMostUpdatedSierraAccountData } = useContext(PatronDataContext)
@@ -74,10 +75,15 @@ const PasswordForm = ({ patronData, settingsState }: PasswordFormProps) => {
   })
   const { setStatus, setStatusMessage, editingField, setEditingField } =
     settingsState
+  const editingRef = useRef<HTMLButtonElement | null>()
+  const inputRef = useRef<TextInputRefType | null>()
 
   const cancelEditing = () => {
     setIsEditing(false)
     setEditingField("")
+    setTimeout(() => {
+      editingRef.current?.focus()
+    }, 0)
   }
 
   const validateForm =
@@ -162,6 +168,7 @@ const PasswordForm = ({ patronData, settingsState }: PasswordFormProps) => {
               }}
             >
               <PasswordFormField
+                ref={inputRef}
                 label="Enter current pin/password"
                 name="currentPassword"
                 handler={handleInputChange}
@@ -235,11 +242,15 @@ const PasswordForm = ({ patronData, settingsState }: PasswordFormProps) => {
             </Text>
             {editingField === "" && (
               <EditButton
+                ref={editingRef}
                 buttonLabel="Edit password"
                 buttonId="edit-password-button"
                 onClick={() => {
                   setIsEditing(true)
                   setEditingField("password")
+                  setTimeout(() => {
+                    inputRef.current?.focus()
+                  }, 0)
                 }}
               />
             )}
--- a/src/components/MyAccount/Settings/PasswordForm.tsx
+++ b/src/components/MyAccount/Settings/PasswordForm.tsx
@@ -41,9 +41,11 @@ const PasswordFormField = forwardRef<TextInputRefType, PasswordFormFieldProps>(
       >
         <SettingsLabel icon="actionLockClosed" text={label} />
         <TextInput
+          sx={{
+            width: { base: "100%", md: "300px" },
+          }}
           ref={ref}
-          marginLeft={{ sm: "m", lg: 0 }}
-          width="320px"
+          marginLeft={{ base: "m", lg: 0 }}
           id={name}
           name={name}
           type="password"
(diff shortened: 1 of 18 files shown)


src/components/MyAccount/Settings/PasswordForm.tsx
```diff
@@ -221,20 +240,25 @@ const PasswordForm = ({ patronData, settingsState }: PasswordFormProps) => {
...
             </Text>
             {editingField === "" && (
               <EditButton
+                ref={focusRef as React.Ref<HTMLButtonElement>}
+                buttonLabel="Edit password"
                 buttonId="edit-password-button"
                 onClick={() => {
                   setIsEditing(true)
                   setEditingField("password")
+                  setTimeout(() => {
```
7emansell, line 259: Vera said: Does this shift the focus to the text input? If so, there actually should be two different refs, just to make it explicit what is being focused on.

src/components/MyAccount/Settings/SettingsInputForm.tsx
```diff
@@ -42,9 +43,26 @@ const SettingsInputForm = ({
...
+  const editingRef = useRef<HTMLButtonElement | null>()
+
+  const focusLastInput = () => {
+    const lastIndex = tempInputs.length - 1
+    if (inputRefs.current[lastIndex]) {
+      inputRefs.current[lastIndex]?.focus()
+    }
+  }
+
+  useEffect(() => {
```
7emansell, line 56: @charmingduchess: 
> With three text inputs open for the phone settings, this focus is triggering on every keystroke input. ie Typing a single character into input 2 jumps the focus to input 3. I think you want an empty array to only run on the first render. I think this will make typescript unhappy but that's ok.

7emansell: updated it to `tempInputs.length` which I think is right because then it updates only when a new input is added/removed

src/components/MyAccount/Settings/SettingsInputForm.tsx
```diff
@@ -183,10 +211,15 @@ const SettingsInputForm = ({
                     sx={{
                       width: { base: "87%", md: "300px" },
                     }}
+                    ref={(el) => (inputRefs.current[index] = el)}
```
7emansell, line 214: @charmingduchess: 

> Not sure if you came across [this ("How to manage a list of refs using a ref callback")](https://react.dev/learn/manipulating-the-dom-with-refs), but your solution is react.dev approved! Looking at their solution, though, I think it might be important to remember to delete refs from the array once the input is removed from the dom.

I tried adding 5 inputs, then deleting one from the middle, and the count on the array of input refs actually went up to 6. It doesn't seem to cause any problems, but I think to avoid bugs down the line, the inputRefs array should match the inputs that are actually present on the dom.

Actually just encountered a bug! after opting to stay logged in from the timed logout modal, the focus is lost and the last input is null, which leads to a loss of focus. I think this can be avoided by removing refs as the inputs are removed.

7emansell: Updated to now remove ref (L104). Discussed offline but the issue with modal focus isn't going to be addressed here

7emansell:
Got a new ticket number so closed the old PR and copied Vera's comments over

src/components/MyAccount/Settings/PasswordForm.tsx
7emansell: @EdwinGuzman this is what we discussed– basically now I'm using 1 ref for everything (except in `SettingsInputForm`, which uses an array of refs to match the array of inputs, and can't really be combined with a button ref). This makes it confusing to predict where the focus will go just from reading the code, but it does work, so I'm torn.

EdwinGuzman: For readability and debugging where focus goes, it seems like having two refs is the approach to follow. When reading `useRef<HTMLButtonElement | TextInputRefType | null>`, I was thinking that ref is doing too many things. I think we discussed one ref where possible and updating the `ref.current` value but that's not the case here.

EdwinGuzman: To be clear, what other problems are there?

src/components/MyAccount/Settings/UsernameForm.tsx
```diff
@@ -151,21 +162,35 @@ const UsernameForm = ({ patron, usernameState }: UsernameFormProps) => {
...
         <>
-          <Text size="body1" sx={{ marginBottom: 0 }}>
+          <Text
+            size="body1"
+            sx={{ marginBottom: 0, width: { base: "l", sm: "250px" } }}
+          >
             {usernameInSierra}
           </Text>
           <EditButton
+            ref={focusRef as React.Ref<HTMLButtonElement>}
```
EdwinGuzman, line 172: From the comment in `PasswordForm` regarding breaking up the ref into two refs, doing so will allow this `as React.Ref...` to be removed.

src/components/MyAccount/Settings/SettingsInputForm.tsx
EdwinGuzman: The linter is saying that `focusLastInput` should also go in here. I'm not sure that's right but if it should go here, then the function should be wrapped in a useCallback.

src/components/MyAccount/Settings/SettingsInputForm.tsx
```diff
@@ -195,6 +228,7 @@ const SettingsInputForm = ({
                     isClearable
                     isClearableCallback={() => handleClearableCallback(index)}
                   />
+                  {index == 0 && <div style={{ width: "57px" }}> </div>}
```
EdwinGuzman, line 231: I'm not a fan of these magic numbers (72px on line 252). I don't have any good suggestions right now but I'll take a closer look.

src/components/MyAccount/Settings/SettingsInputForm.tsx
```diff
@@ -42,9 +43,26 @@ const SettingsInputForm = ({
 
   const [tempInputs, setTempInputs] = useState([...inputs])
 
+  const inputRefs = useRef<Array<TextInputRefType | null>>([])
+  const editingRef = useRef<HTMLButtonElement | null>()
+
+  const focusLastInput = () => {
+    const lastIndex = tempInputs.length - 1
+    if (inputRefs.current[lastIndex]) {
+      inputRefs.current[lastIndex]?.focus()
```
EdwinGuzman, line 52: I was going to ask if it was possible to use the `Map` function like the code example in the React docs resource. It'll work but I imagine this part here would make it difficult to keep track of the previous element to focus?

src/components/MyAccount/Settings/PasswordForm.tsx
charmingduchess: I think the issue is just conflicting directives. But I think me and Edwin agree this is a case where we can opt for readability.

src/components/MyAccount/Settings/SettingsInputForm.tsx
charmingduchess: With an Array we get the order maintained for free, which is crucial here. FWIW, the docs do say "This lets you maintain your own _array_ or a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)".

src/components/MyAccount/Settings/SettingsInputForm.tsx
charmingduchess: hmm yes I think the most idiomatic way of doing this is to make the `focusLastInput` a `useCallback` with a dependency of `tempInputs.length`. I'm not too concerned about it. However, I think it is good to see that the `useEffect` is going to run when that value updates, instead of having to check a `useCallback` for a dependency array.

src/components/MyAccount/Settings/PasswordForm.tsx
7emansell: All refs now split up for readability again 👍

src/components/MyAccount/Settings/SettingsInputForm.tsx
7emansell: Same– there are a lot of random fixed values in the designs which prevent using spacing tokens the way I'd like to

src/components/MyAccount/Settings/SaveCancelButtons.tsx
```diff
@@ -4,7 +4,7 @@ type SaveCancelButtonProps = {
   isDisabled?: boolean
   onCancel: () => void
   onSave: () => void
-  inputType?: "phones" | "emails"
+  inputType?: string
```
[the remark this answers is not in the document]
charmingduchess: was there a change lost here? i remember there being some conditional logic in the saveCancelButton component around this. I think we discussed it before, but I think it would be ok to add the extra `inputType="password"` in `PasswordForm.tsx`

charmingduchess (approved):
Looks good!

src/components/MyAccount/Settings/SettingsInputForm.tsx
EdwinGuzman: Right, I said I'll take another pass but I don't want to hold up pushing this out for another review. Okay to add comments post merge?

EdwinGuzman (approved):
This addresses all the a11y/vqa issues. The clear X button is an issue that we'll debug on the DS so for now, it won't be included in this update.

PR #413, opened 2024-12-09
SCC-3762 - Hold Confirmation Details

dgcohen:
## Ticket:

- JIRA ticket [SCC-3762](https://newyorkpubliclibrary.atlassian.net/browse/SCC-3762)

- Adds the function fetchHoldDetails that returns a patron ID (used to redirect to 404 on logged-in patron ID mismatch) and the pickup location (rendered in the item table).
- Renders a link to the bib in the confirmation message.
- Renders a search link beneath the FAQ
- Server-side redirect handling on confirmation page
- Adds a distinct Item details table for confirmation (and renames the request version)

## How has this been tested?

- Added unit tests

- NA

### Checklist:

[SCC-3762]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-3762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/HoldPages/HoldConfirmationItemDetails.tsx
charmingduchess, line 1: Can this be refactored to not use ternaries? Instead, something like `{pickupLocationLabel && <.../>}`.

charmingduchess (changes requested):

src/components/HoldPages/HoldConfirmationItemDetails.tsx
dgcohen: I tried to remove the ternary but it ended up breaking the page due to a DS issue where the List component errors out with no useful error message when it receives a null child instead of a fragment. I restored the ternaries and will bring this up with the DS team.

EdwinGuzman (commented):
Looks good but some comments.

The description `List` issue is interesting. I think the issue is with `DetailElement` not returning a proper default value. Share the bib where this error is happening to replicate and test, thanks.

src/components/HoldPages/HoldConfirmationItemDetails.tsx
dgcohen: I was able to fix this by passing the list children as a listItems prop in DetailElement nstead of as children.

pages/hold/confirmation/[id].tsx
```diff
@@ -29,8 +42,14 @@ interface HoldConfirmationPageProps {
...
 export default function HoldConfirmationPage({
   isEDD = false,
+  pickupLocationLabel,
+  discoveryBibResult,
 }: HoldConfirmationPageProps) {
   const metadataTitle = `Request Confirmation | ${SITE_NAME}`
+
+  const bib = new Bib(discoveryBibResult)
+  const item = bib?.items[0]
+
```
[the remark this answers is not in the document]
Comment (dgcohen) [reply pages/hold/confirmation/[id].tsx:52]: We've run into this issue before where models can't be stringified since they contain functions. What we can do is pass just the needed fields but this would slightly limit the flexibility and increase the number of props passed to the component.

The alternative is to remove the model initialization on the server side to avoid running it twice, but we would need to call the discovery response parameters directly which might be fine.

src/components/HoldPages/HoldConfirmationItemDetails.tsx
dgcohen: Actually nevermind, that didn't work.

dgcohen:
> Looks good but some comments.
> 
> The description `List` issue is interesting. I think the issue is with `DetailElement` not returning a proper default value. Share the bib where this error is happening to replicate and test, thanks.

I'm a little reluctant to replace with a <dl> element since this doesn't really address the root issue of the List component not being able to accept null children and would involve duplication of styles when the fragment ternary syntax works.

pages/hold/confirmation/[id].tsx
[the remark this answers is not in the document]
Comment (EdwinGuzman) [reply pages/hold/confirmation/[id].tsx:52]: You don't need `bib` in this component so you can only return `item`. It would limit to just the current fields you need but you don't need anything else according to the designs (at least for now).

EdwinGuzman:
> I'm a little reluctant to replace with a dl element since this doesn't really address the root issue of the List component not being able to accept null children and would involve duplication of styles when the fragment ternary syntax works.

Yeah, I agree. It's not a blocker so it can wait for a better DS fix, but you'll just have to live with the not-so-great-looking syntax for now.

charmingduchess (approved):

src/components/BibPage/BibDetail.tsx
```diff
@@ -14,6 +14,7 @@ import type {
 } from "../../types/bibDetailsTypes"
 import { rtlOrLtr, isItTheLastElement } from "../../utils/bibUtils"
 import type { ReactNode } from "react"
+import { child } from "winston"
```
EdwinGuzman, line 17: Don't think you meant this but it doesn't seemed used.

PR #415, opened 2024-12-10
SCC-3762 - Patron Eligibility Errors

dgcohen:
## Ticket:

- JIRA ticket [SCC-3762](https://newyorkpubliclibrary.atlassian.net/browse/SCC-3762)

- Server function fetchPatronEligibility that fetches a patron's eligibility status from Discovery API
- HoldRequestErrorBanner component responsible for showing the user request errors based on the request error status (including ineligibilty reasons).
- Eligibility validation on the server side.

## How has this been tested?

- Unit tests

- NA

### Checklist:

[SCC-3762]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-3762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/HoldPages/HoldRequestErrorBanner.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  const onContact = (metadata: ItemMetadata) => {
+    setItemMetadata(metadata)
+    onOpen()
+  }
+
+  return (
+    <Banner
+      type="negative"
+      heading={HOLD_PAGE_ERROR_HEADINGS?.[errorStatus] || null}
```
EdwinGuzman, line 51: Does this work if it's both dot notation and bracket?

src/components/HoldPages/HoldRequestErrorBanner.tsx
```diff
@@ -0,0 +1,177 @@
...
+      data-testid="hold-request-error"
+      // TODO: Ask DS team to make button link variant match the default link styles
```
EdwinGuzman, line 54: Start creating tickets for these types of issues and CC me and Marty.

EdwinGuzman: Link the design and the PR where the issue comes up

src/components/HoldPages/HoldRequestErrorBanner.tsx
```diff
@@ -0,0 +1,177 @@
...
+            })()}
+          </Box>
+        </>
+      }
+      mb="s"
+    />
+  )
+}
+
+const PatronErrors = ({
```
EdwinGuzman, line 119: This component might get lost within the file and maybe it can be its own file. Even though it's a small component, I think you can break out the logic from the component.

EdwinGuzman (commented):
Some minor comments. Can you share a few test patron accounts to try to get all use cases locally? (this can be done on slack)

src/components/HoldPages/HoldRequestErrorBanner.tsx
dgcohen: Yeah it allows you to have a null check on an attribute with quotes (helpful if there's a digit or something) but it's not necessary here.

dgcohen: Actually, it is necessary because it's a dynamic key.

pages/api/hold/request/[id]/edd.ts
```diff
@@ -1,23 +1,43 @@
...
+
+    if (patronEligibilityStatus.status === 401) {
+      switch (jsEnabled) {
+        case true:
+          return res.status(401).json({
+            patronEligibilityStatus,
+          })
+
+        // Server side redirect on ineligibility when Js is disabled
+        // TODO: Move this to seaprate API route
```
Comment (EdwinGuzman) [inline pages/api/hold/request/[id]/edd.ts:35]: Why do you want to move this to a separate API route?

src/server/api/hold.ts
```diff
@@ -230,3 +176,65 @@ export async function postEDDRequest(
...
+    }
+
+    return { status: 200, ...eligibilityResult }
+  } catch (error) {
+    console.error(
+      `Error fetching hold request eligibility in fetchPatronEligibility server function, patronId: ${patronId}`,
+      error.message
+    )
+
+    return { status: 500 }
```
EdwinGuzman, line 238: Can you add a message and say it was a server issue? It seems this would be the case when the NYPL Platform eligibility endpoint is down, right?

EdwinGuzman (commented):
Latest round looks good. Some comments but I think these should be fixed:

1) When going to an EDD page, the exact reason does not come up. Shouldn't it? It does come up for the hold request form:
Screenshot 2024-12-12 at 12 58 56 PM

2) When a patron cannot place holds on either Holds or EDD, disable the form. I can still submit a request on the on-site use page and can fill out the EDD form:
Screenshot 2024-12-12 at 12 57 14 PM

3) The following is low priority so not necessary for this PR. The no-js form submission redirects to a broken result:
Screenshot 2024-12-12 at 12 58 08 PM

pages/api/hold/request/[id]/edd.ts
Comment (dgcohen) [reply pages/api/hold/request/[id]/edd.ts:35]: To be able to remove the jsEnabled flag from the client call.

pages/hold/request/[id]/index.tsx
```diff
@@ -85,40 +94,44 @@ export default function HoldRequestPage({
       const response = await fetch(`${BASE_URL}/api/hold/request/${holdId}`, {
         method: "POST",
         body: JSON.stringify({
-          patronId: patronId.value,
-          source: source.value,
-          pickupLocation: pickupLocation.value,
+          patronId: patronId?.value,
```
Comment (charmingduchess) [inline pages/hold/request/[id]/index.tsx:97]: What case would allow for this request to be made with undefined values for these required fields? I think the optional chaining implies that case is expected.

pages/hold/request/[id]/index.tsx
```diff
@@ -85,40 +94,44 @@ export default function HoldRequestPage({
...
+      switch (response.status) {
+        // Patron is ineligible to place holds
+        case 401:
+          setFormPosting(false)
+          setErrorStatus("patronIneligible")
+          setPatronEligibilityStatus(responseJson?.patronEligibilityStatus)
+          break
+        // Server side error placing the hold request
+        case 500:
+          setFormPosting(false)
```
Comment (charmingduchess) [inline pages/hold/request/[id]/index.tsx:114]: Here's a suggestion a la Refactoring...
You can pull out the invocations within each case to functions that summarize the comments above them. That way, you can also reuse a `serverSideErrorPlacingHoldRequest` call  in the catch block on line 115.

pages/hold/request/[id]/index.tsx
```diff
@@ -138,18 +151,11 @@ export default function HoldRequestPage({
         {/* Always render the wrapper element that will display the
           dynamically rendered notification for focus management */}
         <Box tabIndex={-1} ref={bannerContainerRef}>
-          {alert && (
-            <HoldRequestBanner
+          {errorStatus && (
```
Comment (charmingduchess) [inline pages/hold/request/[id]/index.tsx:154]: I like the look of this refactor.

src/components/HoldPages/HoldRequestErrorBanner.tsx
```diff
@@ -0,0 +1,112 @@
...
+      // TODO: Ask DS team to make button link variant match the default link styles
+      sx={{
+        button: {
+          color: "ui.error.primary",
+          "&:hover": { color: "ui.error.primary" },
+        },
+      }}
+      content={
+        <Box>
+          {HOLD_PAGE_CONTACT_PREFIXES?.[errorStatus] ? (
```
charmingduchess, line 56: can this ternary be replaced with a `&&`?

src/components/HoldPages/PatronIneligibilityErrors.tsx
```diff
@@ -0,0 +1,73 @@
...
+  if (!hasSpecificReason) return null
+
+  return (
+    <Box mt="xs">
+      <Text mb="xs">This is because:</Text>
+      <List
+        type="ul"
+        margin={0}
+        listItems={[
+          ...(expired
```
charmingduchess, line 40: I think this awkward array construction can be avoided if you pass the children in explicitly instead of as ListItems. eg 
```
<List type="ul" margin={0}>
        {expired && (
          <>
            Your account has expired -- Please see{" "}...
```

__test__/pages/hold/eddRequestPage.test.tsx
```diff
@@ -44,6 +44,26 @@ const mockReq = {
   },
 }
 
+const fillRequiredEDDFormFields = async () => {
+  // Fill in all required form fields
+  await userEvent.type(
+    screen.getByPlaceholderText(EDD_FORM_FIELD_COPY.emailAddress.placeholder),
```
charmingduchess, line 50: this is a great pattern

__test__/pages/hold/eddRequestPage.test.tsx
```diff
@@ -335,11 +331,94 @@ describe("EDD Request page", () => {
...
+      )
+    })
+
+    it("shows an error listing ineligibility reasons when the patron is ineligibile to place holds", async () => {
+      await waitFor(() => {
+        expect(screen.getByTestId("hold-request-error")).toBeInTheDocument()
+      })
+
+      expect(
+        screen.getByText("There is a problem with your library account.", {
```
charmingduchess, line 386: Can you use that same pattern here with the constants for the patron  error text? Or is that not possible because the patronIneligibilityErrors include JSX elements?

charmingduchess (changes requested):

__test__/pages/hold/eddRequestPage.test.tsx
dgcohen: I can't use that pattern for the ineligibility reasons because of the JSX but I can for the header.

pages/api/hold/request/[id]/edd.ts
```diff
@@ -40,7 +60,7 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
     // JS-Disabled functionality
 
     // Redirect to confirmation page
-    res.redirect(
+    return res.redirect(
```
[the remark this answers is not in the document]
Comment (dgcohen) [reply pages/api/hold/request/[id]/edd.ts:64]: it was but i added it for consistency after adding it elsewhere as one of Edwin's suggestions.

pages/api/hold/request/[id]/edd.ts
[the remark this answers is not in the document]
Comment (dgcohen) [reply pages/api/hold/request/[id]/edd.ts:1]: yes but i think we should do this as a follow-up optimization ticket since it'll involve potentially refactoring the fetching function as well.

EdwinGuzman (approved):
The updates look good but tbh I cannot test this locally. I get merge conflicts when I pull locally. I think it's because of a force push in the history. So I will approve based on the updates but will have to test on `train-research-catalog` once it's deployed.

pages/hold/request/[id]/index.tsx
Comment (charmingduchess) [reply pages/hold/request/[id]/index.tsx:106]: This maybe was unclear. Can you remove the optional chaining? I don't see why we would want to send a hold request that didn't have these values defined.

charmingduchess (changes requested):

pages/hold/request/[id]/index.tsx
Comment (charmingduchess) [reply pages/hold/request/[id]/index.tsx:106]: @dgcohen

charmingduchess (approved):

PR #421, opened 2024-12-17
Scc 4142/spec coll messaging

charmingduchess:
This is a preliminary refactor to facilitate [this ticket](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4124). Would love some eyes on this before I go all the way down this path. 
To summarize, I created an ItemAvailability model which builds a string that represents it's availability, location, and special collections status. The model also handles returning the correct message component.

## Ticket:

- JIRA ticket [xxx](url)

-

-

### Checklist:


src/models/ItemAvailability.tsx
dgcohen, line 1: Is there a reason this functionality needs its own class? I feel like the key building could be added to the item model with a component handling the presentation logic.

src/components/ItemTable/ItemAvailabilityComponents/ContactALibrarian.tsx
dgcohen, line 1: Small stylistic preference, but can we rename the directory this is in to ItemAvailability? Don't think we need the word components if it's in the main components directory.

src/components/ItemTable/ItemAvailabilityComponents/ContactALibrarian.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const onContact = (metadata: ItemMetadata) => {
+    setItemMetadata(metadata)
+    onOpen()
+  }
+  return (
+    <>
+      {" - Please "}
+      <Button
+        id="contact-librarian"
+        buttonType="link"
```
dgcohen, line 18: This reminds me that I wanna get an unstyled button variant into the DS. Will bring this up with Marty.

EdwinGuzman: Yes, this came up in the context of the error message and the button in the banner in your PR so make sure to include that. 

In this case, this `ContactALibrarian` button should not care where it's rendered, so the update should probably be made in the banner component.

src/models/ItemAvailability.tsx
charmingduchess: it's so focused on a specific cluster of data that I think it justifies its own class. I am going to move the presentation logic to the component.

PR #424, opened 2024-12-20
SCC-4428 - Pass item source in kebab case to hold forms

dgcohen:
## Ticket:

- JIRA ticket [SCC-4428 ](url)

- We had missed a small bit of functionality on DFE that converts the source to kebab case before being sent in hold request params.

## How has this been tested?

- Added a unit test

- NA

### Checklist:

### DIFF
--- a/src/models/Item.ts
+++ b/src/models/Item.ts
@@ -82,7 +82,7 @@ export default class Item {
       .includes("all")
   }
 
-  get holdRequestSource(): string {
+  get formattedSourceForHoldRequest(): string {
     return convertCamelToShishKabobCase(this.source)
   }
 
--- a/src/models/Item.ts
+++ b/src/models/Item.ts
@@ -12,6 +12,7 @@ import {
   locationEndpointsMap,
 } from "../utils/itemUtils"
 import { appConfig } from "../config/config"
+import { convertCamelToShishKabobCase } from "../utils/appUtils"
 
 /**
  * The Item class contains the data and getter functions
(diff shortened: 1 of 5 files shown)


src/utils/appUtils.ts
```diff
@@ -140,3 +140,19 @@ export const convertToSentenceCase = (str: string) =>
...
+ *
+ * e.g. camelToShishKabobCase("RecapPul")
+ *        => "recap-pul"
+ *      camelToShishKabobCase("firstCharCanBeLowerCase")
+ *        => "first-char-can-be-lower-case"
+ */
+export const convertCamelToShishKabobCase = (str: string) =>
+  str
+    // Change capital letters into "-{lowercase letter}"
+    .replace(/([A-Z])/g, (c, p1, i) => {
```
EdwinGuzman, line 155: What's the `p1` and `i` for?

EdwinGuzman (approved):
Just a question

src/utils/appUtils.ts
charmingduchess: i think `i` is the index needed to skip the first character, and p1 is a placeholder arg

src/models/Item.ts
```diff
@@ -81,6 +82,10 @@ export default class Item {
       .includes("all")
   }
 
+  get holdRequestSource(): string {
```
charmingduchess, line 85: this sounds like it means where the hold request is coming from (which is something that we think about on LSP, though not necessarily in this app). to avoid confusion, maybe `formattedSourceForHoldRequest`

src/utils/appUtils.ts
dgcohen: Just updated these var names for clarity

src/models/Item.ts
dgcohen: good call! done

EdwinGuzman (approved):

charmingduchess (approved):

PR #427, opened 2024-12-20
SCC-4427 - VQA Round 2 Fixes

dgcohen:
## Ticket:

- JIRA ticket [SCC-4427](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4427)

- Updates error banner copy, styles, and adds HoldContactButton component for ease of re-use.
- Only renders ineligibility reasons as a List when more than one (Apoorva request not in the VQA doc)
- Disables the Submit button when the patron is ineligible to place holds.
- Misc styling changes on hold request forms
- Remove isRequired from EDD form fields to prevent chrome's native handling of focus on errors.

## How has this been tested?

 - Updated unit tests

- NA

### Checklist:

[SCC-4427]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/HoldPages/HoldContactButton.tsx
+++ b/src/components/HoldPages/HoldContactButton.tsx
@@ -1,0 +1,54 @@
+import { Button } from "@nypl/design-system-react-components"
+import { useContext, type ReactNode } from "react"
+
+import type Item from "../../models/Item"
+import type { ItemMetadata } from "../../types/itemTypes"
+import { FeedbackContext } from "../../context/FeedbackContext"
+
+interface HoldContactButtonProps {
+  item: Item
+  children: ReactNode
+}
+
+export const HoldContactButton = ({
+  item,
+  children,
+}: HoldContactButtonProps) => {
+  const { onOpen, setItemMetadata } = useContext(FeedbackContext)
+
+  const onContact = (metadata: ItemMetadata) => {
+    setItemMetadata(metadata)
+    onOpen()
+  }
+
+  return (
+    <Button
+      id="hold-contact"
+      onClick={() =>
+        onContact({
+          id: item.id,
+          barcode: item.barcode,
+          callNumber: item.callNumber,
+          bibId: item.bibId,
+          notificationText: `Request failed for call number ${item.callNumber}`,
+        })
+      }
+      buttonType="link"
+      // TODO: Ask DS team to make button link variant match the default link styles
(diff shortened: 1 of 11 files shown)


charmingduchess (approved):

src/components/HoldPages/HoldContactButton.tsx
EdwinGuzman, line 1: Nice update

EdwinGuzman (commented):
The VQA updates looks good but the removal of `isRequired` raises some problems. **There needs to be validation for empty strings and manually trigger the error for each field.**

- Right now, if you got to an EDD page and not fill out the form, you can hit submit and it'll take you to the confirmation page. This is not correct.

**Expected**:
- If you tab through the inputs and leave them blank, you'll get the invalid text. If you attempt to hit submit, then you'll be focused on the "Starting page number" field (which is correct) and the error banner appears.

Additionally, add " (required)" to each input field label so it's clear to the patron that the field is required, even though it's not programmatically required by the browser. When the validation is added, it be clear that empty fields are not allowed.

I'm not flagging this as "Request changes" so that once this is fixed, you can merge it in next week without my approval.

dgcohen:
> The VQA updates looks good but the removal of `isRequired` raises some problems. **There needs to be validation for empty strings and manually trigger the error for each field.**
> 
> * Right now, if you go to an EDD page and not fill out the form, you can hit submit and it'll take you to the confirmation page. This is not correct.
> 
> **Expected**:
> 
> * If you tab through the inputs and leave them blank, you'll get the invalid text. If you attempt to hit submit, then you'll be focused on the "Starting page number" field (which is correct) and the error banner appears. This action should happen when the patron attempts to hit submit right away without focusing on any input fields.
> 
> Additionally, add " (required)" to each input field label so it's clear to the patron that the field is required, even though it's not programmatically required by the browser. When the validation is added, it'll be clear that empty fields are not allowed.
> 
> I'm not flagging this as "Request changes" so that once this is fixed, you can merge it in next week without my approval.

Good catch! This was actually caused by a subtle bug related to how i was setting the validated state. Just pushed a fix.

PR #429, opened 2025-01-02
Scc 4405/hold pages js disabled

dgcohen:
## Ticket:

- JIRA ticket [SCC-4405](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4405)

- Adds server-side validation on EDD hold requests
- Fixes bugs causing errors on No-js hold requests (related to JSON parsing/query string formatting)
- Adds formState and validatedFields query params to populate field values/validation state on refreshes/js-disabled redirects.
- Renames initialEDDInvalidFields to defaultValidatedEDDFields

## How has this been tested?

- Added unit tests

- I thought about adding autofocus when js is disabled but it ended up adding a bunch of code, so not sure if it's worth it. Let me know if you think this is important (will check with Clare as well) and I'll add it back.

### Checklist:

[SCC-4405]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pages/hold/request/[id]/edd.tsx
+++ b/pages/hold/request/[id]/edd.tsx
@@ -83,8 +83,8 @@ export default function EDDRequestPage({
   const router = useRouter()
   const isLoading = useLoading()
 
-  // Populate form values from query string in case of js-disabled server-side redirect
-  const formStateFromQuery = router.query?.formState
+  // Derive form values from query string in case of js-disabled server-side redirect
+  const formStateFromServer = router.query?.formState
     ? JSON.parse(router.query.formState as string)
     : []
 
@@ -93,7 +93,8 @@ export default function EDDRequestPage({
     emailAddress: patronEmail,
     patronId,
     source: item.formattedSourceForHoldRequest,
-    ...formStateFromQuery,
+    // Override values with server form state in the case of a no-js request
+    ...formStateFromServer,
   })
 
   useEffect(() => {
--- a/pages/hold/request/[id]/edd.tsx
+++ b/pages/hold/request/[id]/edd.tsx
@@ -84,7 +84,9 @@ export default function EDDRequestPage({
   const isLoading = useLoading()
 
   // Populate form values from query string in case of js-disabled server-side redirect
-  const formStateFromQuery = JSON.parse(router.query?.formState as string)
+  const formStateFromQuery = router.query?.formState
+    ? JSON.parse(router.query.formState as string)
+    : []
 
   const [eddFormState, setEddFormState] = useState<EDDRequestParams>({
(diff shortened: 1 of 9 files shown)


pages/hold/request/[id]/edd.tsx
```diff
@@ -188,8 +196,10 @@ export default function EDDRequestPage({
   )
 }
 
-export async function getServerSideProps({ params, req, res }) {
+export async function getServerSideProps({ params, req, res, query }) {
   const { id } = params
+  const { formInvalid } = query
+  const serverValidationFailed = formInvalid ? JSON.parse(formInvalid) : false
```
Comment (EdwinGuzman) [inline pages/hold/request/[id]/edd.tsx:202]: Is it possible to pass the `formInvalid` object to the component so you can then remove the `router.query?.formState` call? The values are already being parsed in the server so it doesn't seem necessary to do it again in the client.

EdwinGuzman (commented):
This works very well. I left a comment about the general architecture but some questions:
- shouldn't the query params be escaped?
- it feels a bit unsafe having the patron's email in the query params. I don't think there's anything that can be done.
- What if a value like `requestNotes` is very long? Thankfully, I think the limit is big so there's no need to worry but something to keep in mind.
  
example: `validatedFields=[{"key":"emailAddress","isInvalid":false},{"key":"startPage","isInvalid":true},{"key":"endPage","isInvalid":true},{"key":"chapterTitle","isInvalid":true}]&formState={"source":"sierra-nypl","emailAddress":"spaghetti@pasta.com","startPage":"","endPage":"","chapterTitle":"","author":"","date":"","volume":"","issue":"","requestNotes":""}`

charmingduchess (approved):

pages/hold/request/[id]/edd.tsx
Comment (dgcohen) [reply pages/hold/request/[id]/edd.tsx:202]: The router.query?.formState call is responsible for pre-populating the form fields, whereas the formInvalid query param is just responsible for setting the page's status. I could derive the "invalid" status using the form's state, but it would add some additional logic to the component, and I felt that this was cleaner since there was already a switch statement in getServerSideProps that sets the page status before the component renders.

EdwinGuzman (approved):

EdwinGuzman (approved):

pages/hold/request/[id]/edd.tsx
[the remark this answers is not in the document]
Comment (dgcohen) [reply pages/hold/request/[id]/edd.tsx:1]: That's a good suggestion, will open a ticket to do that as a follow up.

PR #447, opened 2025-01-13
Fix nonsense cache

charmingduchess:
Paul found a memory issue due to this cache being WACKY! 
- Updated the client cache be an object, 
- update cache setter with keys that match what we are querying the cache for!


EdwinGuzman (approved):

EdwinGuzman:
Update the changelog as well for this hotfix

nonword (approved):

PR #452, opened 2025-01-17
Hotfix for Duplicated "View All" Loading Text on the Bib Page

dgcohen:
This hotfix resolves an issue where the "View All" loading text is being rendered twice on the Bib page on production. The issue arose due to a recent change in the design system, which led to the label component being rendered twice.

Changes Made:

- Removed the duplicate label component that was being rendered, ensuring the "View All" loading text is only displayed once.
- Styled the ProgressIndicator component to match the previous design, ensuring visual consistency with the old styles.

### DIFF
--- a/src/components/ItemTable/ItemTableControls.tsx
+++ b/src/components/ItemTable/ItemTableControls.tsx
@@ -63,19 +63,15 @@ const ItemTableControls = ({
             indicatorType="circular"
             isIndeterminate
             marginLeft="auto"
+            labelPlacement="right"
             sx={{
-              "> div": {
-                display: "flex",
-                justifyContent: "flex-start",
-                alignItems: "center",
-                flexDirection: "row",
-                "> div": {
-                  marginRight: "var(--nypl-space-xs)",
-                },
-              },
+              "div[role=progressbar]": { marginRight: "xs" },
               label: {
-                fontWeight: "var(--nypl-fontWeights-medium)",
-                fontSize: "var(--nypl-fontSizes-desktop-body-body2)",
+                fontWeight: "medium",
+                fontSize: {
+                  base: "mobile.body.body2",
+                  md: "desktop.body.body2",
+                },
                 marginBottom: 0,
               },
             }}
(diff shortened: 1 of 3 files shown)


src/components/ItemTable/ItemTableControls.tsx
```diff
@@ -56,38 +56,30 @@ const ItemTableControls = ({
...
+                alignItems: "center",
+                flexDirection: "row",
+                "> div": {
+                  marginRight: "var(--nypl-space-xs)",
+                },
+              },
+              label: {
+                fontWeight: "var(--nypl-fontWeights-medium)",
+                fontSize: "var(--nypl-fontSizes-desktop-body-body2)",
+                marginBottom: 0,
```
EdwinGuzman, line 79: These are inline css so you should be able to use the design token equivalent instead of the CSS var.

src/components/ItemTable/ItemTableControls.tsx
```diff
@@ -56,38 +56,30 @@ const ItemTableControls = ({
...
-              mb={0}
-              // Label component does not expect tabIndex prop, so we are ignoring the typescript error that pops up.
-              // Add any additional props above this for typescript validation.
-              // @ts-expect-error
-              tabIndex={-1}
-            >
-              {viewAllLoadingMessage}
-            </Label>
-          </Box>
+          <ProgressIndicator
```
EdwinGuzman, line 59: There is a new prop `labelPlacement` that you should use. You might be able to remove a lot of the CSS here.

dgcohen: Nice! Just added it and i was able to get rid of a lot the styles.

charmingduchess (commented):
Looks good. Could you also revert [this test](https://github.com/NYPL/research-catalog/blob/main/__test__/pages/bib/bibPage.test.tsx#L317) to be a queryBy and not queryAllBy? It was updated when this duplication started happening.

src/components/ItemTable/ItemTableControls.tsx
EdwinGuzman: Yes, great.

charmingduchess (approved):

EdwinGuzman (approved):

PR #466, opened 2025-02-04
add searched_from to prod

charmingduchess:


EdwinGuzman:
Is there a reason why this is being merged/deployed directly to `production`?

charmingduchess:
Just a mistake.

PR #467, opened 2025-02-04
SCC-4472 - Add google tag manager snippet

dgcohen:
## Ticket:

- JIRA ticket [SCC-4472](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4472)

- Adds Google Tag Manager (GTM) snippets to the app.

## How has this been tested?

- Verified that the GTM snippets are rendering correctly in the application.
- This has not yet been tested on the Google Analytics dashboard.
- In Next.js, the <noscript> tag always renders. While this shouldn't impact analytics data, we should verify that page view and event tracking numbers are accurate.

- NA

### Checklist:

[SCC-4472]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -85,6 +85,7 @@ export const appConfig: AppConfig = {
     Q9RrzMLTqxKnVVn2+hVpk4F/8tMsGCdd4s/AJqEQBy5lsq7ji1B63XYqi5fc1SnJ
     EQIDAQAB
     -----END PUBLIC KEY-----`,
+  gtmTrackingId: "GTM-RKWC",
   features: {
     reverseProxyEnabled: {
       development: false,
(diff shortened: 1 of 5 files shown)


src/config/config.ts
```diff
@@ -85,6 +85,7 @@ export const appConfig: AppConfig = {
     Q9RrzMLTqxKnVVn2+hVpk4F/8tMsGCdd4s/AJqEQBy5lsq7ji1B63XYqi5fc1SnJ
     EQIDAQAB
     -----END PUBLIC KEY-----`,
+  gtmTrackingId: "GTM-RKWC",
```
EdwinGuzman, line 88: I don't think you need to have this id in the config file. It's a bit more work having to declare it here and importing it in each file where it's needed. If I understand correctly, there's only one instance of GTM so there is no need to have a "qa" and "prod" one, so, if it makes sense, put the id directly in the code snippets.

EdwinGuzman (approved):

charmingduchess (approved):
Not totally sure what I'm looking at here, but it runs!

PR #469, opened 2025-02-07
Scc 4372/num renewals

charmingduchess:
## Ticket:

- JIRA ticket [SCC-4372](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4372)

- add number of renewals to checkout tab
- update spacing for checkouts, fees, and holds tabs

## How has this been tested?

Updated checkouts processing tests to include number of renewals counts, checkouts tab tests for updated text. I also updated a test for the testPickupTimes page, which was breaking. Does the Form component from DS not have an aria role of 'form' anymore?

- NA

### Checklist:

[SCC-4372]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/MyAccount/CheckoutsTab/CheckoutsTab.tsx
```diff
@@ -30,24 +30,43 @@ const CheckoutsTab = () => {
     "Manage checkout",
   ]
   const [checkoutToRenew, setCheckoutToRenew] = useState<string>(null)
+  const tableStyling = checkoutsHeaders.map((_, index) => {
+    const width = index === 0 ? null : "16%"
```
dgcohen, line 34: This works, but could it be achieved using only CSS, perhaps with the :first-child pseudo-selector?

src/components/MyAccount/CheckoutsTab/CheckoutsTab.tsx
```diff
@@ -30,24 +30,43 @@ const CheckoutsTab = () => {
...
     checkout.callNumber,
     checkout.dueDate,
     checkout.isResearch ? null : (
-      <RenewButton
-        setCheckoutToRenew={setCheckoutToRenew}
-        isCheckoutRenewing={checkoutToRenew === checkout.id}
-        checkout={checkout}
-        patron={patron}
-      />
+      <Flex alignItems="end" justifyContent={"space-between"}>
```
dgcohen, line 45: No need for curly braces here

src/components/MyAccount/FeesTab/FeesTab.tsx
```diff
@@ -17,6 +17,8 @@ const FeesTab = ({ fines }: { fines: Fine }) => {
       columnHeadersBackgroundColor={"ui.gray.x-light-cool"}
       columnHeaders={feesHeaders}
       tableData={feesData.concat(finalRow)}
+      isScrollable
+      columnStyles={feesHeaders.map((_) => ({ minWidth: "120px" }))}
```
dgcohen, line 21: Does this need to be an array with minWidth on each item? Could it be simplified by applying a minWidth style instead?

src/components/MyAccount/RequestsTab/RequestsTab.tsx
```diff
@@ -44,6 +44,11 @@ const RequestsTab = () => {
     "Pickup by",
     "Manage request",
   ]
+  const tableStyles = holdsHeaders.map((_, index) => {
+    const width = index === 0 ? null : "16%"
```
dgcohen, line 48: Same question about psuedoselectors

src/components/MyAccount/CheckoutsTab/CheckoutsTab.tsx
```diff
@@ -30,24 +30,43 @@ const CheckoutsTab = () => {
...
+          setCheckoutToRenew={setCheckoutToRenew}
+          isCheckoutRenewing={checkoutToRenew === checkout.id}
+          checkout={checkout}
+          patron={patron}
+        />
+        {checkout.numberOfRenewals ? (
+          <Text
+            size="caption"
+            isItalic
+            backgroundColor={{ base: "green", md: "yellow", mt: "tomato" }}
```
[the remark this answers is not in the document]
charmingduchess: omg no! I was just having trouble getting the the breakpoints to work. they clearly don't because this never showed up. Good catch.

src/models/modelTests/MyAccount.test.ts
```diff
@@ -83,8 +83,8 @@ describe("MyAccountModel", () => {
         },
       }
       const fetcher = new MyAccount(mockSierraClient, "12345")
-      const processedCheckouts = await fetcher.getCheckouts()
-      expect(processedCheckouts).toStrictEqual(processedCheckouts)
+      const builtProcessedCheckouts = await fetcher.getCheckouts()
```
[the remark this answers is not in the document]
charmingduchess: Yea I wasn't sure about this. The fixture is called processedCheckouts, and I wasn't sure what to call the actual data returned from the method.

src/components/MyAccount/FeesTab/FeesTab.tsx
charmingduchess: Yea this is a question. The `columnStyles` expects that, but it really makes more sense for when the different columns require different styles. Also, regarding your comment above about using psuedoselectors, using CSS instead of the prop *feels* like an override, even though it's for expected customization. I'm inclined to use the prop just because it's there, but I'm not totally sold.

dgcohen (approved):

PR #483, opened 2025-03-03
Scc 4417/format filters

charmingduchess:
## Ticket:

- JIRA ticket [4552](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4552)

- Replaces materialType with format in all cases across the repo. This is because we are now using a newly indexed format value instead of the more broadly defined materialTypes for format.

## How has this been tested?
Running searches and using available formats to filter results. 

-

### Checklist:


src/types/searchTypes.ts
```diff
@@ -8,11 +8,11 @@ type SubjectLiteral = string
 type ContributorLiteral = string
 type CreatorLiteral = string
 type Issuance = string
-type MaterialTypeFilter = string
+type formatFilter = string
```
dgcohen, line 11: Should this be capitalized since it's a type?

CHANGELOG
```diff
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Updated
 
 - Send new searches back to home, not /search [SCC-4487](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4487)
-- Removed legacy barcode hotfix (SCC-4474)
```
dgcohen, line 13: This line doesn't need to be removed

dgcohen (changes requested):
Looks like a straightforward change despite the massive diff. A couple of small comments but otherwise looks good!

dgcohen (approved):

PR #485, opened 2025-03-05
update logs

charmingduchess:
There are mysterious errors blocking hold requests in QA. Updating the logs for more granularity to investigate. Also added client-level logging to nypl client gets and posts, and sierra client gets, posts and puts. I did most of the updated requests, and they work! However, I would appreciate some scrutiny on my instance method manipulation.

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -42,7 +42,6 @@ const initializeLogger = () => {
     }
     return JSON.stringify(logObject)
   })
-  console.log(process.env.LOG_LEVEL || "error")
   const transports = [
     new winston.transports.Console({ level: process.env.LOG_LEVEL || "error" }),
     new winston.transports.File({
@@ -51,7 +50,7 @@ const initializeLogger = () => {
       format: combine(winston.format.uncolorize(), formatter),
       maxsize: 5242880,
       maxFiles: 5,
-      level: process.env.LOG_LEVEL || "error"
+      level: process.env.LOG_LEVEL || "error",
     }),
   ]
 
--- a/logger.js
+++ b/logger.js
@@ -42,21 +42,22 @@ const initializeLogger = () => {
     }
     return JSON.stringify(logObject)
   })
-
+  console.log(process.env.LOG_LEVEL || "error")
   const transports = [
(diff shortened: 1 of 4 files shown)


dgcohen (approved):

nonword (approved):
Approving on condition that this not go to prod as is, since we don't want to be logging request details there. Note you can theoretically enable debug logging inside the nypldataapiclient via `log_level` option to the constructor, so you could do:
```
    const nyplApiClient = new NyplApiClient({
      ...
      log_level: process.env.NYPL_API_LOG_LEVEL || 'error'
    })
```
And for sierra-client you may be able to use [setLogLevel](https://github.com/NYPL-discovery/sierra-wrapper/blob/bf92ef8bcab86eca099b3d85432d3e2bb7fdd412/index.js#L234)

charmingduchess:
Are you saying that the logs would be too noisy in production? They log to cloudwatch when not running locally or on Vercel.

charmingduchess:
Sierra doesn't have logs for requests, so I'm leaving the update for that client. I did opt for the plaform client internal logs.

charmingduchess:
Ok... final form is using homebaked logs and checking env for log level. The nypl data client logs are suuuper noisy, printing entire response bodies. I cut a ticket to make quieter info level logs with request urls, but for now, this will have to do.

logger.js
```diff
@@ -42,21 +42,22 @@ const initializeLogger = () => {
     }
     return JSON.stringify(logObject)
   })
-
+  console.log(process.env.LOG_LEVEL || "error")
   const transports = [
-    new winston.transports.Console({ level: "debug" }),
+    new winston.transports.Console({ level: process.env.LOG_LEVEL || "error" }),
```
nonword, line 47: All logging was debug before?

logger.js
```diff
@@ -42,21 +42,22 @@ const initializeLogger = () => {
     }
     return JSON.stringify(logObject)
   })
-
+  console.log(process.env.LOG_LEVEL || "error")
```
nonword, line 45: Remove this?

PR #490, opened 2025-03-13
Big deploy

charmingduchess:
Prod deployment of:

### Updated

- Send new searches back to home, not /search [SCC-4487](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4487)
- use new format instead of materialType for filter queries [SCC-4417](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4417)
- Removed legacy barcode hotfix (SCC-4474)
- Update NextJS version

### Added

- Added Google Tag Manager snippet and data layer initialization (SCC-4472)
- Hide non prod deployments from search indexing [SCC-4192](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4192)

[SCC-4487]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SCC-4417]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SCC-4192]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/server/nyplApiClient/index.ts
```diff
@@ -58,6 +59,16 @@ const nyplApiClient = async ({
       oauth_url: appConfig.urls.tokenUrl,
     })
     CACHE.clients[clientCacheKey] = nyplApiClient
+    const get = nyplApiClient.get.bind(nyplApiClient)
```
EdwinGuzman, line 62: This is the first time seeing this update. You can get around `bind` if you use `apply`:
```
const get = nyplApiClient.get;

nyplApiClient.get = async function (...args) {
  logger.info(`GET ...`);
  return get.apply(this.args);
}
```

It's minor but you don't need to create a function with `bind`  and can call the original function directly while overriding it.

src/server/sierraClient/index.ts
```diff
@@ -48,6 +49,21 @@ const sierraClient = async () => {
     })
 
     CACHE.client = wrapper
+    const get = wrapper.get.bind(wrapper)
```
EdwinGuzman, line 52: In this case, the bind makes more sense with the `await`. You can probably simplify this with `["get", "post", "put"].forEach(method => {...})`

EdwinGuzman (approved):
Looks good. Minor comments on some code I didn't review before but really nbd.

PR #511, opened 2025-05-06
SCC-4379: Timed logout modal responds to page activity

7emansell:
## Ticket:

- JIRA ticket [SCC-4379](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4379)

- Adds an `inactivityTimer` to the account page that's counting down from 5 minutes, and then resets the logout timer on mouse/keyboard/touch interactions

### Checklist:

### DIFF
--- a/pages/account/[[...index]].tsx
+++ b/pages/account/[[...index]].tsx
@@ -52,7 +52,14 @@ export default function MyAccount({
       }, TIMEOUT_MINUTES)
     }
 
-    const events = ["mousemove", "keydown", "mousedown", "click", "touchstart"]
+    const events = [
+      "mousemove",
+      "keydown",
+      "mousedown",
+      "click",
+      "touchstart",
+      "scroll",
+    ]
     events.forEach((event) => window.addEventListener(event, handleActivity))
 
     handleActivity()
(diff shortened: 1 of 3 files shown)


pages/account/[[...index]].tsx
```diff
@@ -37,6 +37,43 @@ export default function MyAccount({
...
+    const events = [
+      "mousemove",
+      "keydown",
+      "mousedown",
+      "click",
+      "touchstart",
+      "scroll",
+    ]
+    events.forEach((event) =>
+      window.addEventListener(event, handleActivity, { passive: true })
```
Comment (7emansell) [inline pages/account/[[...index]].tsx:64]: the `passive` option tells the browser not to wait to see if `event.preventDefault()`, like blocking a scroll, will be called. Improves performance which I'm trying to do here as there's so many event listeners :/

Comment (charmingduchess) [reply pages/account/[[...index]].tsx:64]: is there a noticeable lag?

Comment (7emansell) [reply pages/account/[[...index]].tsx:64]: No

Comment (charmingduchess) [reply pages/account/[[...index]].tsx:64]: If you are looking to trim the list, I think you can safely remove mousedown. I don't think that would happen unless someone was clicking, and this logic doesn't need to make that distinction.

charmingduchess (approved):

PR #512, opened 2025-05-08
Deploy remove links to shep to QA

charmingduchess:
- use bib subject links to create subject filter instead of SHEP urls
- subject search drop down links to subject filter search instead of SHEP


7emansell (approved):

7emansell:
adding logout modal if thats cool @charmingduchess

PR #522, opened 2025-05-30
SCC-3300: Make RC print-friendly

7emansell:
## Ticket:

- JIRA ticket [SCC-3300](https://newyorkpubliclibrary.atlassian.net/browse/SCC-3300)

- Adds `globals.css` so components at any level can apply a "no-print" style, which is essentially just not displaying on print
- Also adds listeners to `_app.tsx` to un-inject the NYPL header and footer scripts on print
- Adds "no-print" styles to most buttons, banners, menus, and the research catalog header– things that don't have any meaning or use on the printed page
- Will need to iterate with feedback from Sean and Apoorva, this is just my initial guess informed by what legacy catalog does. I'm already concerned by removing buttons that have no use but do offer information implicitly (ex. "Request scan" means a scan of this is theoretically available, etc.)

## How has this been tested?

You can go to `More tools` > `Rendering` in DevTools and force emulate the print media type, OR (easier) just `Cmd + P`. The latter is better because that's what triggers the removal of the header and footer scripts, which really impacts the overall page structure. 

- Interested if there's NYPL guidance on this, will definitely be going over this with Marty and Clare

### Checklist:

[SCC-3300]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-3300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pages/search/index.tsx
+++ b/pages/search/index.tsx
@@ -196,6 +196,7 @@ export default function Search({
               id="results-pagination"
               mt="xxl"
               mb="l"
+              className="no-print"
               initialPage={searchParams.page}
               currentPage={searchParams.page}
               pageCount={Math.ceil(totalResults / RESULTS_PER_PAGE)}


charmingduchess (commented):
This  is a good start, and I think you've nailed the approach for removing components from print. A couple questions that came to my mind looking at the print preview: 
Why is someone printing out the search results? What is the actual information they need captured? And what is currently problematic about the default UI that gets in the way of it being useful to print as is? 

A couple notes:
- I don't think it makes sense to include status, since that information is not useful when it is not dynamic. 
- We def need to make sure that information regarding special collections access is present, but a link to the finding aid may not be useful to someone printing a page.
- In my mind, the biggest problem we need to address with these styles is condensing the space each bib takes up. Right now, each one is 1/3-1/2 a full page. Users familiar with the export list functionality are used to a text document with zero styling, and cmd+p on a legacy search has 6 or more records per page. I think we can be pretty liberal with removing styles for the print view, in particular removing unneeded spacing. 
  - Potentially we could also reduce the size of the heading text... but I feel like that actually is good to keep extra large.

src/components/Layout/Layout.tsx
```diff
@@ -70,9 +70,12 @@ const Layout = ({
                       outlineColor: "ui.white",
                     },
                   },
+                  "@media print": {
```
charmingduchess, line 73: does adding the no-print class name not work for this component?

charmingduchess: Can you add it as a prop on the breadcrumbs component?

charmingduchess (approved):

PR #523, opened 2025-06-04
SCC-4751: 404 reworking

7emansell:
## Ticket:

- JIRA ticket [SCC-4751](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4751)

- Removes the 404 redirect for bib/hold pages and replaces it with just returning the 404 page component, not changing the URL
- Updates 404 page to include the `Layout`

** GA tracking of 404s can continue as before since the metadata still updates to page title = 404

## How has this been tested?

Try nonexistent bibs or paths.

### Checklist:

[SCC-4751]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pages/hold/request/[id]/index.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react-hooks/rules-of-hooks */
```
Comment (7emansell) [inline pages/hold/request/[id]/index.tsx:1]: This is necessary bc the hooks in this component are now conditional on `notFound = false`. This is still safe though imo

__test__/pages/404/404.test.tsx
```diff
@@ -1,33 +1,24 @@
...
 describe("404", () => {
   it("should display 404 text", () => {
-    render(<Custom404 />)
+    render(<Custom404 activePage="account" />)
+    const container = screen.getByRole("main")
 
-    const FourOhFourText = "404 Not Found"
-    const heading = screen.getByRole("heading")
-    expect(heading).toHaveTextContent(FourOhFourText)
+    const fourOhFourText = "404 Not Found"
```
charmingduchess, line 11: if you export this string from `pages/404/index.tsx` then it will be even less likely to break when updated later

charmingduchess (changes requested):
This is an elegant solution! 
The blanket disabling of react hook rules checking makes me a little nervous. Can you disable it for those lines only?
Just out of curiosity, you mention that the metadata title is remaining '404', so it should not impact GA reporting. Where does that metadata title get passed to GA?

7emansell:
> This is an elegant solution! The blanket disabling of react hook rules checking makes me a little nervous. Can you disable it for those lines only? Just out of curiosity, you mention that the metadata title is remaining '404', so it should not impact GA reporting. Where does that metadata title get passed to GA?

It is annoying but otherwise there's 9+ places per file I need to disable the `eslint` rule. If it makes you feel more comfortable, it's really not breaking the rules of React hooks, just how `eslint` enforces them. 

As far as I understand, the metadata title of any page is passed to GA when the page is visited.

7emansell:
VQA cleared (RC header now included in 404 view, no other visual updates)

pages/bib/[id]/index.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react-hooks/rules-of-hooks */
```
Comment (charmingduchess) [inline pages/bib/[id]/index.tsx:1]: ahh idk about this. I think the implementation falls pretty squarely into 
> 🔴 Do not call Hooks after a conditional return statement.
from [react docs](https://react.dev/reference/rules/rules-of-hooks)

Although this case doesn't seem super risky, I think it makes sense to refactor to match the react recommendation

charmingduchess (approved):
Great! I like how reusable the `tryInstantiate` method is.

src/utils/appUtils.ts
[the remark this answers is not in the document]
7emansell: Yup first time using the constructor type as the return type though which i thought was cool

PR #538, opened 2025-06-18
SCC-3792: Rework Discovery API search error handling

7emansell:
## Ticket:

- JIRA ticket [SCC-3792](https://newyorkpubliclibrary.atlassian.net/browse/SCC-3792)

- Makes the 404 page prettier according to [these designs](https://www.figma.com/design/K09KeGAVEXpleym34nTlue/DRB---Error-Pages?node-id=4487-1497&p=f&m=dev)
- Adds error branches to the search API handler based on [these expected errors from Discovery API
](https://github.com/NYPL/discovery-api/blob/3203997ac59eb2cabc86046a96db25ad7bc9fac7/routes/resources.js#L24)
- Adds `SearchError` to render adjusted versions of these three error categories on the search page
- Keeps the 404 work the same on bib/hold pages– their errors are much more covered and have specific warnings to render
- ** Important note that this work (and what was done [here](https://github.com/NYPL/research-catalog/pull/523)) departs from the standard Next error flow, which would have us redirect to a new error page entirely. Instead, we're conditionally displaying a message within the existing page 

## How has this been tested?

Throw errors from `src/server/api/search` or look at these handy screenshots
Screenshot 2025-06-18 at 3 22 07 PM
Screenshot 2025-06-18 at 3 21 48 PM
Screenshot 2025-06-18 at 3 21 22 PM

### Checklist:

[SCC-3792]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-3792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/searchUtils.ts
```diff
@@ -335,13 +335,13 @@ export function mapQueryToSearchParams({
  * Otherwise returns `null`
  */
 export function checkForRedirectOnMatch(
-  results: SearchResultsResponse | Error,
+  results: SearchResultsResponse,
   query
 ): object {
-  const hasOneResult =
-    !(results instanceof Error) && results?.results?.totalResults === 1
+  const hasOneResult = results?.results?.totalResults === 1
```
7emansell, line 341: removing this error handling based on the assumption that now, if you have results being passed here, they should be valid results

src/server/api/search.ts
```diff
@@ -48,27 +48,66 @@ export async function fetchResults(
...
+              ? resultsResponse.reason.message
+              : resultsResponse.reason,
+        }
+      }
+      if (aggregationsResponse.status === "rejected") {
+        logServerError("fetchResults", aggregationsResponse.reason)
+        return {
+          status: 500,
+          message:
+            aggregationsResponse.reason instanceof Error
```
7emansell, line 80: necessary to strip the string out of the `Error`

src/server/api/search.ts
```diff
@@ -48,27 +48,66 @@ export async function fetchResults(
...
+    }
+
+    // Assign results values for each response
+    const results = resultsResponse.value
+
+    const aggregations = aggregationsResponse.value
+
+    // Handle invalid parameter rejection or empty results (422, 404)
+    if (
+      results.status === 422 ||
```
charmingduchess, line 94: are these actual error codes that are returned from the discovery api?

charmingduchess: If we are getting 4xx errors from discovery api, that indicates malformed queries made by RC. let's log these so we can keep tabs if these errors ever do turn up.

charmingduchess (changes requested):

src/server/api/search.ts
```diff
@@ -48,27 +48,66 @@ export async function fetchResults(
...
-    resultsResponse.status === "fulfilled" && resultsResponse.value
+    const [resultsResponse, aggregationsResponse] = await Promise.allSettled([
+      client.get(`${DISCOVERY_API_SEARCH_ROUTE}${resultsQuery}`),
+      client.get(`${DISCOVERY_API_SEARCH_ROUTE}${aggregationQuery}`),
+    ])
 
-  const aggregations =
-    aggregationsResponse.status === "fulfilled" && aggregationsResponse.value
+    // Handle failed promises (500)
+    if (
```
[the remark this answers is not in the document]
7emansell: No it is not, removed

pages/search/index.tsx
```diff
@@ -109,9 +104,25 @@ export default function Search({
...
 
+  if (errorStatus) {
+    return <SearchError errorStatus={errorStatus} />
+  }
+
+  const { itemListElement: searchResultsElements, totalResults } =
+    results.results
+
+  const aggs = results?.aggregations?.itemListElement
+  // if there are no results, then applied filters correspond to aggregations
```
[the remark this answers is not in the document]
7emansell: Removing this check since yes, it should be caught in `fetchResults`

src/server/api/search.ts
```diff
@@ -95,13 +90,15 @@ export async function fetchResults(
       results.status === 404 ||
       !(results?.totalResults > 0)
     ) {
+      logServerError("fetchResults", results.message ?? "No results found")
```
charmingduchess, line 93: can you log the request url here too?

charmingduchess (approved):
looks good! one small change plz

src/components/SearchResults/SearchError.tsx
```diff
@@ -0,0 +1,93 @@
...
+            </p>
+          </>
+        )
+      case 500:
+        return (
+          <>
+            <Heading level="h3" tabIndex={-1} id="search-results-heading">
+              Something went wrong on our end.
+            </Heading>
+            <Text marginBottom="0">
```
oliviawongnyc, line 41: FYI, `Text`'s `noSpace` prop does the same as `marginBottom="0"`, but this is not a blocker.

src/components/SearchResults/SearchError.tsx
```diff
@@ -0,0 +1,93 @@
...
+          flexDir="column"
+          marginTop="l"
+          marginBottom="xxl"
+          marginLeft="l"
+          marginRight="l"
+          alignItems="center"
+          justifyContent="center"
+          textAlign="center"
+        >
+          {errorContent(errorStatus)}
```
oliviawongnyc, line 88: Instead of an `errorContent` function that needs to be created earlier in the component and then called later, you could create a variable `let errorContent` and then add your switch statement so that the value is already determined by the time it gets to the return statement. This is not a blocker, and likely does not make a big difference in this component, but is generally a better practice to reduce unnecessary work.

```
  let errorContent

  switch (errorStatus) {
    case 404:
      errorContent = (
```

src/assets/errorImage.png
oliviawongnyc, line 1: Is this image being used? I am not seeing it in the screenshots you included.

src/utils/searchUtils.ts
```diff
@@ -334,12 +334,9 @@ export function mapQueryToSearchParams({
...
  * Otherwise returns `null`
  */
-export function checkForRedirectOnMatch(
-  results: SearchResultsResponse | Error,
-  query
-): object {
-  const hasOneResult =
-    !(results instanceof Error) && results?.results?.totalResults === 1
+export function checkForRedirectOnMatch(results, query): object {
+  console.log
```
oliviawongnyc, line 338: Stray console.log?

oliviawongnyc (approved):
Edwin and I took a look at this together. Couple non-blocking suggestions!

EdwinGuzman (approved):

PR #545, opened 2025-07-02
SCC-4544: New and cuter Subnav

7emansell:
## Ticket:

- JIRA ticket [SCC-4544](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4544)

- Updates the `SubNav` to use DS `SubNav` component, which is pretty much just a visual update (see [designs](https://www.figma.com/design/8mW2O0U7wyDdkg1JAD6Lht/Enhanced-Browse))
  - Links all go to the same place as before, except Subject Heading Explorer is now Browse the Catalog, in anticipation of EB

## How has this been tested?

Confirm that logging in and out changes visibility of log out link in the subnav **this can't be tested until Sierra is back up

### Checklist:

[SCC-4544]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Adding "Discovery" to `SearchResults`, `AggregationResults`, and `SearchResultsElement` to parallel `DiscoveryBibResult` and indicate the API structure [SCC-4148](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4148)
 - Replace subnav with DS `SubNav` [SCC-4544](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4544)
+- Remove duplicate SearchResultSort component between mobile and desktop [SCC-4086](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4086)
 
 ### Fixed
 
--- a/__test__/pages/search/searchResults.test.tsx
+++ b/__test__/pages/search/searchResults.test.tsx
@@ -51,22 +51,13 @@ describe("Search Results page", () => {
           results={{ results, status: 200 }}
         />
       )
-      const mobileSortBy = screen.getAllByLabelText("Sort by")[0]
-      expect(mobileSortBy).toHaveValue("relevance")
-      await userEvent.selectOptions(mobileSortBy, "Title (A - Z)")
-      expect(mobileSortBy).toHaveValue("title_asc")
+      const sortBy = screen.getAllByLabelText("Sort by")[0]
+      expect(sortBy).toHaveValue("relevance")
+      await userEvent.selectOptions(sortBy, "Title (A - Z)")
+      expect(sortBy).toHaveValue("title_asc")
 
       expect(mockRouter.asPath).toBe(
         "/?q=spaghetti&sort=title&sort_direction=asc"
-      )
-
-      const desktopSortBy = screen.getAllByLabelText("Sort by")[1]
-      expect(desktopSortBy).toHaveValue("title_asc")
-      await userEvent.selectOptions(desktopSortBy, "Title (Z - A)")
-      expect(desktopSortBy).toHaveValue("title_desc")
-
-      expect(mockRouter.asPath).toBe(
-        "/?q=spaghetti&sort=title&sort_direction=desc"
       )
     })
     it("returns the user to the first page on sorting changes", async () => {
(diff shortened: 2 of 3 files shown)


src/components/RCSubNav/RCSubNav.tsx
```diff
@@ -0,0 +1,75 @@
...
+            isSelected={activePage === "shep"}
+            aria-current={activePage === "shep" ? "page" : undefined}
+          >
+            Browse the Catalog
+          </SubNavLink>
+        </>
+      }
+      secondaryActions={
+        <>
+          <div style={{ display: isAuthenticated ? "flex" : "none" }}>
```
7emansell, line 49: annoying but the best way I can think of to get around what the `SubNav` is doing here, which is checking valid React elements and list-ifying them

src/components/RCSubNav/RCSubNav.tsx
```diff
@@ -0,0 +1,75 @@
...
+          >
+            Search the Catalog
+          </SubNavLink>
+          <SubNavLink
+            href="/research/research-catalog/subject_headings"
+            id="subnav-browse"
+            isSelected={activePage === "shep"}
+            aria-current={activePage === "shep" ? "page" : undefined}
+          >
+            Browse the Catalog
```
charmingduchess, line 43: This should say Subject Heading Explorer until we release the new EB

src/components/RCSubNav/RCSubNav.tsx
charmingduchess: hmm this might warrant a DS update. `null` is a valid return value from a React component, so the Subnav should be able to gracefully ignore it

charmingduchess: @EdwinGuzman what do you think?

charmingduchess (changes requested):

src/components/RCSubNav/RCSubNav.tsx
charmingduchess: it looks like it would just entail adding conditional chaining to [this line](https://github.com/NYPL/nypl-design-system/blob/cecfed5f9f1a31cb9b0be33023e2f7f5f4a186a7/src/components/SubNav/SubNav.tsx#L212). It does actually seem like a bug, since there is an error being thrown before an appropriate error is logged

CHANGELOG
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Updated
 
 - Adding "Discovery" to `SearchResults`, `AggregationResults`, and `SearchResultsElement` to parallel `DiscoveryBibResult` and indicate the API structure [SCC-4148](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4148)
+- Replace subnav with DS `SubNav` [SCC-4544](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4544)
```
EdwinGuzman, line 13: nit: add "cuter" before `DS SubNav`

src/components/RCSubNav/RCSubNav.tsx
```diff
@@ -0,0 +1,75 @@
+import { type RCPage } from "../../types/pageTypes"
```
EdwinGuzman, line 1: I think this is invalid syntax... it should be `import type { RCPage } ...`.

src/components/RCSubNav/RCSubNav.tsx
EdwinGuzman: The validation that the `SubNav` is trying to do is to verify that every child passed is either `SubNavLink` or `SubNavButton`. If it's not, it's going to log a warning but should not cause other issues.

But I think I understand what the goal is and why the `null` issue was brought up. Is it correct that the "log out" link is dynamic, so the conditional null value is what's causing the issue?

@7emansell this is something that should be worked on now before v4. I would prefer if we entirely remove the log message and check for valid components, and just let devs do what they need to do. I'll create a ticket and assing to you if that's ok?

EdwinGuzman (commented):
Nice to finally see the subnav!

7emansell:
> But I think I understand what the goal is and why the null issue was brought up. Is it correct that the "log out" link is dynamic, so the conditional null value is what's causing the issue?

Yes, I think so

> @7emansell this is something that should be worked on now before v4. I would prefer if we entirely remove the log message and check for valid components, and just let devs do what they need to do. I'll create a ticket and assing to you if that's ok?

Yup sounds good

EdwinGuzman:
I'll create the ticket. For now, the workaround is okay.

charmingduchess (approved):

PR #559, opened 2025-07-28
Scc 4792/playwright homepage

chrismulholland:
## Ticket: 4792

- JIRA ticket [https://newyorkpubliclibrary.atlassian.net/browse/SCC-{4792}]

- Adds Playwright tests for the RC home page.  The tests verify that all homepage elements that are supposed to be there, are indeed there.  Header tests are not exhaustive, we're only testing for presence of elements now.  Footer tests are absent.  More complete header and footer tests will come at a later date.  Page object modeling has been used for the tests and related page objects have been created.  They reside in the pages folder.

## How has this been tested?

verified locally.

-

### Checklist:

- [ x] I updated the CHANGELOG with the appropriate information and JIRA ticket number (if applicable).
- [ x] All new and existing tests passed.


playwright/tests/100_homepage.spec.ts
```diff
@@ -0,0 +1,131 @@
+import { test, expect } from "@playwright/test"
```
7emansell, line 1: why is this file named `100_homepage` rather than maybe just `homepage`?

playwright/tests/100_homepage.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+import { RC_Home_Page } from "../pages/rc_home_page"
+
+test("Verify elements on the Research Catalog home page", async ({ page }) => {
```
7emansell, line 4: could more of these tests be grouped? For example, `test("Verify header appears on the home page")` and then the tests that relate to the header run as a group. This will make it so we don't need so many console.logs and it'll be easier to isolate issues. I'm thinking something along [these lines](https://github.com/NYPL/digital-collections/blob/main/playwright/tests/about.spec.ts)

CHANGELOG
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Prerelease
 
+- Added Playwright tests for RC home page elements. found in /playwright. Deleted dummy test file that was created for initial Playwright install. [SCC-4792] (https://newyorkpubliclibrary.atlassian.net/browse/SCC-4792)
```
7emansell, line 10: great but can you move this under the `## Added` heading

playwright/pages/rc_home_page.ts
```diff
@@ -0,0 +1,108 @@
...
+    this.search_subtmit_button = page.getByRole("button", {
+      name: "Search",
+      exact: true,
+    })
+    this.advanced_search_link = page.getByRole("link", {
+      name: "Advanced Search",
+    })
+    this.info_banner = page.getByText("Banner informative iconNew!")
+    this.body_main_heading = page.getByRole("heading", {
+      name: "Explore the Library's Vast",
```
7emansell, line 52: Just curious, any reason why this isn't the full heading text ?

7emansell (changes requested):
Looks great overall just a couple tweaks and clarifications on naming.
Also, this is majorly nitpicky but may be helpful down the road– Github uses Markdown as its formatting language, so for the PR description, this is how you can [format links](https://www.markdownguide.org/basic-syntax/#links) to make it more readable.

playwright/tests/100_homepage.spec.ts
chrismulholland: I like to prefix test file names with numbers so they appear grouped when in a list.  for instance, tests in the "100s" might be homepage, header, footer.  200 tests might be related to account, 300 for search, etc.

playwright/tests/100_homepage.spec.ts
chrismulholland: yes, for sure.  I should have broken out the header assertions into their own test.  I will do that.
I use console.log before each assertion so those statements appear in the console when tests run.  I do this so I can easily see what is being tested.  Something of a guide to use for manual testing, and replaces writing out the manual steps we used to keep in browserstack test mangement.

CHANGELOG
chrismulholland: done.

playwright/pages/rc_home_page.ts
chrismulholland: good question.  I will revisit this.

playwright/tests/100_homepage.spec.ts
chrismulholland: I broke out the global header assertions into their own test and removed the console.log for those assertions.

clarissarichard: Building upon Emma's suggestion, I found it helpful to group similar tests using `test.describe()` blocks as described in the Playwright docs [here](https://playwright.dev/docs/api/class-test#test-describe). For manual testing purposes, these blocks could serve as the test case name, while the tests inside the block could serve as the test steps. Just a thought!

playwright/pages/base_page.ts
```diff
@@ -0,0 +1,93 @@
...
+      .getByTestId("layout-breadcrumbs")
+      .getByRole("link", { name: "Research", exact: true })
+    this.research_catalog_breadcrumb = page
+      .getByTestId("layout-breadcrumbs")
+      .getByRole("link", { name: "Research Catalog" })
+    // header elements below breadcrumbs
+    this.h1 = page.getByRole("heading", {
+      level: 1,
+      name: "Research Catalog",
+    })
```
clarissarichard, line 84: I'd be wary of using potentially reserved names like `h1` and recommend renaming this locator to something like `researchCatalogHeading` so that it is also easier to read later in the tests.

clarissarichard (commented):
Left a few suggestions and clarifying questions!

playwright/pages/rc_home_page.ts
7emansell: Don't think this change has been pushed up.

playwright/tests/100_homepage.spec.ts
7emansell: Gotcha. this isn't a pattern we use in this repo and not documented so it's unclear what the numbers refer to– could we just use the page name, following the convention of the folder names in `/pages` (homepage can be an exception to this since it's just `pages/index`)

chrismulholland: I will update the folder structure to this:
tests/Home/homepage.spec.ts

playwright/tests/100_homepage.spec.ts
chrismulholland: describe blocks have been added, and all console.log statements have been removed.

playwright/tests/100_homepage.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+  console.log("Checking if the header 'Get Help' is visible...")
+  await expect(rcHomePage.header_get_help).toBeVisible()
+  console.log("Checking if the header 'Search' is visible...")
+  await expect(rcHomePage.header_search).toBeVisible()
+
+  console.log("Checking if the header 'My Account' is visible...")
+  await expect(rcHomePage.header_my_account).toBeVisible()
+
+  console.log("Checking if heading is visible...")
+  await expect(rcHomePage.isHeadingVisible()).resolves.toBe(true)
```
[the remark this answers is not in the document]
chrismulholland: outdated and has been removed.

playwright/pages/base_page.ts
```diff
@@ -0,0 +1,93 @@
...
+      .getByRole("link", { name: "Give" })
+    this.header_get_help = page
+      .getByLabel("Header bottom links")
+      .getByRole("link", { name: "Get Help" })
+    this.header_search = page.locator("#searchButton")
+    // breadcrumbs
+    this.home_breadcrumb = page.getByRole("link", { name: "Home" })
+    this.research_breadcrumb = page
+      .getByTestId("layout-breadcrumbs")
+      .getByRole("link", { name: "Research", exact: true })
```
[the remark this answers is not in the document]
chrismulholland: fixed.

playwright/pages/rc_home_page.ts
```diff
@@ -0,0 +1,105 @@
...
+    // Footer
+    this.footer_container = page.locator("footer#footer")
+    this.help_and_feedback = page.getByRole("button", {
+      name: "Help and Feedback",
+    })
+  }
+
+  async goto() {
+    await this.page.goto("/research/research-catalog")
+  }
```
clarissarichard, line 104: You won't need this function if `baseURL` is set in `playwright.config.ts`, but this is something you can address in your next PR!

clarissarichard (approved):
Looks great!

PR #561, opened 2025-07-31
SCC-4568: Browse form and more index page

7emansell:
## Ticket:

- JIRA ticket [SCC-4568](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4568)

- Creates the `BrowseForm` component
- Makes `SearchError` reusable as the error for browse results --> `ResultsError`
- Adds the "empty" state to the browse index page- if no keywords
- Updates `Layout` and `RCSubnav`
- Updates browse types and API handler with new endpoint changes (`totalResults`)

## How has this been tested?

Test some queries! Browse page should now do everything, except reflect the sort and pagination in the UI 

- Focus from empty state not picking up once there's a heading on browse page but that's same as search, just a little weirder since the display doesn't change as much. Noting to flag for Clare for sure

### Checklist:

[SCC-4568]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/BrowseForm/BrowseForm.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const component = <BrowseForm />
+
+  it("submits a browse term by default", async () => {
+    render(component)
+    const input = screen.getByRole("textbox")
+
+    await userEvent.type(input, "Ornithology")
+    submit()
+    expect(mockRouter.asPath).toBe(
+      "/browse?q=Ornithology&search_scope=has&sort=preferredTerm&sort_direction=asc"
```
7emansell, line 34: Will revisit these tests once sort and pagination are in the mix

pages/browse/index.tsx
```diff
@@ -60,25 +149,35 @@ export async function getServerSideProps({ req, query }) {
   const patronTokenResponse = await initializePatronTokenAuth(req.cookies)
   const { browseType = "subjects" } = query
 
+  const browseParams = mapQueryToBrowseParams(query)
+  const isAuthenticated = patronTokenResponse.isTokenValid
+
+  // Skip request if no keywords
+  if (!browseParams.q || browseParams.q.length === 0) {
```
7emansell, line 156: recent update from Apoorva's end, need to run by Sean if we explicitly want to prevent users from ever seeing the unqueried index

charmingduchess (changes requested):
🍝

charmingduchess (changes requested):
I think this update makes sense! Thanks for incorporating the changes. One suggestion and then I think we're good to go.

PR #573, opened 2025-08-15
EB 1.0 VQA/QA 

7emansell:
QA and VQA fixes on browse index page.
Notably:
- switching to `Menu` instead of `Select` component, and updating tests to correspond
- pulling banner components out of `Layout` and adding more responsive styling
- Copy fixes

### DIFF
--- a/pages/_error.tsx
+++ b/pages/_error.tsx
@@ -36,7 +36,9 @@ function Error({ activePage }: ErrorPageProps) {
             height={68}
             style={{ marginBottom: "48px" }}
           />
-          <Heading level="h3">Something went wrong on our end</Heading>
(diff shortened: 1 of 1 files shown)


src/components/ResultsError/ResultsError.tsx
```diff
@@ -27,7 +27,7 @@ export default function ResultsError({ errorStatus, page }: ResultsErrorProps) {
     case 404:
       errorContent = (
         <>
-          <Heading level="h3" tabIndex={-1} id={headingID}>
+          <Heading level="h2" size="heading3" tabIndex={-1} id={headingID}>
```
charmingduchess, line 30: is this a styling update or was there a callout for bad heading numbers?

PR #575, opened 2025-08-20
SCC-4815: Browse bib results page

7emansell:
## Ticket:

- JIRA ticket [xxx](https://newyorkpubliclibrary.atlassian.net/browse/SCC-{fill in the ticket number})

-

-

### Checklist:

### DIFF
--- a/pages/browse/subjects/[...slug].tsx
+++ b/pages/browse/subjects/[...slug].tsx
@@ -1,19 +1,87 @@
-import RCHead from "../../../src/components/Head/RCHead"
-import Layout from "../../../src/components/Layout/Layout"
+import { useRouter } from "next/router"
 import { SITE_NAME } from "../../../src/config/constants"
+import { fetchResults } from "../../../src/server/api/search"
+import initializePatronTokenAuth from "../../../src/server/auth"
+import {
+  mapQueryToSearchParams,
+  checkForRedirectOnMatch,
+} from "../../../src/utils/searchUtils"
+import Search from "../../search"
+import type { SearchResultsResponse } from "../../../src/types/searchTypes"
+import type { HTTPStatusCode } from "../../../src/types/appTypes"
+import type { RCPage } from "../../../src/types/pageTypes"
+
+interface SubjectSearchProps {
+  bannerNotification?: string
+  results: SearchResultsResponse
+  isAuthenticated: boolean
+  errorStatus?: HTTPStatusCode | null
+  activePage?: RCPage
+  metadataTitle?: string
+}
 
 /**
  * The Browse subject headings bib results page is responsible for fetching and displaying bib results
  * filtered by at least one subject heading, as well as displaying and controlling pagination, sort,
  * and other filters.
  */
-export default function SubjectHeadingResults() {
-  const metadataTitle = `Subject Heading Results | ${SITE_NAME}`
+export default function SubjectHeadingResults({
+  bannerNotification,
+  results,
+  isAuthenticated,
+  errorStatus = null,
+  activePage = "browse",
+  metadataTitle,
+}: SubjectSearchProps) {
+  const router = useRouter()
+  const { subject } = router.query
 
-  return (
-    <>
-      <RCHead metadataTitle={metadataTitle} />
-      <Layout activePage="sh-results"></Layout>
-    </>
-  )
+  // Merge the fixed subject into the query
+  const modifiedQuery = {
+    ...router.query,
+    subject: Array.isArray(subject) ? subject[0] : subject, // Ensure single value
+  }
+
+  const modifiedProps = {
+    bannerNotification,
+    results,
+    isAuthenticated,
+    errorStatus,
+    metadataTitle,
+    activePage,
+    initialQuery: modifiedQuery,
+  }
+  console.log(activePage)
+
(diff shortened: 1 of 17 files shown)


src/components/SearchBrowseForm/SearchBrowseForm.tsx
```diff
@@ -44,11 +48,19 @@ const SearchBrowseForm = ({
...
   onSubmitFocusId = null,
   children,
+  activePage,
 }: SearchBrowseFormProps) => {
   const router = useRouter()
   const isLoading = useLoading()
-  const { setPersistentFocus } = useFocusContext()
+  const [showBackButton, setShowBackButton] = useState(false)
 
+  useEffect(() => {
```
7emansell, line 57: Not using session storage bc it ended up not solving the flash problem– it's still async on the client side. The only server side way to do this is by putting a param in the URL which actually does seem like overkill. 
Ultimately is the flash too annoying or is it worth it

src/components/SearchBrowseForm/SearchBrowseForm.tsx
```diff
@@ -44,11 +48,22 @@ const SearchBrowseForm = ({
...
   children,
+  activePage,
 }: SearchBrowseFormProps) => {
   const router = useRouter()
   const isLoading = useLoading()
-  const { setPersistentFocus } = useFocusContext()
 
+  const [backUrl, setBackUrl] = useState<string | null>(null)
+
+  useEffect(() => {
```
7emansell, line 58: Nvm about session storage it still doesn't fix the flash. How annoying is the flash anyways maybe i'm overreacting

pages/browse/subjects/[...slug].tsx
```diff
@@ -1,19 +1,123 @@
...
+
+  if (results.status !== 200) {
+    return { props: { errorStatus: results.status } }
+  }
+
+  const redirect = checkForRedirectOnMatch(results, query)
+  if (redirect) return { redirect }
+
+  return {
+    props: {
```
Comment (charmingduchess) [inline pages/browse/subjects/[...slug].tsx:115]: if you coerce `slug` into an array and pass as a prop would that allow you to remove the multiple `isArray` checks on the component and in `buildSubjectQuery`?

Comment (charmingduchess) [reply pages/browse/subjects/[...slug].tsx:115]: I also think slug might not be necessary in the query params. You should be able to rely on its presence in the URL. You can include it in the `pathname` that you are pushing on page and sort changes. it's hard to tell with the ES index acting so erratically, but I'm pretty sure it doesn't need to be in both places (query and url param)

charmingduchess (changes requested):
Sorry for super chatty review. I left a lot of recommendations for some refactoring, but I'm actually most interested in figuring out if we can remove `slug` from the query params in general (I think it's possible).

pages/browse/subjects/[...slug].tsx
```diff
@@ -1,19 +1,123 @@
...
+      bannerNotification,
+      results,
+      isAuthenticated: patronTokenResponse.isTokenValid,
+      metadataTitle: `Subject Heading Results | ${SITE_NAME}`,
```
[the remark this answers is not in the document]
Comment (7emansell) [reply pages/browse/subjects/[...slug].tsx:119]: checking with seanpoorva rn

7emansell:
I think i implemented every suggested change from above, lmk if these are what you were thinking. To me already so so much cleaner now, thank you @charmingduchess . Also readier to expand to authors/genres for sure

src/utils/utilsTests/searchUtils.test.ts
```diff
@@ -250,5 +250,20 @@ describe("searchUtils", () => {
         expect(heading).toEqual('Displaying 3 of 3 results for LCCN "1234"')
       })
     })
+    describe("browse result searches", () => {
+      it("returns heading with browseOptions when slug and browseType are provided", () => {
+        const heading = getSearchResultsHeading(
```
charmingduchess, line 255: is this a valid case? where there would be an advanced search query param along side a subject slug?

src/utils/browseUtils.ts
```diff
@@ -156,3 +160,36 @@ export function buildSubjectLinks(
...
+  filter: string
+}) {
+  const collapsedFilters = collapseMultiValueQueryParams(query)
+
+  // Merge the locked slug in, it should be first and unique in its filter array
+  const merged = {
+    ...collapsedFilters,
+    [filter]: [
+      slug,
+      ...(collapsedFilters[filter] ?? []).filter((f) => f !== slug),
```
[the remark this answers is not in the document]
7emansell: I think this is not necessary exactly– won't mess up the results, but just trying to clean up the query– but can remove if the idea is that we want to intervene as little as possible once the user is messing with the url

pages/browse/subjects/[slug].tsx
```diff
@@ -107,7 +107,7 @@ export async function getServerSideProps({ req, query, params }) {
       bannerNotification,
       results,
       isAuthenticated: patronTokenResponse.isTokenValid,
-      metadataTitle: `Subject Heading Results | ${SITE_NAME}`,
+      metadataTitle: `Search | ${SITE_NAME}`,
```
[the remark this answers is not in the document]
Comment (7emansell) [reply pages/browse/subjects/[slug].tsx:110]: Yupppppp yup

PR #593, opened 2025-09-25
SCC-4881: Custom multiselect and some advanced search form updates

7emansell:
## Ticket:

- JIRA ticket [SCC-4881](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4881)

- Duplicates the DS `Multiselect` into a `GroupedMultiselect` that allows titles within the accordion panel
- Rebuilds advanced search with `Multiselect`s and the new vocabularies instead of the `CheckboxGroups` we had before (somewhat out of scope for this ticket but is the proof that the `GroupedMultiselect` can be integrated)
- Rewrites tests for new selection behavior

Note for reviewing: Adding the `Multiselect`s seemed to really slow down keyboard events when I was working on this locally (unfortunately makes sense since there's so much rerendering) but tried a bunch of variations with state and memoization, plus getting rid of the reducer, and nothing really improved it. And then it didn't seem very slow on Vercel so I kind of just left it. Lmk if it seems slow though

### Checklist:

[SCC-4881]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pages/search/advanced.tsx
```diff
@@ -146,6 +139,44 @@ export default function AdvancedSearch({
...
+    { value: "format", label: "Format", options: formatOptions },
+    {
+      value: "buildingLocation",
+      label: "Item location",
+      options: buildingLocationOptions,
+    },
+    { value: "language", label: "Language", options: languageOptions },
+  ]
+  const multiselects = fields.map((field) => {
+    if (searchFormState["filters"][field.value]) {
```
charmingduchess, line 153: why is this dependent on the field being in the state?

src/components/AdvancedSearch/GroupedMultiselect/GroupedMultiselect.tsx
charmingduchess, line 1: could you leave some annotations in this file indicating where it departs from the DS port?

charmingduchess: Also, `GroupedMultiSelect` implies to me that this would be multiselects in a group. Maybe we call this `MultiSelectWithGroupTitles` or something?

pages/search/advanced.tsx
```diff
@@ -168,63 +199,59 @@ export default function AdvancedSearch({
...
-                  handleCheckboxChange("buildingLocation", e)
+            <Flex
+              direction="column"
+              gap={{ base: "m", md: "43px" }}
+              grow="1"
+              mt="m"
+              width={{ base: "100%", md: "50%" }}
+            >
+              {multiselects}
+              <GroupedMultiSelect
```
charmingduchess, line 236: This is a nit, but i think it might make sense to include the `groupedMultiSelect` in the multiselects array.

charmingduchess (approved):
Some minor changes suggested, but looks great

PR #597, opened 2025-10-01
SCC 4855/playwright call number search

chrismulholland:
## Ticket:

- JIRA ticket [4855](https://newyorkpubliclibrary.atlassian.net/browse/SCC-4855)

- Adds Playwright test for call number search. 

## How has this been tested?

all tests pass locally

-

### Checklist:

### DIFF
--- a/playwright/pages/search_page.ts
+++ b/playwright/pages/search_page.ts
@@ -38,9 +38,6 @@ export class SearchPage {
     this.searchResultsTitle = page
       .locator("#search-results-list")
       .getByRole("link", { name: new RegExp(this.searchterm) })
-    this.searchResultsCallNumber = page.locator(
-      '#search-results-list table td:has-text("CALL NUMBER") td p span'
-    )
   }
 
   async searchFor(searchterm: string, searchType = "Keyword") {
--- a/playwright/pages/search_page.ts
+++ b/playwright/pages/search_page.ts
@@ -9,6 +9,7 @@ export class SearchPage {
   readonly searchterm: string
   readonly search_dropdown: Locator
   readonly searchType: string
+  readonly searchResults: Locator
   readonly searchResultsTitle: Locator
   readonly searchResultsCallNumber: Locator
 
@@ -33,7 +34,8 @@ export class SearchPage {
         "i"
       ),
     })
-    this.searchResult = page
+    this.searchResults = page.locator("#search-results-list h3 a")
+    this.searchResultsTitle = page
       .locator("#search-results-list")
       .getByRole("link", { name: new RegExp(this.searchterm) })
     this.searchResultsCallNumber = page.locator(
--- a/playwright/pages/search_page.ts
+++ b/playwright/pages/search_page.ts
@@ -10,6 +10,7 @@ export class SearchPage {
   readonly search_dropdown: Locator
   readonly searchType: string
   readonly searchResultsTitle: Locator
+  readonly searchResultsCallNumber: Locator
 
   constructor(page: Page, searchterm: string, searchType = "Keyword") {
     this.page = page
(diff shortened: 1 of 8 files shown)


playwright/tests/Search/search_title.spec.ts
```diff
@@ -10,8 +10,12 @@ test.beforeEach(async ({ page }) => {
 })
 
 test.describe("Title Search", () => {
-  test("Do a title search and assert that at least 10 returned titles contain the supplied keyword", async () => {
+  test("Do a title search and assert that at least 10 returned titles contain the supplied keyword", async ({
+    page,
+  }) => {
     await searchPage.searchFor(searchterm, "Title")
+    // sleep for 5 seconds to allow results to load
+    // await page.waitForTimeout(5000)
```
7emansell, line 18: timeout no longer necessary? When I run these, and the other ones that are checking author and title, playwright categorizes them as flaky since they usually timeout on the first run

playwright/tests/Search/search_call_number.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+})
+
+test.describe("Call Number Search", () => {
+  test("Do a call number search and assert that at the item returned contains the supplied call number", async ({
+    page,
+  }) => {
+    await searchPage.searchFor(searchterm, "Call number")
+    await expect(searchPage.resultsHeading).toBeVisible()
+    await expect(
+      searchPage.page.getByTestId("card-content").getByText("JFD 93-1962")
```
7emansell, line 19: is it possible to not use a test id here?

playwright/pages/search_page.ts
```diff
@@ -28,12 +29,16 @@ export class SearchPage {
...
-        `^Displaying (\\d+-\\d+|\\d+) of (over )?\\d{1,3}(,\\d{3})* results for ${this.searchType}(s)? "${this.searchterm}"$`
+        `^Displaying (\\d+-\\d+|\\d+) of (over )?\\d{1,3}(,\\d{3})* results for ${this.searchType}s? "${this.searchterm}"$`,
+        "i"
       ),
     })
     this.searchResult = page
       .locator("#search-results-list")
       .getByRole("link", { name: new RegExp(this.searchterm) })
+    this.searchResultsCallNumber = page.locator(
+      '#search-results-list table td:has-text("CALL NUMBER") td p span'
```
7emansell, line 40: this is another place that might be pretty brittle– is there another way to select this text?

7emansell (changes requested):

playwright/tests/Search/search_call_number.spec.ts
chrismulholland: what would be better, to just assert that the text "JFD 93-1962" appears on the page?

playwright/pages/search_page.ts
chrismulholland: do you think something like this would be better?

// 1. Locate the static text label "Call number"
const callNumberLabel = page.getByText('Call number', { exact: true });

// 2. Select the immediately following sibling element (which holds the value)
const callNumberValue = callNumberLabel.locator('xpath=./following-sibling::*[1]');

playwright/tests/Search/search_title.spec.ts
7emansell: I often see failures (times out waiting for elements to appear) on the first run, and then it is fine on the next run. ideally, they would pass 100% of the time, since these tests aren't to confirm discovery api performance, just that when the page loads, the expected elements are there. So maybe worth upping the time allowed?

playwright/pages/search_page.ts
chrismulholland: I was able to remove this without affecting the test.

7emansell (changes requested):
I think you've merged the uuid test work into this PR, which is fine, but make sure this work and the changelog entries aren't duplicated. 
this work is also missing the change with removing the "card-content" and replacing with a new "search-results-list" selector that we talked about, i think that would make this more flexible overall

PR #726, opened 2026-04-13
SCC-5195: Series in bib details

7emansell:
## Ticket:

- JIRA ticket [SCC-5195](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5195)

- Updates how we build the series links on the bib page (adding handling for `seriesUniformTitle` and `seriesAddedEntry`, removing `seriesStatement`)

## How has this been tested?

Good example bibs (they have been correctly indexed as of 4/16): 
b11603668
b22098736
b22697988
b21979887
b11018902 
b19705975
b14477060
b16470373
b16470380

### Checklist:

[SCC-5195]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/models/BibDetails.ts
+++ b/src/models/BibDetails.ts
@@ -13,7 +13,10 @@ import {
   convertToSentenceCase,
   encodeURIComponentWithPeriods,
 } from "../utils/appUtils"
-import { getFindingAidFromSupplementaryContent } from "../utils/bibUtils"
+import {
+  getFindingAidFromSupplementaryContent,
+  getSeriesSearchUrl,
+} from "../utils/bibUtils"
 import type {
   AnnotatedMarc,
   AnnotatedMarcField,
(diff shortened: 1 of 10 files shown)


src/models/BibDetails.ts
```diff
@@ -259,10 +311,39 @@ export default class BibDetails {
...
+    }
+
+    return resourceEndpointDetails
+  }
+
   combineBibDetailsData(
     resourceEndpointDetails: AnyBibDetail[],
     annotatedMarcDetails: AnyMarcDetail[]
   ): AnyBibDetail[] {
+    // Merge series added entry and series uniform title fields
```
7emansell: Annoying but this has to happen somewhere, other than merging parallels into primaries this is the only thing on the bib where we munge two fields together into one ... this felt like the clearest juncture to do it at

charmingduchess: This logic is def a candidate to move to the discovery API. Would you want to make that update?

src/models/BibDetails.ts
```diff
@@ -115,6 +115,55 @@ export default class BibDetails {
       : null
   }
 
+  buildLinkedSeriesDetail(field): LinkedBibDetail | null {
```
charmingduchess: Is there enough duplication to combine this with the author display logic?

src/models/BibDetails.ts
7emansell: discussed offline, will continue to do this here

src/models/BibDetails.ts
charmingduchess, line 1: Glad to see some code removed. Curious about the motivation to remove the deduping logic?

charmingduchess (commented):
Just a couple questions

src/models/BibDetails.ts
7emansell: I was thinking the halfway failure case might be some literal fields returned with no display strings– but no idea how thatd ever happen. So if there's no display strings, and somehow there are literals, fallback should just be the literals

src/models/BibDetails.ts
7emansell: yes, combined

src/models/BibDetails.ts
```diff
@@ -69,94 +72,67 @@ export default class BibDetails {
...
-          value: literals,
-        }
-      : null
-  }
-
-  buildLinkedSeriesDetail(literalField): LinkedBibDetail | null {
-    let displayField = "seriesDisplay"
-    let label = "Series"
     switch (literalField) {
+      case "creatorLiteral":
```
charmingduchess: Right now, there are at least two places to look for hardcoded strings related to these field mappings. Maybe we can overload the field mappings with these parameters and centralize these constants? It would also let us eliminate this switch statement

src/models/BibDetails.ts
charmingduchess: gotcha. can you add some commentary to the buildLinkedDisplayFieldDetail? before the method something like, "build a linked field with ..., default to the literalfield

charmingduchess (changes requested):

src/models/BibDetails.ts
7emansell: moved out into constants

charmingduchess (approved):
beautiful!

PR #731, opened 2026-04-22
SCC-5243: Lists API handlers

7emansell:
## Ticket:

- JIRA ticket [SCC-5243](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5243)

- Adds helpers for all anticipated lists actions on the frontend
- Adds unit tests
- JSDoc written by AI <3

### Checklist:

[SCC-5243]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/server/sierraClient/index.ts
```diff
@@ -35,40 +35,43 @@ async function buildClient() {
...
-  }
-  const deleteRequest = wrapper.deleteRequest.bind(wrapper)
-  wrapper.deleteRequest = async function (path, body) {
-    logger.info("Sierra request", {
-      path,
-      method: "DELETE",
-      body: JSON.stringify(body),
-    })
-    return await deleteRequest(path, body)
+  if (!(wrapper as any).__patched) {
```
7emansell, line 38: unrelated attempt at not double-attaching the logger methods to this client– currently all sierra logs are happening twice and i think this is why

src/server/api/lists.ts
```diff
@@ -0,0 +1,296 @@
...
+  try {
+    const client = await nyplApiClient()
+    const path = sort
+      ? `/patrons/${patronId}/lists?sort=${sort}`
+      : `/patrons/${patronId}/lists`
+    const response = await client.get(path)
+    if (response.error) {
+      logServerError("fetchLists", `${response.error} Request: ${path}`)
+      return {
+        status: response.statusCode,
```
charmingduchess: this property can be updated on the response to be `status` instead of `statusCode`, if you would like

src/server/api/lists.ts
```diff
@@ -0,0 +1,296 @@
...
+}: {
+  patronId: string
+  listName: string
+  records?: string[]
+  description?: string
+}) {
+  try {
+    const client = await nyplApiClient()
+    const path = "/list"
+    // TO DO: Add handling for default first list
```
charmingduchess: this behavior might want to be in the lists service.

src/server/api/lists.ts
charmingduchess, line 1: The error handling for these methods is repetitive. I'd like to see it pulled out into a `handleListsError` method. If you want to take it step further, if you move the 200 response return to outside the catch block, you could make it a `callListsServiceAndHandleError`... but that one is up to you.

charmingduchess (changes requested):

src/server/api/lists.ts
7emansell: no it's fine as long as it's there

src/server/api/lists.ts
7emansell: Hmm. i could go either way on this.. because regardless of this endpoint's behavior, frontend needs to know if the active user has any lists, so it might be redundant for the lists service to internally check that to return the correct first list title

src/server/api/lists.ts
7emansell: good idea, moved all the way out

src/server/api/lists.ts
```diff
@@ -1,6 +1,35 @@
 import { logServerError } from "../../utils/logUtils"
 import nyplApiClient from "../nyplApiClient"
 
+async function callListsServiceAndHandleError({
```
charmingduchess, line 4: this looks great! i imagine we could use this pattern in a lot of the server files

charmingduchess (approved):

PR #732, opened 2026-04-27
SCC-5244: Lists models

7emansell:
## Ticket:

- JIRA ticket [SCC-5244](https://newyorkpubliclibrary.atlassian.net/browse/SSCC-5244)

- Establishes `List` and `ListRecord` model
- Apologies bc this PR got a little into the rendering weeds since that's how I process this best but all that patron data context can be ignored for now, just need review on the new `/models`

### Checklist:

[SCC-5244]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/types/listTypes.ts
```diff
@@ -0,0 +1,23 @@
+import type { HTTPStatusCode } from "./appTypes"
+
+export interface ListErrorResponse {
+  statusCode: HTTPStatusCode
+  name: string
+  error: string
+}
+
+export interface ListResult {
```
charmingduchess, line 9: I'm not sure that Result adds anything to this model name. Is it to differentiate from a different kind of List or ListRecord in the future?

src/models/ListRecord.ts
```diff
@@ -0,0 +1,25 @@
+import type { ListRecordResult } from "../types/listTypes"
+import { formatMMDDYYYY } from "../utils/dateUtils"
+
+/**
+ * The ListRecord class represents a single record in a List and contains the data
+ * and getter functions used in search results and in My Account display.
+ */
+export default class ListRecord {
```
charmingduchess, line 8: Can you add a default for `addedDate`? I'm thinking about how this would function on the "add to lists" flow. My brain is a little scrambled by pydantic modelling, so if you were thinking something else, let me know.

charmingduchess (commented):
This is looking good! I'm wondering about an assumption that Lists and ListsRecords will always be instantiated by an API result, though that may just be the names of the types and arguments. Let's keep an eye out and make sure these models are compatible with a new list generated by a patron.

7emansell:
> Let's keep an eye out and make sure these models are compatible with a new list generated by a patron.

Not sure I'm understanding this case... how could a patron independently generate a list without making an API request? @charmingduchess

src/types/listTypes.ts
7emansell: it's to differentiate from a List model! ListResult is the unprocessed result we get back from the API. Like `ListResult` is to `List` as `DiscoveryBibResult` is to `Bib`

src/models/ListRecord.ts
7emansell: hmm I'm not sure what the logical default would be here– i'm thinking if the add record operation somehow doesn't return this value on a record, that should be an error, not caught by a default

charmingduchess (approved):
concerns addressed offline - we are not going to be displaying lists except for once they've been returned from the api, so don't need to worry about modelling them before that

PR #737, opened 2026-05-07
Scc 5374/playwright request scan

chrismulholland:
## Ticket:

- JIRA ticket [5374](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5374)

- Added Playwright tests for requesting a scan of an item

-

### Checklist:


playwright/fixtures/requestBibs.ts
```diff
@@ -0,0 +1,109 @@
+/**
+ * Pool of bib IDs with requestable physical items on the QA server. Chris has a local playwright_experimental branch with a script for getting these bib ids.
```
7emansell: thinking the second sentence of this comment is not gonna be helpful to future devs. Is there a reason why you didn't want to push up this script?

7emansell (changes requested):
Looks great overall!  i am still worried about the underlying issue with repeated requests but hopefully we will make some progress on it with [this ticket](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5371) and update these tests accordingly. In the meantime, i think picking a random bib each time is the best we can do.
Just one question/comment on how the bib list was generated

playwright/fixtures/requestBibs.ts
chrismulholland: I have an experimental branch locally with a script that simply visits https://qa-www.nypl.org/research/research-catalog/search?q=&filters[buildingLocation][0]=ma&filters[format][0]=a&page=2
and scrapes the bnums from the title hrefs.  It prints them to console and I just copied them for use here.  Eventually I'll need to refresh the bib list from a different set but I thought this would keep us for a while.
If you think we should have a solution built into this repo, we should discuss a strategy.  thanks!

7emansell: I think if we're generating a list of particular bibs that our tests rely on, we should include that script in the repo! You can just add the script in this PR

chrismulholland: Sure, that sounds like a good idea.  just fyi, it may take me a few days to get back to this.

chrismulholland: @7emansell , I added in a script to get a random bib that has active onsite and scan request buttons. and all tests are passing for me :)

CHANGELOG
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added Playwright tests for requesting a scan of an item [SCC-5374](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5374)
+- Added Playwright tests for requesting an item for onsite [SCC-5323](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5323)
+- Added series and genre to advanced search fields [SCC-5179](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5179)
```
7emansell, line 14: Correct the changelog

7emansell (changes requested):
I'm seeing repeated failures for both tests across all three browsers when I run these locally (specifically, the check for the "Cancel request" button at the end of the test). Could be my setup but not a great sign– are you seeing successes every time you run these?
Also- glad we have a test for this but I'm realizing it's gonna be very tedious for Scan & Delivery to filter out the many test scans (we push to QA at least once a day if not more– that's a lot of fake requests). Can we set this up so that the scan requests only run on push to production?

CHANGELOG
chrismulholland: that's probably a good thing to do.  I can cancel this pull request and save the request delivery and scan tests for production.  we won't have the request of same item problem on prod either.

7emansell: Okay– I'm asking that the tests are set up to run on push to production, not to remove the tests entirely. That can happen in this PR

chrismulholland: I was talking about this with Sam.  She suggested that I pause on this work for now while I document how requests work across the QA and production environments, and who's involved.   And keeping https://newyorkpubliclibrary.atlassian.net/browse/SCC-5371 in mind, too.  I'll be sure to share that doc with you for your input.  thanks.

PR #742, opened 2026-05-29
SCC-5247: CRUD lists

7emansell:
## Ticket:

- JIRA ticket [SCC-5247](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5247)

Sorry in advance for how many files this PR is!! Partially bc I'm redoing my initial take on how lists/list records should be initialized– no more models. Like the other account data, lists will now be built in the `MyAccount` factory and live/be updated in the `PatronDataContext`.

- Adds Create/edit list modal and buttons
- Adds delete modal and buttons
- Adds duplicate, download
- Exposes some more internal API endpoints for the above operations, replaces use of PUT to PATCH (`node-nypl-data-api-client` doesn't have PUT for no good reason, [ticket to add it here](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5415), but in the meantime I updated `ListsService` and API Gateway to use PATCH for the update list operations)
- Revises how bib data is merged with list records and how sorting works: see `listUtils` > `buildListRecords`

## How should this be tested?

Test user `23333094983077` now has a list with 300+ items, so that's a good list to test sorting on. 
Otherwise, try all four operations from the all lists view and from the single list view. 

- Does the "duplicate" from the single list view seem to take a long time loading or is it just me? 

### Checklist:

[SCC-5247]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/MyAccount/ListsTab/ListActions/CreateEditList.tsx
```diff
@@ -0,0 +1,276 @@
...
+  const { updatedAccountData, setUpdatedAccountData } =
+    useContext(PatronDataContext)
+  const { patron, lists } = updatedAccountData
+
+  const [listName, setListName] = useState(list?.listName || "")
+  const [listDescription, setListDescription] = useState(
+    list?.description || ""
+  )
+  const [isSubmitting, setIsSubmitting] = useState(false)
+  const [isSubmitted, setIsSubmitted] = useState(false)
```
charmingduchess: is it possible to derive `isSubmitted` from `isSubmitting` states?  eg if the previous state of `isSubmitting` is true and the current state is false, then `isSubmitting` is true. Might be a better pattern to link these intrinsically if they are referring to the same thing.

src/components/MyAccount/ListsTab/ListActions/CreateEditList.tsx
```diff
@@ -0,0 +1,276 @@
...
+              lists: updatedLists,
+            })
+          } else {
+            setUpdatedAccountData({
+              ...updatedAccountData,
+              lists: [data.list, ...lists],
+            })
+          }
+        }
+        setStatus("success")
```
charmingduchess, line 104: is there a status besides success or failure?

src/components/MyAccount/ListsTab/ListActions/CreateEditList.tsx
```diff
@@ -0,0 +1,276 @@
...
+        setStatusMessage(
```
charmingduchess, line 105: if statusMessage can be dependent on status, can you refactor so it is derived from status?

src/components/MyAccount/ListsTab/ListActions/ListActionsMenu.tsx
```diff
@@ -0,0 +1,205 @@
...
+          })
+          if (response.ok) {
+            const data = await response.json()
+            if (data && data.list) {
+              setUpdatedAccountData({
+                ...updatedAccountData,
+                lists: [data.list, ...lists],
+              })
+            }
+            setStatus("success")
```
charmingduchess: echoing an earlier comment, it might even be useful for debugging down the line to have statuses reflect what succeeded or failed. i could see something like the following: { "duplicate-success": "Your list has been duplicated"}. Copy updates would also be easier, since all of the messages would be centralized.

src/components/MyAccount/ListsTab/ListDisplay.tsx
```diff
@@ -1,27 +1,66 @@
...
+
+  const bannerRef = useRef<HTMLDivElement>(null)
+
+  // Manage status banner display for list actions
+  const [status, setStatus] = useState<StatusType>("")
+  const [statusMessage, setStatusMessage] = useState<string>("")
+  useEffect(() => {
+    if (status !== "" && bannerRef.current) {
+      setTimeout(() => {
+        bannerRef.current?.focus()
```
charmingduchess, line 50: what condition are you solving for here? setTimeout feels a little flakey.

src/components/MyAccount/ListsTab/ListDisplay.tsx
```diff
@@ -1,27 +1,66 @@
...
+      }, 100)
+    }
+  }, [status])
+
+  // DS Modal controls used for Delete list modal
+  const { onOpen: openModal, onClose: closeModal, Modal } = useModal()
+  const [modalProps, setModalProps] = useState<BaseModalProps>()
+
+  if (!list) return null
```
charmingduchess: any reason not to have this early return right after we get the list from the patron data fetcher?

src/components/MyAccount/ListsTab/ListDisplay.tsx
```diff
@@ -99,9 +193,129 @@ const ListDisplay = ({ list }: { list?: List }) => {
...
+                    window.open(
+                      `${BASE_URL}/account/lists/${data.list.id}`,
+                      "_blank"
+                    )
+                  }
+                  setStatus("success")
+                  setStatusMessage("Your list has been duplicated.")
+                } else {
+                  setStatus("failure")
+                  setStatusMessage("Your list could not be duplicated.")
```
charmingduchess: this section is very similar to what's going on in ListActionsMenu. Definitely pull out the duplicated strings, and/or maybe the onClick (and more) is identical.

src/components/MyAccount/ListsTab/ListRecordsTable.tsx
```diff
@@ -46,39 +53,51 @@ const ListRecordsTable = ({ list }: { list: List }) => {
...
 
-  // TO DO: caching...
+  // TO DO: caching??
   useEffect(() => {
     const fetchBibData = async () => {
       if (!list || sortedRecords.length === 0) return
       setIsLoading(true)
 
-      const isLocalSort = activeSort.includes("added_date")
+      const canSortWithoutBibData = activeSort.includes("added_date")
```
charmingduchess, line 62: great rename

src/components/MyAccount/ListsTab/ListRecordsTable.tsx
```diff
@@ -46,39 +53,51 @@ const ListRecordsTable = ({ list }: { list: List }) => {
...
       const startIndex = (currentPage - 1) * LIST_RECORDS_PER_PAGE
       const endIndex = startIndex + LIST_RECORDS_PER_PAGE
 
-      // If sorting by title/author, fetch ALL records so the Discovery API can sort globally
-      // otherwise: sorting by date added, only need to fetch the 20 records for the current page
-      const recordsToFetch = isLocalSort
+      // If sorting by title/author/call number, need to fetch bib data for all records
+      // otherwise: sorting by date added, only need to fetch bib data for the current page
```
charmingduchess, line 67: i wonder if this front end component needs to be thinking about that. is there a way to implement this where the front end sends the query to the server, and the server silently decides whether to actually refetch? it feels a little in the weeds for front end logic, though maybe its not feasible to delegate that responsibility.

charmingduchess: It also looks like this component is implementing pagination in a way that feels similarly overloaded. its possible that the patron data model might be a place to abstract this logic to.

charmingduchess (changes requested):
thought i would do this in two portions. looks good so far

src/components/MyAccount/ListsTab/ListsDisplay.tsx
```diff
@@ -1,51 +1,80 @@
...
   const sortMenuRef = useRef<HTMLDivElement | null>(null)
   const [activeSort, setActiveSort] = useState("modified_date_desc")
 
+  const [isLoading, setIsLoading] = useState(false)
+
+  // Manage status banner display for list actions
+  const bannerRef = useRef<HTMLDivElement>(null)
+  const [status, setStatus] = useState<StatusType>("")
+  const [statusMessage, setStatusMessage] = useState<string>("")
+  useEffect(() => {
```
charmingduchess, line 41: ditto this useeffect. can it be done with the focus context? or is it maybe functionality that can be built into it?

charmingduchess: if it's unavoidable, i think it falls into [this](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state) category and probably doesn't need a useeffect

src/components/MyAccount/Settings/StatusBanner.tsx
```diff
@@ -7,23 +7,36 @@ type StatusBannerProps = {
   statusMessage: string
 }
 
-const successContent = <Text>Your changes were saved.</Text>
+const generalSuccessContent = <Text>Your changes were saved.</Text>
 
 const generalFailureContent = <Text>Your changes were not saved.</Text>
 
 const specificFailureContent = (statusMessage: string) => {
```
charmingduchess, line 14: sorry to repeat again, but looking at this component, i want to plug that mapping approach i described above. especially since you're verifying both status and status message in those conditionals, it makes sense to have statuses map directly to a status message.

src/utils/listUtils.ts
```diff
@@ -48,57 +56,201 @@ export const listResultsHeading = (list, currentPage) => {
   }`
 }
 
+export const buildListRecordWithBibData = (
+  listRecord: ListRecord | ListRecordResult,
+  bibData?: any
+): ListRecord => {
+  const bibResult = bibData?.result || bibData || {}
+
+  const addedDate =
```
charmingduchess: is this about a discrepancy in the data? let's update the api if so.

charmingduchess (changes requested):

src/components/MyAccount/ListsTab/ListDisplay.tsx
7emansell: In fact no reason to have this line at all– there's no error fallback, `list` should be a required prop for this component. rmd

src/components/MyAccount/ListsTab/ListDisplay.tsx
7emansell: Just giving time for the banner to pop up, otherwise sometimes it loses the rendering race. Not not flaky but not sure if there's any other way to do this and is a common pattern for focus

src/components/MyAccount/ListsTab/ListActions/CreateEditList.tsx
7emansell: I think **most** of the time it'd be possible to derive `isSubmitted` but sometimes it's a race condition between the modal closing and `isSubmitted` being defined correctly. `isSubmitting` represents like... the loading while `isSubmitted` just means there was an action taken in the modal, so their state feels like it should be separate

src/utils/listUtils.ts
7emansell: No it's that this function can be used for either raw list records or for already hydrated list records and that's the formatting difference on that field

src/components/MyAccount/ListsTab/ListsDisplay.tsx
7emansell: it needs to be post render of the banner and post modal closing, so it has to use the `ref` otherwise focus could go to empty div. Have experimented but this is the only way i've found to make sure that happens every time

7emansell:
@charmingduchess All status comments noted– I'm gonna make another PR off the most recent lists one to make a status mapping and clean it all up (which will also apply to the account settings status banner).

src/components/MyAccount/ListsTab/ListRecordsTable.tsx
7emansell: maybe we can discuss offline i see the gnarliness but want to walk through if it's actually doing more/less than it needs to

src/components/MyAccount/ListsTab/ListDisplay.tsx
charmingduchess: going to revisit in a further pr

src/utils/listUtils.ts
charmingduchess: as discussed offline, let's update the property on the hydrated version to be something like `formattedDate`, and a comment describing the 2 uses of the method

charmingduchess (changes requested):

src/utils/listUtils.ts
```diff
@@ -48,57 +56,201 @@ export const listResultsHeading = (list, currentPage) => {
...
-      const indexB = pageRecords.findIndex((r) => r.uri === b.uri)
+      const indexA = indexMap.get(a.uri) ?? Infinity
+      const indexB = indexMap.get(b.uri) ?? Infinity
       return indexA - indexB
     })
+  } else {
+    updatedRecords.sort((a: any, b: any) => {
+      let valA = ""
+      let valB = ""
+      if (activeSort.includes("title")) {
```
charmingduchess: why cant these sorts happen in discovery api?

7emansell: They can and were– i was testing to see if this was faster. It's not noticeably better (at least to my eye) so im reverting to the discoapi sort

src/components/MyAccount/ListsTab/ListRecordsTable.tsx
charmingduchess: offline discussion: all of my account is very client-fetch heavy due to architecting around tab components. we may want to refactor this in the future, but not as part of lists work. refactor could include each tab actually being their own pages with their own server files and getServerSideProps fetching

charmingduchess (approved):

PR #744, opened 2026-06-01
Add default list handling

7emansell:
## Ticket:

- no ticket just trying to split up Lists work into more bite sized chunks

- Adds the default list to all user accounts and disables deletion and editing
- Note that for test users who already had list(s), their default undeletable list will be set as the oldest created one– no reason this should be an issue in production though (when lists are deployed, every user will have their default list created on login)
- An alternate way to handle this is by adding a `isDefaultList` field in ListsService instead of letting FE handle it but I'm not particularly pressed either way...

## How has this been tested?

Log in to a clean account (`webapplicant0`, message me for pw) and confirm that their one list is initialized with title "My workspace (default)" and is not deletable or editable (no option in the menu, no button in the single list view)

### Checklist:

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added lists tab and all lists table [SCC-5245](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5245)
 - Added single list display [SCC-5246](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5246)
 - Added CRUD operations for lists [SCC-5247](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5247)
+- Added default list handling
 
 ### Updated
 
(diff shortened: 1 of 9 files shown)


mwbenowitz (commented):
👍 to the code here, looks good.

I just want to raise the question about the implementation of the `isDefaultList` functionality here in the RC frontend. I worry that it violates the division of concerns and will make this functionality opaque down the line. Especially if we move into a world in which lists are potentially shared across products (which I know was envisioned in the `ListService`)

So I guess a light preference from me for implementing that there, open to discussion though.

7emansell:
> Especially if we move into a world in which lists are potentially shared across products (which I know was envisioned in the ListService)
> So I guess a light preference from me for implementing that there, open to discussion though.

Agreed and discussed with Vera offline, PR to ListsService [here](https://github.com/NYPL/ListsService/pull/18)

7emansell:
Changed minds cc @charmingduchess @mwbenowitz sticking with frontend implementation (for posterity: since FE has to be aware of default lists even with BE implementation + who's to say future consumers will want a default list)

PR #746, opened 2026-06-03
Adding Lists context to search/bib

7emansell:
## Ticket:

- no ref, another Lists chunk

- Adds `ManageBibInList` button, which so far handles saving and removing a bib from the one default list
- Adds `PatronDataContext` to search and bib pages, only populated with the patron and their lists (to avoid fetching all the other slow Sierra data) which is enough to display whether a bib is in a list or not
- Adds authentication check and redirect of focus back to where you left it when you try to save a bib not logged in
- Reworks delete record endpoint (also [some work in ListsService](https://github.com/NYPL/ListsService/commit/d1c46065f43080e13dabf084b047b1c12256ac89)), can be extended easily to delete in bulk but right now only use case is deleting one record at a time, so the endpoint reflects that
- To come: the full manage bib menu, where you can add/remove to multiple lists and create a list etc. 

## How has this been tested?

Best tested on `webapplicant0` since Hank has only the one default list. Try adding and removing bibs from search results and the bib page, try logging out and getting redirected back in and confirming you end up where you were originally

### Checklist:

### DIFF
--- a/pages/api/account/lists/records.ts
+++ b/pages/api/account/lists/records.ts
@@ -27,5 +27,14 @@ export default async function handler(
       patronId,
     })
     res.status(response.status || 200).json(response)
+  } else if (req.method === "DELETE") {
+    const response = await addRecordsToList({
+      records: uris,
+      listId,
+      patronId,
+    })
+    res.status(response.status || 200).json(response)
+  } else {
+    return res.status(405).json({ error: "Method not allowed" })
   }
 }
--- a/pages/api/account/lists/records.ts
+++ b/pages/api/account/lists/records.ts
@@ -1,21 +1,31 @@
 import type { NextApiRequest, NextApiResponse } from "next"
-import { fetchBibRecords } from "../../../../src/server/api/lists"
+import {
+  addRecordsToList,
+  fetchBibRecords,
+} from "../../../../src/server/api/lists"
 import type { ListRecordsSort } from "../../../../src/types/listTypes"
 
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse
 ) {
-  if (req.method !== "GET") {
-    return res.status(405).json({ error: "Method not allowed" })
-  }
-
+  // Records requested in query for easy GETting, note that this flips the pattern of other list endpoints
   const { uris, sort } = req.query
+  const { listId, patronId } = req.body
 
   if (!uris || typeof uris !== "string") {
     return res.status(400).json({ error: "Missing or invalid uris" })
   }
 
-  const response = await fetchBibRecords(uris, sort as ListRecordsSort)
-  res.status(response.status || 200).json(response)
+  if (req.method === "GET") {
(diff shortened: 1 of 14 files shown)


pages/api/account/lists/records.ts
```diff
@@ -1,21 +1,41 @@
...
 
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse
 ) {
-  if (req.method !== "GET") {
-    return res.status(405).json({ error: "Method not allowed" })
-  }
-
+  // Records requested in query for easy GETting, note that this flips the pattern of other list endpoints
```
charmingduchess: what is the pattern of the other list endpoints? might be good to summarize here.

src/components/SearchResults/ManageBibInList.tsx
```diff
@@ -0,0 +1,212 @@
...
+      .
+    </Text>
+  )
+  // Focus the Save button upon returning from the login redirect
+  useEffect(() => {
+    const params = new URLSearchParams(window.location.search)
+    const focusTarget = params.get("focus")
+
+    if (focusTarget === `manage-bib-${bib.id}`) {
+      buttonRef.current?.focus()
```
charmingduchess: can we use the focus context here? and assign `manage-bib-${bib.id}` as the id of the buttons?

src/components/SearchResults/SearchResult.tsx
```diff
@@ -50,6 +56,18 @@ const SearchResult = ({ bib }: SearchResultProps) => {
     return acc
   }, [])
 
+  // Manage status banner display for list actions
+  const bannerRef = useRef<HTMLDivElement>(null)
```
charmingduchess, line 60: I feel like all of this logic can live on the status banner. it doesn't seem like there are any cases where the status banner is rendered but we don't want to focus on it. can the status banner handle the status banner focusing by just automatically focusing on itself on the initial render?

src/components/SearchResults/ManageBibInList.tsx
```diff
@@ -0,0 +1,212 @@
...
+              patronId: updatedAccountData?.patron?.id?.toString(),
+              listId: defaultListId,
+            }),
+          }
+        )
+        if (response.ok) {
+          const responseData = await response.json()
+          const updatedList = responseData.list || responseData
+
+          if (setUpdatedAccountData) {
```
charmingduchess: when would this not be defined?

charmingduchess (changes requested):

src/components/SearchResults/ManageBibInList.tsx
7emansell: Wouldn't atp, rmd

src/components/SearchResults/SearchResult.tsx
7emansell: I don't know if it can all live on the status banner but i think this concern is resolved in https://github.com/NYPL/research-catalog/pull/749

PR #747, opened 2026-06-05
SCC-5269: Warning log for no results with single filter

KaseyW31:
## Ticket:

- JIRA ticket [SCC-5269](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5269)

- Logs a warning when a search query is performed with no keyword and a single filter, and no results are found

- Run RC locally and use Postman/curl to send a request to a bad filter (e.g. http://local.nypl.org:8080/research/research-catalog/search?filters%5Bformat%5D%5D
), manually setting referrer in the headers. Verify that the warning log has been logged locally with the correct referrer.

## Accessibility updates?

- N/A

### Developer checklist

[SCC-5269]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Prerelease
 
+### Added
+
 - Added warn log for no results from query with a single filter and no keyword [SCC-5269](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5269)
 
 ## [2.0.0] 2026-06-04
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Prerelease
 
+- Added warn log for no results from query with a single filter and no keyword [SCC-5269](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5269)
+
 ## [2.0.0] 2026-06-04
 
 ### Added
--- a/__test__/pages/search/search.test.tsx
+++ b/__test__/pages/search/search.test.tsx
@@ -2,7 +2,7 @@ import { getServerSideProps } from "../../../pages/search/index"
 import initializePatronTokenAuth from "../../../src/server/auth"
 import { fetchSearchResults } from "../../../src/server/api/search"
 import { mapQueryToSearchParams } from "../../../src/utils/searchUtils"
-import { SearchQueryParams } from "../../../src/types/searchTypes"
+import type { SearchQueryParams } from "../../../src/types/searchTypes"
 import { logSingleFilterNoResults } from "../../../src/utils/logUtils"
 
 jest.mock("../../../src/server/auth")
--- a/__test__/pages/search/search.test.tsx
+++ b/__test__/pages/search/search.test.tsx
@@ -15,6 +15,24 @@ const mockReq = {
   },
 }
 
+jest.mock("@nypl/node-utils", () => ({
+  logger: {
+    info: jest.fn(),
+    warn: jest.fn(),
+    error: jest.fn(),
+  },
+}))
+import { logger } from "@nypl/node-utils"
+
+const mockReqWithReferer = {
+  ...mockReq,
+  headers: {
+    host: "local.nypl.org:8080",
+    referer:
+      "http://local.nypl.org:8080/research/research-catalog/bib/b16767329",
+  },
+}
+
 describe("Search page", () => {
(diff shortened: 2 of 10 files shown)


pages/browse/authors/[slug].tsx
```diff
@@ -97,7 +100,20 @@ export async function getServerSideProps({ req, query, params }) {
...
+  // Check for possible bad incoming link (no results with one filter and no keyword)
+  if (
+    results.status === 404 &&
+    filtersObjectLength(searchParams.filters) === 1 &&
+    !searchParams.q
+  )
+    logServerWarn(
+      "fetchSearchResults",
+      `Possible bad incoming link; referer: ${referer}`
+    )
```
Comment (mwbenowitz) [inline pages/browse/authors/[slug].tsx]: If this is repeated a few times, would it make sense to refactor it out to a utility method? That way future functionality changes would only need to be made once. (And I could see us changing the definition of "bad incoming link" as we learn more about this.)

mwbenowitz (commented):
Functionality seems good to me! Just one suggestion about code organization that would hopefully help keep this maintainable.

__test__/pages/search/search.test.tsx
7emansell, line 1: All unit tests look great

pages/browse/authors/[slug].tsx
```diff
@@ -97,7 +100,20 @@ export async function getServerSideProps({ req, query, params }) {
...
+  const results = await fetchSearchResults(searchParams)
+
+  // Check for possible bad incoming link (no results with one filter and no keyword)
+  if (
+    results.status === 404 &&
+    filtersObjectLength(searchParams.filters) === 1 &&
+    !searchParams.q
+  )
+    logServerWarn(
+      "fetchSearchResults",
```
Comment (7emansell) [inline pages/browse/authors/[slug].tsx]: Nit: this isn't happening in `fetchSearchResults`. More informative location could be "Search results gSSP"

pages/browse/authors/[slug].tsx
Comment (7emansell) [reply pages/browse/authors/[slug].tsx]: agreed! A method that takes `location, results, searchParams, referer` for example. Be careful about where this method is declared though– since it uses the serverside logger, it can't live in most of the clientside util files or Next will try to bundle it for the client on build. Could go in `logUtils`

pages/browse/authors/[slug].tsx
```diff
@@ -97,7 +100,20 @@ export async function getServerSideProps({ req, query, params }) {
...
+      `Possible bad incoming link; referer: ${referer}`
```
Comment (7emansell) [inline pages/browse/authors/[slug].tsx]: Two things here: this log should be more specific (like "Link to single filter, no results"). Also, if the `referer` is `undefined`, we know the user likely manually typed the filter value in so it's being treated as direct navigation. If that's the case, we don't care that the filter value returns nothing (and we don't learn anything from the log since it's undefined). I think we can just not log at all if there's no `referer`

CHANGELOG
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Prerelease
 
+- Added warn log for no results from query with a single filter and no keyword [SCC-5269](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5269)
```
7emansell, line 12: Beautiful, you just need to make sure to add the section heading as well if it's the first of its type under this release. Here, it should be "###Added"

7emansell (changes requested):
Overall great just some nitpicky minor tweaks!! Also I know I advised against pulling this out into a method... now that I see it all, it does make sense for readability (it'll save like 3 lines of code I guess too). 
Also, as your first RC PR I would try to give feedback on the general PR description/changelog but it's pretty much perfect :) Thank you for the attention to detail/previous PR patterns

src/utils/logUtils.ts
```diff
@@ -7,3 +9,28 @@ export const logServerError = (location: string, message: string): void => {
...
+  referer: string | undefined
+): void => {
+  if (
+    referer &&
+    results.status === 404 &&
+    filtersObjectLength(searchParams.filters) === 1 &&
+    !searchParams.q
+  )
+    logServerWarn(
+      "Search results gSSP",
```
7emansell: this is the arg where location should be set

7emansell (changes requested):
Just one fix otherwise fantastic!

7emansell (approved):
Awesome. you can go ahead and merge this into main!

PR #749, opened 2026-06-09
Fix status/focus

7emansell:
## Ticket:

- n/a, comes from lots of statii on Lists

**status** (first two commits)
- In various places (mainly My Account) we display a status banner when a user takes a successful or failed action
- Passing that displayed message and type (success/failure) was using two states. By mapping a bunch of statuses to their corresponding type and message, now it can just be one state, distinguished by a string in `STATIC_STATUS_MESSAGES` or `DYNAMIC_STATUS_MESSAGES`

**focus**
- Additionally, removes `useEffect` + `ref` approach to focus from list status banners
- Adds 100ms delay to the part of `FocusContext` that actually focuses the active element ID, which allows the DOM to flush and display whatever it wants (like a banner!)

All status banner interactions should be exactly the same

## Accessibility updates?

n/a

## Developer checklist

### DIFF
--- a/pages/bib/[id]/index.tsx
+++ b/pages/bib/[id]/index.tsx
@@ -55,6 +55,7 @@ import { PatronDataProvider } from "../../../src/context/PatronDataContext"
 import MyAccount from "../../../src/models/MyAccount"
 import { StatusBanner } from "../../../src/components/MyAccount/Settings/StatusBanner"
 import type { StatusBannerState } from "../../../src/components/MyAccount/Settings/StatusBanner"
+import { idConstants } from "../../../src/context/FocusContext"
 
 interface BibPropsType {
   discoveryBibResult: DiscoveryBibResult
@@ -103,15 +104,7 @@ export default function BibPage({
   const controllerRef = useRef<AbortController>()
 
   // Manage status banner display for list actions
-  const bannerRef = useRef<HTMLDivElement>(null)
   const [status, setStatus] = useState<StatusBannerState | null>(null)
-  useEffect(() => {
-    if (status && bannerRef.current) {
-      setTimeout(() => {
-        bannerRef.current?.focus()
-      }, 100)
-    }
-  }, [status])
 
   if (errorStatus) {
     return (
@@ -251,7 +244,7 @@ export default function BibPage({
         </Box>
         <div
           tabIndex={-1}
-          ref={bannerRef}
+          id={idConstants.listStatusBanner}
           style={{ marginTop: "-16px", marginBottom: "32px" }}
         >
           {status && (
--- a/pages/bib/[id]/index.tsx
+++ b/pages/bib/[id]/index.tsx
@@ -53,8 +53,8 @@ import UserGuideBanner from "../../../src/components/Banners/UserGuideBanner"
 import { ManageBibInList } from "../../../src/components/List/ManageBibInList"
 import { PatronDataProvider } from "../../../src/context/PatronDataContext"
 import MyAccount from "../../../src/models/MyAccount"
-import type { StatusType } from "../../../src/components/MyAccount/Settings/StatusBanner"
 import { StatusBanner } from "../../../src/components/MyAccount/Settings/StatusBanner"
+import type { StatusBannerState } from "../../../src/components/MyAccount/Settings/StatusBanner"
 
 interface BibPropsType {
   discoveryBibResult: DiscoveryBibResult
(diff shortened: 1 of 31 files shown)


src/utils/statusUtils.tsx
charmingduchess, line 1: 👍🏻 👍🏻 👍🏻

src/components/MyAccount/Settings/UsernameForm.tsx
```diff
@@ -91,21 +85,19 @@ const UsernameForm = ({ patron, usernameState }: UsernameFormProps) => {
       const responseMessage = await response.json()
       if (responseMessage !== "Username taken" && response.status === 200) {
         await getMostUpdatedSierraAccountData()
-        setUsernameStatus("success")
+        setUsernameStatus(STATIC_STATUS_MESSAGES["account-success"])
```
charmingduchess: accessing a mapping with string keys here is a little risky. dot notation will make it more obvious when there's a typo in the what status you're trying to access (eg `STATIC_STATUS_MESSAGES["userrname-failure"]` vs `STATIC_STATUS_MESSAGES.usernameFailure`).

src/context/FocusContext.tsx
```diff
@@ -51,10 +52,13 @@ export const FocusProvider = ({ children }: { children: React.ReactNode }) => {
   const setFocusById = useCallback(
     (id: string) => {
       if (isClient) {
-        const el = document.getElementById(id)
-        if (el) {
-          el.focus()
-        }
+        // Focus happens after React flushes DOM updates
```
charmingduchess, line 60: we discussed offline, but let's try and avoid this timeout when we dont need to. maybe you could have a delayedFocus state in the context, and expose a method from the context like setPersistentFocusWithDelay = (id) => { setUseDelay(true) setPersistentFocus(id) }

pages/bib/[id]/index.tsx
```diff
@@ -103,16 +104,7 @@ export default function BibPage({
...
-  const [status, setStatus] = useState<StatusType>("")
-  const [statusMessage, setStatusMessage] = useState<string>("")
-  useEffect(() => {
-    if (status !== "" && bannerRef.current) {
-      setTimeout(() => {
-        bannerRef.current?.focus()
-      }, 100)
-    }
-  }, [status])
+  const [status, setStatus] = useState<StatusBannerState | null>(null)
```
Comment (charmingduchess) [inline pages/bib/[id]/index.tsx:107]: it looks like there are two different setstatus methods being used in lists and the other other account settings. if we're centralizing the status options, i think it makes sense to always have the same function be used to update the status. it would probably also make sense to have that state live on the statusBanner.

src/context/FocusContext.tsx
```diff
@@ -73,7 +77,9 @@ export const FocusProvider = ({ children }: { children: React.ReactNode }) => {
   )
 
   useEffect(() => {
-    setIsClient(true)
+    setTimeout(() => {
+      setIsClient(true)
+    }, 100)
```
charmingduchess: do you need both of these setTimeouts? i fear we are now adding 200 ms to every focus activity

charmingduchess (changes requested):

src/context/FocusContext.tsx
7emansell: No! Eek sorry . Too many experiments, this one is removed

src/components/MyAccount/ProfileHeader.tsx
```diff
@@ -12,34 +12,29 @@ import type { IconListElementPropType } from "./IconListElement"
...
   const { isLargerThanMobile } = useNYPLBreakpoints()
-  const [usernameStatus, setUsernameStatus] = useState<StatusType>("")
-  const [usernameStatusMessage, setUsernameStatusMessage] = useState<string>("")
+  const [usernameStatus, setUsernameStatus] =
+    useState<StatusBannerState | null>(null)
   const usernameBannerRef = useRef<HTMLDivElement>(null)
 
   useEffect(() => {
-    if (usernameStatus !== "" && usernameBannerRef.current) {
+    if (usernameStatus && usernameBannerRef.current) {
```
charmingduchess: can this be replaced by focus context?

src/components/List/ManageBibInListMenu.tsx
```diff
@@ -21,41 +21,41 @@ import {
...
   const createListButtonRef = useRef<HTMLButtonElement>(null)
   const internalBannerRef = useRef<HTMLDivElement>(null)
-  const [listCreationStatus, setListCreationStatus] = useState<StatusType>("")
-  const [listCreationStatusMessage, setListCreationStatusMessage] =
-    useState<string>("")
+  const [listCreationStatus, setListCreationStatus] =
+    useState<StatusBannerState | null>(null)
   useEffect(() => {
-    if (listCreationStatus !== "" && internalBannerRef.current) {
+    if (listCreationStatus && internalBannerRef.current) {
```
charmingduchess: ditto

charmingduchess (approved):

PR #752, opened 2026-06-12
SCC 5420/playwright nyql - Adds query search tests

chrismulholland:
## Ticket:

- JIRA ticket [5420](https://newyorkpubliclibrary.atlassian.net/browse/5420)

Adds Playwright tests for query searches.

All tests pass locally.

### Accessibility checklist

### Developer checklist

### DIFF
--- a/playwright/pages/search_page.ts
+++ b/playwright/pages/search_page.ts
@@ -32,21 +32,11 @@ export class SearchPage {
   }
 
   get searchResultsHeading() {
-    // Escape special regex characters in searchterm
-    const escapedSearchterm = this.searchterm.replace(
-      /[.*+?^${}()|[\]\\]/g,
-      "\\$&"
-    )
-
-    // Query searches use different heading format: "for query: <syntax>"
-    // Standard searches use: "for <SearchType> "<searchterm>""
-    const headingPattern =
-      this.searchType.toLowerCase() === "query"
-        ? `Displaying (\\d+-\\d+|\\d+) of (over )?\\d{1,3}(,\\d{3})* results for query: ${escapedSearchterm}`
-        : `Displaying (\\d+-\\d+|\\d+) of (over )?\\d{1,3}(,\\d{3})* results for ${this.searchType}s? "${escapedSearchterm}"`
-
     return this.page.getByRole("heading", {
-      name: new RegExp(headingPattern, "i"),
+      name: new RegExp(
+        `Displaying (\\d+-\\d+|\\d+) of (over )?\\d{1,3}(,\\d{3})* results for ${this.searchType}s? "${this.searchterm}"`,
+        "i"
+      ),
     })
   }
 
--- a/playwright/pages/search_page.ts
+++ b/playwright/pages/search_page.ts
@@ -19,10 +19,7 @@ export class SearchPage {
 
     this.search_dropdown = page.getByLabel("Select a category")
     this.search_input = page.getByRole("textbox")
-    this.search_submit_button = page.getByRole("button", {
-      name: "Search",
-      exact: true,
-    })
+    this.search_submit_button = page.locator(".ds-searchBar-button")
 
     this.searchResultsContainer = page.locator("#search-results-list")
     this.searchResults = page.locator("#search-results-list h3 a")
@@ -46,6 +43,9 @@ export class SearchPage {
     await this.search_dropdown.selectOption({ label: searchType })
 
     await this.search_input.fill(searchterm)
-    await this.search_submit_button.click()
+
+    // Firefox can intermittently show a portal overlay during hydration.
+    // Submitting with Enter avoids pointer interception on the button.
+    await this.search_input.press("Enter")
   }
 }
--- a/playwright/pages/search_page.ts
+++ b/playwright/pages/search_page.ts
@@ -32,11 +32,21 @@ export class SearchPage {
   }
 
   get searchResultsHeading() {
+    // Escape special regex characters in searchterm
+    const escapedSearchterm = this.searchterm.replace(
+      /[.*+?^${}()|[\]\\]/g,
+      "\\$&"
+    )
+
+    // Query searches use different heading format: "for query: <syntax>"
+    // Standard searches use: "for <SearchType> "<searchterm>""
+    const headingPattern =
+      this.searchType.toLowerCase() === "query"
+        ? `Displaying (\\d+-\\d+|\\d+) of (over )?\\d{1,3}(,\\d{3})* results for query: ${escapedSearchterm}`
+        : `Displaying (\\d+-\\d+|\\d+) of (over )?\\d{1,3}(,\\d{3})* results for ${this.searchType}s? "${escapedSearchterm}"`
+
     return this.page.getByRole("heading", {
-      name: new RegExp(
-        `Displaying (\\d+-\\d+|\\d+) of (over )?\\d{1,3}(,\\d{3})* results for ${this.searchType}s? "${this.searchterm}"`,
-        "i"
-      ),
+      name: new RegExp(headingPattern, "i"),
     })
   }
 
--- a/playwright/tests/search/search_nyql.spec.ts
+++ b/playwright/tests/search/search_nyql.spec.ts
@@ -1,0 +1,46 @@
+import { test, expect } from "@playwright/test"
+import { SearchPage } from "../../pages/search_page"
+
+let searchPage: SearchPage
+const queryBaseUrl =
+  process.env.PLAYWRIGHT_QUERY_BASE_URL ??
+  "https://train-research-catalog.nypl.org/research/research-catalog"
+
+test.beforeEach(async ({ page }) => {
+  searchPage = new SearchPage(page, "")
+  await page.goto(queryBaseUrl)
+})
+
+test.describe("Query Search", () => {
+  test("Do a query search and assert that it returns results", async () => {
+    await searchPage.searchFor('author = "Meillassoux, Quentin"', "Query")
+
+    await expect(searchPage.searchResultsHeading).toBeVisible({
+      timeout: 15000,
+    })
+    // print the titles of the first 5 search results to the console
+    const titles = await searchPage.searchResults.allTextContents()
+    console.log("Search results titles:")
+    for (const title of titles.slice(0, 5)) {
+      console.log(title)
+    }
+
+    // now go to the advanced search page and do an Author/Contributor search and assert the results are the same as the query search
+    await searchPage.page.goto(queryBaseUrl + "/search/advanced")
+    // enter "Meillassoux, Quentin" into the Author/Contributor textbox
+    await searchPage.page
+      .getByRole("textbox", { name: "Author/Contributor" })
+      .fill("Meillassoux, Quentin")
+    // click the search button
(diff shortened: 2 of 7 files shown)


playwright/tests/search/search_nyql.spec.ts
```diff
@@ -0,0 +1,245 @@
+import { test, expect } from "@playwright/test"
+import { SearchPage } from "../../pages/search_page"
+
+let searchPage: SearchPage
+
+test.beforeEach(async ({ page }) => {
+  // No console listener is needed here; Playwright does not fail tests on console hydration warnings by default.
```
7emansell, line 7: please clean up unnecessary comments

playwright/pages/search_page.ts
```diff
@@ -19,10 +19,7 @@ export class SearchPage {
 
     this.search_dropdown = page.getByLabel("Select a category")
     this.search_input = page.getByRole("textbox")
-    this.search_submit_button = page.getByRole("button", {
-      name: "Search",
-      exact: true,
-    })
+    this.search_submit_button = page.locator(".ds-searchBar-button")
```
7emansell, line 22: Is there a reason why you had to switch to this locator? The pattern with the others seems more flexible– a DS update or renaming the IDs could break this

playwright/tests/search/search_nyql.spec.ts
```diff
@@ -0,0 +1,245 @@
...
+    await searchPage.searchFor('author = "Meillassoux, Quentin"', "Query")
+
+    await expect(searchPage.searchResults.first()).toBeVisible({
+      timeout: 15000,
+    })
+    const queryTitles = (await searchPage.searchResults.allInnerTexts())
+      .map((t) => t.trim())
+      .filter(Boolean)
+
+    // now go to the advanced search page and do an Author/Contributor search and assert the results are the same as the query search
```
7emansell, line 29: not opposed necessarily but why do these tests check their results against advanced search but the other search scope tests don't? Also, want to make sure these search terms were purposely selected.. ideally we'd test against ones we can reasonably expect not to change too much (for example, no living authors)

playwright/tests/search/search_nyql.spec.ts
```diff
@@ -0,0 +1,245 @@
...
+      await expect(searchPage.searchResults.first()).toBeVisible({
+        timeout: 15000,
+      })
+      const advancedTitles = (await searchPage.searchResults.allInnerTexts())
+        .map((t) => t.trim())
+        .filter(Boolean)
+
+      expect(advancedTitles).toHaveLength(queryTitles.length)
+      expect([...advancedTitles].sort()).toEqual([...queryTitles].sort())
+      // if exact same order is expected, use this instead:
```
7emansell, line 118: Do we anticipate requiring exact same order? If not, can remove this comment

playwright/tests/search/search_unique_identifier.spec.ts
```diff
@@ -15,8 +16,10 @@ test.describe("Unique Identifier Search", () => {
   }) => {
     await searchPage.searchFor(searchterm, "Unique identifier")
     await expect(searchPage.searchResultsHeading).toBeVisible()
-    // click on the title of the first search result
-    await searchPage.searchResults.first().click()
+    // navigate to the first search result via its href to avoid portal overlay
```
7emansell, line 19: Is this only an issue on this test?

playwright/tests/search/search_unique_identifier.spec.ts
```diff
@@ -15,8 +16,10 @@ test.describe("Unique Identifier Search", () => {
...
+    const href = await searchPage.searchResults.first().getAttribute("href")
+    expect(href, "Expected first search result to have an href").toBeTruthy()
+    await page.goto(href!)
```
7emansell, line 22: Does this need the non-null assertion? I think the test will fail on the previous line if if no href

7emansell (changes requested):
looks great just a couple questions

playwright/pages/search_page.ts
samanthaandrews: Also curious about the switch here!

playwright/tests/search/search_nyql.spec.ts
```diff
@@ -0,0 +1,245 @@
...
+    })
+    const queryTitles = (await searchPage.searchResults.allInnerTexts())
+      .map((t) => t.trim())
+      .filter(Boolean)
+
+    await searchPage.page.goto("search/advanced")
+    await searchPage.page
+      .getByRole("textbox", { name: "Keyword" })
+      .fill("pterosaur")
+    // click the search button
```
samanthaandrews, line 67: this comment can probably be removed; seems easy to understand from the code

playwright/tests/search/search_nyql.spec.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    // now go to the advanced search page and do an Author/Contributor search and assert the results are the same as the query search
+    await searchPage.page.goto("search/advanced")
+    await searchPage.page
+      .getByRole("textbox", { name: "Author/Contributor" })
+      .fill("Meillassoux, Quentin")
+    await searchPage.page
+      .locator("form")
+      .locator("#submit-advanced-search")
+      .click()
```
samanthaandrews, line 37: I'm thinking we should move this into the page file instead of the spec file

playwright/tests/search/search_nyql.spec.ts
```diff
@@ -0,0 +1,245 @@
...
+    await searchPage.searchFor('callnumber = "^JFE 24"', "Query")
+
+    await expect(searchPage.searchResults.first()).toBeVisible({
+      timeout: 15000,
+    })
+    const queryTitles = (await searchPage.searchResults.allInnerTexts())
+      .map((t) => t.trim())
+      .filter(Boolean)
+
+    // now go to the advanced search page and do a Title search and assert the results are the same as the query search
```
samanthaandrews, line 172: is this doing a title search? not sure this comment is necessary since the code seems self-explanatory

playwright/tests/search/search_nyql.spec.ts
```diff
@@ -0,0 +1,245 @@
...
+  })
+})
+test.describe('identifier = "b10670401"', () => {
+  test('identifier = "b10670401"', async () => {
+    await searchPage.searchFor('identifier = "b10670401"', "Query")
+
+    await expect(searchPage.searchResults.first()).toBeVisible({
+      timeout: 15000,
+    })
+    // expect 1 result to be returned and the link on the title to contain "b10670401"
```
samanthaandrews, line 203: same with this comment; the code is clear, no comment necessary

playwright/pages/search_page.ts
```diff
@@ -46,6 +43,9 @@ export class SearchPage {
     await this.search_dropdown.selectOption({ label: searchType })
 
     await this.search_input.fill(searchterm)
-    await this.search_submit_button.click()
+
+    // Firefox can intermittently show a portal overlay during hydration.
+    // Submitting with Enter avoids pointer interception on the button.
+    await this.search_input.press("Enter")
```
samanthaandrews, line 49: This comment and new locator are genuinely helpful!

playwright/pages/search_page.ts
samanthaandrews: Actually maybe this locator can be removed entirely? It looks like the click is now happening on the search input (which seems good!) but unsure if this button is necessary anymore

playwright/tests/search/search_nyql.spec.ts
```diff
@@ -0,0 +1,245 @@
...
+      expect([...advancedTitles].sort()).toEqual([...queryTitles].sort())
+      // if exact same order is expected, use this instead:
+      //expect(advancedTitles).toEqual(queryTitles)
+      expect(queryTitles.length).toBeGreaterThan(0)
+    }).toPass({ timeout: 15000 })
+  })
+})
+
+test.describe("callnumber = JFE 86-3252", () => {
+  test("callnumber = JFE 86-3252", async () => {
```
samanthaandrews, line 126: This describe and test name are the same, and I'm not exactly sure what they're testing. Let's update at least one of them to make things clear for future us.

playwright/tests/search/search_nyql.spec.ts
samanthaandrews: Ahh ok actually, upon reading the whole diff, what I'm seeing is a lot of code duplication. It looks like in multiple tests we run a Query search, capture the first-page titles, run the search, then assert the two results match. We could do something like this to reduce duplicate code:

```

const titlesOf = async () =>
  (await searchPage.searchResults.allInnerTexts())
    .map((t) => t.trim())
    .filter(Boolean)

const advancedCases = [
  { query: 'author = "Meillassoux, Quentin"', field: "Author/Contributor", value: "Meillassoux, Quentin", compare: "equal" },
  { query: 'keyword = "pterosaur"',            field: "Keyword",            value: "pterosaur",            compare: "equal" },
  { query: "keyword = pterosaur",              field: "Keyword",            value: "pterosaur",            compare: "equal" },
  { query: 'callnumber = "^JFE 24"',           field: "Call number",        value: "^JFE 24",              compare: "length" },
  { query: 'subject = "Mitanni (Ancient kingdom)"', field: "Subject",       value: "Mitanni (Ancient kingdom)", compare: "length" },
] as const

test.describe("Query Search", () => {
  for (const { query, field, value, compare } of advancedCases) {
    test(query, async () => {
      await searchPage.searchFor(query, "Query")
      await expect(searchPage.searchResults.first()).toBeVisible({ timeout: 15000 })
      const queryTitles = await titlesOf()

      await searchPage.page.goto("search/advanced")
      await searchPage.page.getByRole("textbox", { name: field }).fill(value)
      await searchPage.page.locator("form").locator("#submit-advanced-search").click()

      await expect(async () => {
        await expect(searchPage.searchResults.first()).toBeVisible({ timeout: 15000 })
        const advancedTitles = await titlesOf()

        expect(queryTitles.length).toBeGreaterThan(0)
        expect(advancedTitles).toHave
        if (compare === "equal") {
          expect([...advancedTitles].es].sort())
        }
      }).toPass({ timeout: 15000 })
    })
  }
})
```

samanthaandrews (changes requested):
Thanks for making a PR to cover the NYQL test cases on the frontend. I do think it would be good for you to schedule time with me and Emma to walk us through your strategy here. Verifying that the NYQL queries match Advanced Search results doesn't seem wrong, but I do know that we are performing similar actions in the API tests, so it's possible that we could instead focus these tests on the frontend functionality that isn't covered by the API tests.

The biggest pieces of feedback I have here are:
- Only include comments when the code itself isn't clear. For the most part, the code is clear, which is a good thing! The comments on top of that only add noise.
- We should reduce duplicate code as much as possible and look for ways to pull out the parts that are duplicated into a reusable function/util. In this case, the duplication can likely be addressed by an array of objects that contain the queries, fields, and values and a loop.

chrismulholland:
@7emansell  @samanthaandrews After a discussion with Sam about the length of this PR, we have decided to close this and follow up with smaller, more manageable pull requests.  Thank you for all the feedback thus far; it will be incorporated in the next round.

PR #753, opened 2026-06-22
SCC-5058: Split heading into screenreader and focusable visual element

KaseyW31:
## Ticket:

- JIRA ticket [SCC-5058](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5058)

- Wraps results heading for search/browse in a container with skeleton loading
- Adds hidden aria-live region for reading out updated heading

On the search results page, perform a new search. The results heading should disappear. After results are loaded, the heading should be focused, and the screen reader should read out the updated heading once.
The same thing should be tested for the browse pages.

## Accessibility updates?

Upon performing a new search, the previous results heading should not be on the page. After new search results are loaded, the results heading should be read out once by the screen reader.

### Accessibility checklist

- [x] The feature works with keyboard inputs (tabbing, arrows, space, enter, esc).
- [x] Tested with a screen reader if the feature involves hidden text or `aria-live`.
- [x] Confirmed focus management is correct for UI updates and DOM `ref`s.
- [x] Verified the feature works when zoomed in to 200% and 400%.

### Developer checklist

[SCC-5058]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5058?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Search/Search.tsx
+++ b/src/components/Search/Search.tsx
@@ -34,6 +34,7 @@ import type {
 import type { RCPage } from "../../types/pageTypes"
 import { getBrowseTypeFromPath } from "../../utils/appUtils"
 import { useRouter } from "next/router"
+import a11yStyles from "../../../styles/utils/a11y.module.scss"
 
 interface SearchProps {
   errorStatus?: HTTPStatusCode | null
@@ -103,6 +104,20 @@ const Search = ({
   const displayFilters = !!aggs?.filter((agg: Aggregation) => agg.values.length)
     .length
 
+  const searchResultsHeading = getSearchResultsHeading(
+    searchParams,
+    totalResults,
+
+    slug
+      ? {
+          slug,
+          browseType: resultsType,
+          role,
+        }
+      : undefined,
+    parsedQuery
+  )
+
   const searchResultBibs = mapElementsToSearchResultsBibs(searchResultsElements)
   return (
     <>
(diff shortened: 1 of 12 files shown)


src/components/Search/Search.tsx
```diff
@@ -103,6 +104,20 @@ const Search = ({
   const displayFilters = !!aggs?.filter((agg: Aggregation) => agg.values.length)
     .length
 
+  const searchResultsHeading = getSearchResultsHeading(
+    searchParams,
+    totalResults,
+
+    slug
```
charmingduchess, line 111: this method signature is a little unwieldy with two optional params at the end. not crucial, but it might make things cleaner if all of the params were passed in on an object, with the `browseOptions` spread to top level properties.

src/components/Search/Search.tsx
```diff
@@ -152,33 +167,35 @@ const Search = ({
...
-                  parsedQuery
+                {isLoading ? (
+                  <SkeletonLoader
+                    contentSize={1}
+                    showImage={false}
+                    headingSize={0}
+                  />
+                ) : (
+                  <Heading
+                    ref={searchResultsHeadingRef}
```
charmingduchess, line 189: is this ref being used anywhere?

charmingduchess (approved):
a very small note that has nothing to do with the work you've done here. if there are other changes that need to be made, feel free to add in my recommendation, but not a blocker by any means.

KaseyW31:
@charmingduchess thank you for the review! Sorry I should've closed this PR earlier, I'll be making a new one for a workaround for the ticket as suggested by Clare

PR #761, opened 2026-06-25
Remove span tag from query option

KaseyW31:
## Ticket:

- No ticket

- Removes span tag from "Query" option in search bar, which was a possible cause of a hydration error during playwright testing

- No specific way to test, but playwright tests should hopefully be less flakey sometimes

### Accessibility checklist

### Developer checklist


7emansell (approved):
👍👍

PR #763, opened 2026-06-30
Update header tests for Visit link

7emansell:
## Ticket:

- n/a

- Playwright tests failing bc of link update within the universal header ([update was made here](https://github.com/NYPL/nypl-header-app/pull/85))
- Updates to look for "Visit" link
- I think this mismatch not being caught is a good argument in favor of RC not testing the header at all in Playwright– we don't control its updates and aren't even necessarily aware of them. Will add to the agenda for our upcoming meeting

### Accessibility checklist

### Developer checklist


KaseyW31 (approved):
👍

chrismulholland (approved):
looks great.

PR #764, opened 2026-07-02
SCC-5167: Advanced search page lag

KaseyW31:
## Ticket:

- JIRA ticket [SCC-5167](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5167)

- Creates isolated components for advanced search that manage their own state, so that fields are not rerendered unnecessarily on input to other fields
- Removed existing search reducer
- Modifies `DateFilter` (+`useDateFilter`) so that it can manage its own state (--> modification to the `DateFilter` component in `SearchResults`)
- `onSubmit`: Date info and text inputs are retrieved using `FormData`; filters are pulled from a global ref that's updated when filters change

This PR is pretty large and just one approach to the performance issue, also there is probably a lot of room for improvement/refactoring. I am very open to suggestions

Also another note- a lot of rendering and some of the lag is cut down, but there is still a ~1s delay (in dev environment) when clicking into an input and the first few keystrokes, not entirely sure of the root cause... especially because the `SearchResults` page also rerenders a bunch of components and doesn't have the same lag

- Verify that there is some reduced lag on the advanced search page when typing inputs and selecting filters
- Verify no change to existing functionality
- Verify that there is no change to the DateFilter behavior on search results page either

### Accessibility checklist

### Developer checklist

[SCC-5375]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SCC-5167]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KaseyW31:
Realizing that this is probably better tied to [SCC-5167](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5167), though I haven't yet implemented a loading state after clicking submit button. Also working on the design system side so that PR can be for [SCC-5375](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5375)

[SCC-5167]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SCC-5375]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

pages/search/advanced.tsx
```diff
@@ -72,73 +60,46 @@ export default function AdvancedSearch({
...
+    dateFrom: "",
+    changeHandler: globalInputChangeHandler,
+  })
+  const { dateError } = dateFilterProps
 
   const handleSubmit = async (e: SyntheticEvent) => {
     e.preventDefault()
-    flushSync(() => dateFilterProps.onBlur())
-    const errors = dateFilterProps.onApply()
+    const formData = new FormData(e.target as HTMLFormElement)
```
charmingduchess: where is FormData defined?

pages/search/advanced.tsx
```diff
@@ -72,73 +60,46 @@ export default function AdvancedSearch({
...
+    const submittedDates = {
```
charmingduchess, line 107: can you put some comments describing what's going on here with the dates lines 96-102?

pages/search/advanced.tsx
```diff
@@ -72,73 +60,46 @@ export default function AdvancedSearch({
...
-    setAlert(false)
-    if (value === null) {
-      dispatch({
-        type: "filter_change",
-        field,
-        payload: [],
-      })
-      return
-    }
+  const filterValuesRef = useRef<Record<string, string[]>>({
```
charmingduchess: It might be useful for future devs to have a comment here explaining the choice to use a ref instead of react state

pages/search/advanced.tsx
```diff
@@ -167,8 +138,13 @@ export default function AdvancedSearch({
...
     setAlert(false)
-    clearDateInputs()
-    dispatch({ type: "form_reset", payload: initialSearchFormState })
+    filterValuesRef.current = {
+      language: [],
+      format: [],
+      buildingLocation: [],
+      collection: [],
+    }
+    setResetKey((k) => k + 1)
```
charmingduchess, line 142: does the key need to be an incrementing integer? a boolean might be clearer if it's just to trigger a useEffect (or to trigger a rerender if that useeffect turns out to be unnecessary)

src/components/AdvancedSearch/MultiSelectMemo.tsx
```diff
@@ -0,0 +1,62 @@
...
+    fieldValue,
+    label,
+    options,
+    onSelectionChange,
+    resetKey,
+    globalInputChangeHandler,
+  }: MultiSelectMemoProps) => {
+    const [selected, setSelected] = useState<string[]>([])
+
+    useEffect(() => {
```
charmingduchess: check out this react.dev resource. [you might not need a useEffect](https://react.dev/learn/you-might-not-need-an-effect) to make this update. both `resetKey` and `selected` are react state. useEffect is best left for situations where we're waiting for updates from things outside of react state.

charmingduchess: on second look, i feel like resetKey updating at all would appropriately trigger a rerender?

charmingduchess (changes requested):
this is working great. i am wondering if you might be able to isolate further what's actually making the difference here. i have a feeling that the memos might not be necessary now that the state of what is selected is totally abstracted into the ref.

src/components/AdvancedSearch/MultiSelectMemo.tsx
charmingduchess: [this specific section](https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes) of that link above i think echoes what i was saying

KaseyW31: That was helpful thank you! I ended up using [this pattern](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)

pages/search/advanced.tsx
KaseyW31: Would a boolean also be confusing? I was thinking that there would be no meaning behind a true or false value, but on second thought it could work as a toggle/switch?

pages/search/advanced.tsx
KaseyW31: I realized the flushSync is no longer necessary, I added a general comment but let me know if I should expand more on it

pages/search/advanced.tsx
KaseyW31: It comes with native React I believe

pages/search/advanced.tsx
charmingduchess: Let's be a little more explicit and describe the pattern and how the consuming components interact with the ref values.

pages/search/advanced.tsx
```diff
@@ -90,15 +85,20 @@ export default function AdvancedSearch({
...
 
+  useEffect(() => {
+    if (alert && !Object.keys(dateError || {}).length) {
+      setPersistentFocus(idConstants.advancedSearchError)
+    }
+  }, [alert, dateError, setPersistentFocus])
+
   const handleSubmit = async (e: SyntheticEvent) => {
     e.preventDefault()
+    // Get inputs from isolated components
```
charmingduchess: What is an isolated component?

charmingduchess (commented):
I think we could use a little more commentary since we're adopting a new pattern here. The new components should have a summary at the top of the file (like [this example](https://github.com/NYPL/research-catalog/blob/6c948fe3a88b48bbd6e36b8215e7c21d02faabd9/pages/browse/[[...browseType]].tsx#L50)). Be sure to define what is meant by isolated components as that might not be clear to a future reader.

src/components/AdvancedSearch/MultiSelectMemo.tsx
KaseyW31: Actually I think I can do [this](https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes), but since the components are in a list, I can make the key a combination of the name and reset key

pages/search/advanced.tsx
charmingduchess: I think it can just be a toggle

pages/search/advanced.tsx
KaseyW31: That makes sense. Let me know if the comments I added are a good amount and in the right place!

charmingduchess (approved):
Code looks good. Since we're departing from pure react state controlled components, I think it's important to be very clear in the comments how our approach works. How I see this working is: The isolated components operate as the local state for each input, and they are responsible for updating the form state ref. The form state ref acts as the global form state. It is write-only from the point of view of the isolated components, and is only read from when we are submitting the final form. 

Does that sound like an accurate representation of whats going on?

If so, could you try and include that concept, possibly around line 70 in `pages/search/advanced.tsx`?

KaseyW31:
@charmingduchess Yes that was essentially accurate, though originally the ref was only for MultiSelect filters and not text inputs. But actually I think having the ref be for everything is more understandable. I just pushed that change and tweaked some comments, could you take a look at that?

charmingduchess (approved):
looks good to me!

7emansell:
will review more thoroughly later today, but let's not merge this in until this work includes your Multiselect from the DS @KaseyW31

src/components/AdvancedSearch/DivisionSelect.tsx
7emansell, line 1: To be consistent with the other two, can this be renamed to `IsolatedMultiSelectWithGroupTitles`? lengthy but i think its clearer

src/reducers/searchFormReducer.ts
7emansell, line 1: yayyy

src/hooks/useDateFilter.tsx
7emansell, line 1: Realizing (again) how complicated this hook is– these updates are really nicely done/named

7emansell (commented):
Looks good! and feels a little faster on Vercel to me. Not seeing a speed up on the unit tests locally but maybe the new `Multiselect` will be the key there, will re-review then!!

src/components/AdvancedSearch/DivisionSelect.tsx
KaseyW31: I did that and then realized that I could merge the two of them, lmk if you think that works! If not I can just revert a commit

src/components/AdvancedSearch/IsolatedMultiSelect.tsx
7emansell, line 1: Nice

PR #765, opened 2026-07-06
SCC-5486: Correctly authenticate lists routes

7emansell:
## Ticket:

- JIRA ticket [SCC-5486](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5486)

- Updates API handlers to use logged in patron ID
- Stops lists requests passing patron ID in params or in body
- Major oversight on my part so looking to release this ASAP before lists are hacked to death. Rest of backend routes look okay (all the existing account endpoints follow this pattern)

### Accessibility checklist

### Developer checklist

[SCC-5486]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pages/api/account/lists/list.ts
+++ b/pages/api/account/lists/list.ts
@@ -6,33 +6,29 @@ import {
   updateList,
 } from "../../../../src/server/api/lists"
 import MyAccount from "../../../../src/models/MyAccount"
+import initializePatronTokenAuth from "../../../../src/server/auth"
 
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse
 ) {
-  const { listId } = req.query
-
-  const {
-    description,
-    listName,
-    records,
-    patronId: bodyPatronId,
-  } = req.body || {}
-
-  const patronId = req.query.patronId || bodyPatronId
-  if (!patronId || typeof patronId !== "string") {
-    return res.status(400).json({ error: "Missing or invalid patronId" })
+  const patronTokenResponse = await initializePatronTokenAuth(req.cookies)
+  const cookiePatronId = patronTokenResponse.decodedPatron?.sub
+  if (!cookiePatronId) {
+    return res.status(403).json({ error: "No authenticated patron" })
   }
 
-  const accountModel = new MyAccount(null, patronId)
+  const { listId } = req.query
+
+  const { description, listName, records } = req.body || {}
+  const accountModel = new MyAccount(null, cookiePatronId)
 
   if (req.method == "GET") {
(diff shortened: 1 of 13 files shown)


nonword (approved):
Looks right!

KaseyW31 (approved):
Looks good to me!

pages/api/account/lists/list.ts
```diff
@@ -6,33 +6,29 @@ import {
...
-    records,
-    patronId: bodyPatronId,
-  } = req.body || {}
-
-  const patronId = req.query.patronId || bodyPatronId
-  if (!patronId || typeof patronId !== "string") {
-    return res.status(400).json({ error: "Missing or invalid patronId" })
+  const patronTokenResponse = await initializePatronTokenAuth(req.cookies)
+  const cookiePatronId = patronTokenResponse.decodedPatron?.sub
+  if (!cookiePatronId) {
```
charmingduchess, line 17: lets copy the logic from the [hold request page](https://github.com/NYPL/research-catalog/blob/main/pages/api/hold/request/%5Bid%5D/index.ts#L28)
ie ` if (!cookiePatronId || cookiePatronId !== patronId)`

charmingduchess (changes requested):

pages/api/account/lists/list.ts
7emansell: Considered this but not actually sure why we check the passed patron id at all on those endpoints– like why would we ever care what the frontend thinks the patron id is when the server already knows the id thats authorized to make the request?

charmingduchess: I think there's some race condition where a user might be logged in on one tab and not another? @nonword do you have insight into this?

charmingduchess (approved):

pages/api/account/lists/list.ts
nonword: Ah right. That's a good point. We want to mitigate the back-end returning patron data for a different patron than the front-end expects. That could come up in rare cases in between user sessions where the patron logs in with multiple accounts. Let's discuss that hardening later. I don't think it should hold up this PR

7emansell: Oops just saw this– ticket here so i dont drop this https://newyorkpubliclibrary.atlassian.net/browse/SCC-5491 but yea will go ahead and release this today

PR #768, opened 2026-07-14
SCC-5167: Advanced search form loading state

KaseyW31:
## Ticket:

- JIRA ticket [SCC-5167](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5167)

- Adds a global style for components that can be disabled, which is used for the loading state in advanced search form
- Also updates hold request and search results filters to use this style

- Verify loading state after submitting advanced search form
- Verify that loading state in hold request and search results filters is (pretty much) unchanged

### Accessibility checklist

### Developer checklist

[SCC-5167]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Updated
 
-- Updated advanced search page to use refs and isolated components [SCC-5167](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5167)
+- Improved advanced search page performance and added loading state [SCC-5167](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5167)
 
 ## [2.1.2] 2026-07-07
 
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -9,6 +9,7 @@ import { FeedbackProvider } from "../src/context/FeedbackContext"
 import { FocusProvider } from "../src/context/FocusContext"
 import { BrowseProvider } from "../src/context/BrowseContext"
 import { useRouter } from "next/router"
+import "../public/styles/globals.css"
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
 function App({ Component, pageProps }) {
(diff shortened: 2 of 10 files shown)


charmingduchess (changes requested):
I think this is a great solution, but not exactly in the style that loading state is indicated for most of the app. We historically use the SkeletonLoader component (which replaces the whole page more or less) and/or loading text and animation in the [submit button](https://github.com/NYPL/research-catalog/blob/6c948fe3a88b48bbd6e36b8215e7c21d02faabd9/src/components/MyAccount/RequestsTab/FreezeButton.tsx#L112). Sorry the ticket was so vague on this!

Also, for the future, all of the individual form components have a built in disabled state (though not the Form itself, maybe worth a DS update!), that is preferable to creating a new style.

7emansell (changes requested):
partially agree with Vera– i think replacing the whole form with a skeleton loader seems like overkill, i like the opacity change, but the submit/cancel buttons maybe should be replaced by the skeleton loader?

KaseyW31:
Thanks sounds good, I can confirm with Apoorva again about what she thinks about the SkeletonLoader.
@charmingduchess I don't think that the MultiSelect has a disabled state, but I can try adding it to that and the Form components. Do you think the style is worth keeping in the meantime?

7emansell:
To be clear, should apply a loading state over the whole form, and (probably) individually disable each component, but  should _not_ go into the DS to add a `isDisabled` prop to the Form or Multiselect

KaseyW31:
Apoorva said the preference based on form submission best practices is:
1. Submit button goes into a loading animation, cancel button becomes disabled
2. All form fields become disabled

That's what I have here using the DS updates, hopefully looks better now?

KaseyW31:
Actually I realized the search results filter panels get closed when one is modified because of my change in DS - will revert back to the style approach for now. Maybe a future change can be made in DS for keeping the panels open in a disabled state?

7emansell:
> Actually I realized the search results filter panels get closed when one is modified because of my change in DS - will revert back to the style approach for now. Maybe a future change can be made in DS for keeping the panels open in a disabled state?

Ah good eye. Yea this is a bummer but i also feel like it makes sense that the panels close on disablement. the styling workaround is good for now

PR #769, opened 2026-07-20
SCC-5253: Special collections messaging

7emansell:
## Ticket:

- JIRA ticket [SCC-5253](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5253)

** Final PR in status messaging epic (in other words, if any AC is missing from [these designs](https://www.figma.com/design/WpOWDgvUYeIsiq0XTzrR2g/RC--LSP-Miscellaneous?node-id=3516-16736&m=dev), it shouldn't be) ** 
- Rewrites availability keys and their corresponding text (`ItemAvailability` model and component)
- Removes a bunch of item location filter processing– now using the much simpler `buildingLocationId` on DiscoAPI, so no need to dedupe offsite locations, etc.
- Updates search result card for bibs with no items
- Makes it possible to pass bib metadata to the feedback form, instead of just item metadata (for bibs with no items)

** Also note that we are updating some desk/shelf types and reference room labels in Core, so those switches are not reflected here. Doesn't factor into this code

### Accessibility checklist

- [x] The feature works with keyboard inputs (tabbing, arrows, space, enter, esc).
- [x] Verified the feature works when zoomed in to 200% and 400%.

### Developer checklist

[SCC-5253]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/ItemTable/ItemAvailability/ContactALibrarian.tsx
```diff
@@ -1,39 +1,49 @@
...
         onClick={() =>
-          onContact({
-            id: item.id,
-            barcode: item.barcode,
-            callNumber: item.callNumber,
-            volume: item.volume,
-            bibId: item.bibId,
-          })
+          // If item's available, pass that metadata
+          onContact(
```
charmingduchess, line 28: is this ternary necessary? can you populate the same object with potentially null  values instead?

src/models/ItemAvailability.ts
charmingduchess, line 1: I'm not totally sure that this is what you were referring to in our slack convo (re: "cases that reflect the actual messages being displayed and cases that cover all actual types of records"), but I think the approach should be having every kind of record we know about covered, even if that means returning the same availability type multiple times. if we have every kind of record outlined separately, it will be easier to debug and also to update in the future.

charmingduchess: Mentioned offline too, but wanted to flag here that I don't see coverage for offsite special collections here (like [harry belafonte papers](https://www.nypl.org/research/research-catalog/bib/b22783213))

src/models/ItemFilterData.ts
```diff
@@ -48,37 +47,10 @@ export class LocationFilterData extends ItemFilterData {
     super(aggregation)
   }
 
-  // Override display options to add Offsite if there are any Recap locations
   displayOptions(): FilterCheckbox[] {
```
charmingduchess, line 50: glad to see this go. seems like we were already doing the thing that we are now avoiding with the indexed building locations

src/config/constants.tsx
```diff
@@ -208,26 +208,32 @@ export const DISPLAY_LINKED_FIELD_MAPPING: Record<
...
 }
 
+/* Availability keys describe an availability status, and correspond to the message displayed
+ ** below the request buttons on an item. Note: Many of these do not display a
+ ** message (see the corresponding component). */
 export const AVAILABILITY_KEYS = {
-  // anything not covered by the cases below is EDGE_CASE
-  EDGE_CASE: "edgeCase",
-  // there is only one not available case, so availability is assumed as the default
+  // Not available (all NYPL), if we know due date it will display
```
charmingduchess: and partner?

src/config/constants.tsx
```diff
@@ -208,26 +208,32 @@ export const DISPLAY_LINKED_FIELD_MAPPING: Record<
...
-  ONSITE_NO_FINDING_AID_NO_AEON: "noFindingAidNoAeonOnsite",
-  RECAP_NO_FINDING_AID_NO_AEON: "noFindingAidNoAeonRecap",
+  // Not available, partner item
+  NOT_AVAILABLE_PARTNER: "notAvailablePartner",
+  // Available, general desk reference
+  AVAILABLE_DESK: "availableDeskGeneral",
+  // Available, general shelf reference
+  AVAILABLE_SHELF: "availableShelfGeneral",
+  // Available onsite, special collections, needs appointment, no Aeon link
+  AVAILABLE_ONSITE_APPT: "availableOnsiteApptNoAeon",
```
charmingduchess: if there is a complementary type, ie AVAILABLE_ONSITE_APPT/AVAILABLE_ONSITE_APPT_AEON, can you include the NO_AEON distinction explicitly in the key?

src/config/constants.tsx
```diff
@@ -208,26 +208,32 @@ export const DISPLAY_LINKED_FIELD_MAPPING: Record<
...
-  RECAP_AEON: "RecapAeon",
-  AEON: "Aeon",
-  ONSITE_AEON: "OnsiteAeon",
-  ONSITE_AEON_FINDING_AID: "onsiteAeonFindingAid",
-  RECAP_AEON_FINDING_AID: "recapAeonFindingAid",
-  ONSITE_FINDING_AID: "onsiteFindingAid",
-  RECAP_FINDING_AID: "recapFindingAid",
-  ONSITE_NO_FINDING_AID_NO_AEON: "noFindingAidNoAeonOnsite",
-  RECAP_NO_FINDING_AID_NO_AEON: "noFindingAidNoAeonRecap",
+  // Not available, partner item
```
charmingduchess: not super crucial, but you could also just make these comments be the strings assigned to the constants, just to streamline a little. there's no real reason these strings need to be camel cased.

src/components/ItemTable/RequestButtons.tsx
```diff
@@ -8,11 +8,20 @@ interface RequestButtonsProps {
...
 
 /**
- * The StatusLinks component appears in the Item Table
+ * The AvailabilityLinks component appears in the ItemTable
  * TODO: Pass search keywords to links as ?searchKeywords=${"TODO"}
  */
 const RequestButtons = ({ item }: RequestButtonsProps) => {
   if (item.allLocationsClosed) return null
+
+  if (
```
charmingduchess, line 17: this state could be an additional availability type

src/components/ItemTable/ItemAvailability.tsx
```diff
@@ -1,161 +1,80 @@
...
-      message = <AvailableOnsite location={item.location} />
-      break
-    case NOT_AVAILABLE:
-      // Closed stacks, offsite NYPL should have due date
-      message = <NotAvailable dueDate={item.dueDate} />
-      break
+    // No message displays for these cases
+    case AVAILABLE_ONSITE_APPT_AEON:
+    case AVAILABLE_CLOSED_STACK:
+    case AVAILABLE_GENERAL:
```
charmingduchess: maybe add a `return null` here just to be super explicit?

src/components/ItemTable/ItemAvailability.tsx
```diff
@@ -1,161 +1,80 @@
...
-        <Link
-          isExternal
-          href={appConfig.urls.researchMaterialsHelp}
-          fontSize="sm"
-          className="no-print"
-        >
-          How do I pick up this item and when will it be ready?
-        </Link>
-      )
-    case EDGE_CASE:
```
charmingduchess, line 82: what happens now if there is an item that we don't have an availability key for?

src/models/Item.ts
```diff
@@ -69,9 +70,13 @@ export default class Item {
       isReCAP: this.isReCAP,
       aeonUrl: this.aeonUrl,
       collectionAccessType: this.getCollectionAccessTypeFromItem(item),
-      findingAid: bib.findingAid,
+      hasBarcode: !!this.barcode,
+      isPartnerReCAP: this.isPartnerReCAP(),
     })
     this.collection = item.collection ? item.collection[0] : null
+    this.buildingLocation = item.buildingLocation
```
charmingduchess: can you convert this to use optional chaining with an `||`?

src/models/Item.ts
```diff
@@ -102,8 +107,9 @@ export default class Item {
 
   // Pre-processing logic for setting Item holding location
   getLocationFromItem(item: DiscoveryItemResult): ItemLocation {
-    let location = defaultNYPLLocation
-    if (this.isPartnerReCAP) location = partnerDefaultLocation
+    let location = { ...defaultNYPLLocation }
+
+    if (this.isPartnerReCAP()) location = { ...partnerDefaultLocation }
```
charmingduchess, line 110: why is a spread operator necessary here?

src/models/ItemTableData.ts
```diff
@@ -33,25 +33,27 @@ export default class ItemTableData {
    */
   getTable(): { [key: string]: ReactElement[] } {
     const callNumberCells = this.items?.map((item) =>
-      ItemTableCell({
+      React.createElement(ItemTableCell, {
```
charmingduchess, line 36: why this change? since we don't use this pattern often, could use a comment explaining the choice to use `createElement`

src/models/SearchResultsBib.ts
```diff
@@ -10,12 +12,17 @@ import { ITEMS_PER_SEARCH_RESULT } from "../config/constants"
...
+  callNumber?: string
+  collection?: Collection
 
   constructor(result: DiscoveryBibResult) {
     super(result)
     this.publicationStatement = result.publicationStatement?.length
       ? result.publicationStatement[0]
       : null
+    // Potential bib level fields to check if bib has no items:
+    this.callNumber = result.shelfMark?.[0] || null
```
charmingduchess: do we want these to be null if there *are* items? could help to avoid mysterious bib callnumbers populating instead of item call numbers down the line

charmingduchess (changes requested):

src/components/ItemTable/ItemAvailability/ContactALibrarian.tsx
7emansell: for type strictness no.. plus i still want it to go ahead and open the feedback form even if there's no metadata to set

src/models/ItemAvailability.ts
7emansell: Discussed offline, flagged to bray

src/config/constants.tsx
7emansell: wdym?

src/config/constants.tsx
7emansell: added

src/components/ItemTable/RequestButtons.tsx
7emansell: availability types don't apply to the request buttons, and as it is rn, ItemAvailability doesn't consider whether the item is phys/edd requestable

src/components/ItemTable/ItemAvailability.tsx
7emansell: discussed offline, falls into available special/general and displays nothing

src/models/Item.ts
7emansell: Esp now that it could be changing the location label, seemed safer to reference the copy than change the actual location on the item ?

src/models/SearchResultsBib.ts
7emansell: Wise

src/models/ItemTableData.ts
7emansell: Necessary bc it's a .ts file and im mixing react rendering/ data model... Which this was already doing unfortunately but now it's getting worse. Could we accept me adding a todo/refactor ticket for the item table

src/config/constants.tsx
charmingduchess: sorry i must have missed that the next key has _PARTNER. can you add _NYPL to this key so it is clear without looking at the other keys?

7emansell: gotcha

src/models/ItemTableData.ts
7emansell: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5515

charmingduchess (approved):
🔥 thanks for the updates! looks great

src/components/SearchResults/SearchResultItems.tsx
```diff
@@ -1,9 +1,16 @@
 import { Box, Text } from "@nypl/design-system-react-components"
 import type ItemTableData from "../../models/ItemTableData"
-import StatusLinks from "../ItemTable/AvailabilityLinks"
+import AvailabilityLinks from "../ItemTable/AvailabilityLinks"
 
 interface SearchResultItemsProps {
-  itemTableData: ItemTableData
+  itemTableData:
+    | ItemTableData
+    | {
```
KaseyW31: Would it make sense to define a new interface for this second type just to make it clearer? Maybe something like `NoItemsBibTableData` which I'm assuming this is used for?

KaseyW31 (approved):
Looks good to me! Just the one question but it's pretty minor

src/components/SearchResults/SearchResultItems.tsx
7emansell: Yes absolutely good call out. Added interface

PR #770, opened 2026-07-21
SCC-5485: List request failure banner

KaseyW31:
## Ticket:

- JIRA ticket [SCC-5485](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5485)

- Add error banner to replace the text content of the lists menu if the request to fetch lists fails

- Verify error banner shows up when no lists are fetched

### Accessibility checklist

### Developer checklist

[SCC-5485]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


7emansell (changes requested):
Can the banner be moved to the top level of the menu? The "No lists found" is still a relevant message (for example, if you search for a list that you don't have) and if the lists haven't been fetched, you shouldn't be able to create a list or really do anything at all. 
Sort of hacky but because we always create a default list [here](https://github.com/NYPL/research-catalog/blob/4a4f2e7c82b754fff2387e89931ae4097f273942/src/models/MyAccount.ts#L101), if `lists = []` then you know the fetch has failed– everyone has at least 1 list. So you can use `lists.length > 0?` (probably just under the `Box` L324).

KaseyW31:
@7emansell Makes sense! Are these new changes what you were envisioning and does this approach make sense?

7emansell (approved):
Beautiful this is exactly what i was thinking. Extra props for adding it into the whole status banner/focus system.
plus tested with a fresh user and it worked great (correctly created the default list before hitting this failure)

src/components/List/ManageBibInListMenu.tsx
```diff
@@ -337,137 +337,156 @@ export const ManageBibInListMenu = ({
...
-                          border: "0",
-                          clip: "rect(0, 0, 0, 0)",
-                          height: "1px",
-                          margin: "-1px",
-                          overflow: "hidden",
-                          padding: "0",
-                          position: "absolute",
-                          whiteSpace: "nowrap",
-                          width: "1px",
+          {lists.length > 0 ? (
```
charmingduchess: this ternary is pretty long. can you flip the logic so we can unnest it? if you start with lists.length === 0 && {return vaue}, that should eliminate at least the outer ternary. I think it would also make this more legibile if the showCreateForm component was

src/components/List/ManageBibInListMenu.tsx
charmingduchess, line 272: @7emansell sorry this is totally unrelated to this PR, but just noticed this while looking at these changes. are the settimeouts in this file leftover from before we added a timeout to the focus context? or are they still necessary? if so, would updating the timeout in the context address the issue? its a flakey fix so i'd like to keep it contained to one source if possible

charmingduchess (changes requested):

src/components/List/ManageBibInListMenu.tsx
7emansell: i would have sworn on my life that these were necessary to override the Chakra `Drawer` focus trap behavior (on mobile– should have commented to explain) but I went back in to test and focus wasn't working at all on mobile 🫠 [PR here to resolve](https://github.com/NYPL/research-catalog/pull/773) but i think this is going to be more involved than i'd like since the `Drawer` wants refs and the delay isn't helping

src/components/List/ManageBibInListMenu.tsx
KaseyW31: Made separate functions for both ternaries, does that look better?

charmingduchess: oops sorry i think my comment above got cut off! The last sentence should have been "I think it would also make this more legibile if the showCreateForm component was its own component". If you want to extract the components like you are doing here, I think the most kosher way to do that is to have them be fully extracted as separate components.

charmingduchess: one last note, which is that in general that if theres a boolean condition that can result in a component not rendering, we want to include that in the parent component. 
it's easier to read
`lists.length === 0 ? <CreateListForm/>`
than just see `<CreateListForm/>` and have to click into the component to see that the component might return null

charmingduchess: check out the style [guide here](https://github.com/NYPL/research-catalog/blob/57bd24a4a68946f6e81a220aaa7847ab965471fc/docs/DEVELOPER_GUIDE.md#L246) for more explanation of my last comment. also, let's pull these components out into a different file, since it does not seem idiomatic to have a component defined inside another component. i could see state/rerendering getting wonky like that

KaseyW31: Yeah I was going to say that it makes sense, though moving the components fully out into separate files would require some code duplication and a good amount of props (`showCreateForm, listName, setListName, debouncedListName, isSubmitting, setShowCreateForm, initialFocusRef, setListCreationStatus`...) because they're pretty tied to the state management in `ManageBibInListMenu`. Do you think that would be worth it? (If not I'll revert the components back to functions)
And thanks for noting that, it's good to keep in mind. Though in this case neither component would render null I believe, there's a fallback case for both

7emansell: ^^ i agree with Kasey here– this component is pretty complex but i think splitting up all its child components will make it even less readable

charmingduchess: Cool sounds good. I'd say lets revert to how it was before, but just use early returns instead of the nested ternary.

KaseyW31: Is this what you had in mind? I could also do the same for the `showCreateForm` ternary but that feels a bit like too many conditionals

charmingduchess (approved):

PR #775, opened 2026-07-31
SCC-5441: Drag-select RTL remove leading newline/tab

KaseyW31:
## Ticket:

- JIRA ticket [SCC-5441](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5441)

- Adds a copy handler to search results table and item table. If a copy selection spans multiple table cells but contains text from only one, extra newlines/tabs injected by the browser are removed
~~On desktop, adds left padding to the search results item table data cells and decreases the width of the row heading cells to compensate. Hopefully the padding of 30px is enough?~~

- Perform a search, drag-select a shelfmark from right to left, copy-paste into a text editor, verify no leading whitespace or newline

### Accessibility checklist

### Developer checklist

[SCC-5441]: https://newyorkpubliclibrary.atlassian.net/browse/SCC-5441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/ItemTable/ItemTable.tsx
+++ b/src/components/ItemTable/ItemTable.tsx
@@ -1,6 +1,7 @@
 import { Box, Table } from "@nypl/design-system-react-components"
 import type ItemTableData from "../../models/ItemTableData"
 import styles from "../../../styles/components/ItemTable.module.scss"
+import { handleTableCopy } from "../../utils/appUtils"
 
 interface ItemTableProps {
   itemTableData: ItemTableData
@@ -16,6 +17,7 @@ const ItemTable = ({ itemTableData }: ItemTableProps) => {
     // Display as grid to prevent bug where the outer container stretches to the Table's width on mobile
     <Box display="grid">
       <Table
+        onCopy={(e) => handleTableCopy(e, "\t")}
         className={`${styles.itemTable} itemTable`}
         columnHeaders={tableHeadings}
         columnHeadersBackgroundColor={"ui.gray.x-light-cool"}
--- a/src/components/SearchResults/SearchResultItems.tsx
+++ b/src/components/SearchResults/SearchResultItems.tsx
@@ -27,7 +27,7 @@ const SearchResultItems = ({ itemTableData }: SearchResultItemsProps) => {
             <tr key={heading}>
               <td
                 style={{
-                  width: "181px",
+                  width: "161px",
                   minWidth: "60px",
                 }}
               >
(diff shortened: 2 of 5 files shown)


src/components/ItemTable/ItemTable.tsx
```diff
@@ -16,6 +17,7 @@ const ItemTable = ({ itemTableData }: ItemTableProps) => {
     // Display as grid to prevent bug where the outer container stretches to the Table's width on mobile
     <Box display="grid">
       <Table
+        onCopy={(e) => handleTableCopy(e, "\t")}
```
7emansell, line 20: Nitpicky: since we're also stripping tab characters (I think this is intentionally not a `\n` here since that works in the `ItemTable` but lmk if i'm wrong), can you add a unit test for `\t` and update the `handleTableCopy` jsdoc and changelog to reflect that distinction (we're stripping tab and newline characters, not spaces, so not technically all whitespace. Would not normally get into this but bc this ticket has already been resurrected once I want to make sure our approach is really well documented).

src/utils/utilsTests/appUtils.test.ts
```diff
@@ -45,6 +56,33 @@ describe("appUtils", () => {
       expect(convertToSentenceCase("ISSN")).toEqual("ISSN")
     })
   })
+  describe("handleTableCopy", () => {
```
7emansell, line 59: Nice tests

7emansell (changes requested):
Works great for me on ff and chrome! Minor documentation request

src/components/ItemTable/ItemTable.tsx
KaseyW31: Yeah for sure, the focus is the `\n` and `\t` separators for the search results and item tables respectively. (Though copied segments that are empty strings after trimming whitespace are also filtered out. If extra leading/trailing spaces in the next cell over was copied that should be removed too, but I don't think that would typically happen..) Lmk if the updated wording makes sense

src/components/SearchResults/SearchResultItems.tsx
```diff
@@ -14,6 +15,7 @@ const SearchResultItems = ({ itemTableData }: SearchResultItemsProps) => {
   return (
     <Box>
       <table
+        onCopy={(e) => handleTableCopy(e, "\n")}
```
charmingduchess, line 18: why are we stripping different characters in these different places?

KaseyW31: I'm pretty sure the difference comes from the orientation of the tables. The search results tables have the headers at the left, and the bib page tables have the headers at the top. When trying to copy the call number RTL in search results, it can bleed into the header, which is separated by a newline. But in the bib page it's another data cell, so it's separated by a tab

KaseyW31: Actually it might just be how the different pages define the table - the search results table is native HTML, and the bib page is the DS `<Table>` component, so that could be affecting how the browser injects whitespace.

